PR #1, opened 2025-06-18
Outline proposed project objective and requirements

felixarntz:
This PR adds a `REQUIREMENTS.md`, outlining the overall objective for this project, as well as its key requirements. For easier consumption as Markdown, [see the file in the branch](https://github.com/WordPress/php-ai-client/blob/add/requirements-definition/docs/REQUIREMENTS.md).

It also adds a `GLOSSARY.md` with common terms, which can be helpful to align communication and discussion.

### DIFF
--- a/docs/REQUIREMENTS.md
+++ b/docs/REQUIREMENTS.md
@@ -1,0 +1,101 @@
+# Requirements
+
+This document outlines the requirements for the PHP AI Client, as well as the bigger picture for how it can eventually be used. The concrete technical architecture is defined and outlined in a separate document, based on these requirements.
+
+## Objective
+
+Enable calling any generative AI implementation using a uniform API in various programming languages.
+
+### Context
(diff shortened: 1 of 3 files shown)


swissspidy (approved):

docs/REQUIREMENTS.md
```diff
@@ -0,0 +1,101 @@
+# Requirements
```
swissspidy, line 1: Nit: could also list some constraints such as PHP version requirements

felixarntz: I'd like for this to be only focused on functional requirements for the project, so that ideally it can be evergreen. Just clarified that in the description.

You raise an important point though, we also need to define a few technical requirements like PHP version support. I just opened #3 for that.

Jameswlepage (approved):

PR #2, opened 2025-06-18
Outline proposed architecture based on requirements

felixarntz:
This proposed architecture is based on the current requirements (see #1).

**For easier review, please see the [Markdown version in the PR branch](https://github.com/WordPress/php-ai-client/blob/add/architecture-definition/docs/ARCHITECTURE.md).**

For reference: An older architectural outline and related ideas were discussed in https://github.com/felixarntz/ai-services/issues/22, and to some degree in https://github.com/felixarntz/ai-services/issues/21.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -181,11 +181,13 @@ _Note: Similarly to the previous example, even without specifying the model here
 
 ##### Fluent API
 ```php
+$history = [
+    new UserMessage('Do you spell it WordPress or Wordpress?'),
+    new AgentMessage('The correct spelling is WordPress.'),
+];
+
 $text = AiClient::prompt('Can you repeat that please?')
-    ->withHistory(
-        new UserMessage('Do you spell it WordPress or Wordpress?'),
-        new AgentMessage('The correct spelling is WordPress.')
-    )
+    ->withHistory(...$history)
     ->generateText();
 ```
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -198,16 +198,9 @@ This section shows comprehensive class diagrams for the proposed architecture. F
 
 **Note:** The class diagrams are also not meant to be comprehensive in terms of any specific configuration keys or parameters which are or will be supported. For now, the relevant definitions don't include any specific parameter names or constants.
 
-### Zoomed out view
+### Overview: Fluent API for AI consumption
 
-Below you find the zoomed out overview class diagram, looking at the two entrypoints for the largely decoupled APIs for:
-
-- Consuming AI capabilities.
-    - This is what the vast majority of developers will use.
-- Registering and implementing AI providers.
-    - This is what only developers that implement additional models or custom providers will use.
-
-Zoomed in views with detailed specifications for both of the APIs are found in the subsequent sections.
+This is a subset of the overall class diagram, purely focused on the fluent API for AI consumption.
 
 ```mermaid
 ---
@@ -219,8 +212,80 @@ classDiagram
 direction LR
     namespace Ai {
         class AiEntrypoint {
-            +defaultRegistry() AiProviderRegistry
-            +isConfigured(AiProviderAvailability $availability) bool$
+            +prompt(?string $text) PromptFacade$
+            +message(?string $text) MessageFacade$
+        }
+
+        class PromptFacade {
+            +withText(string $text) self
+            +withImageFile(File $file) self
+            +withAudioFile(File $file) self
+            +withVideoFile(File $file) self
+            +withFunctionResponse(FunctionResponse $functionResponse) self
+            +withMessageParts(...MessagePart $part) self
+            +withHistory(Message[] $messages) self
+            +usingModel(AiModel $model) self
+            +usingSystemInstruction(string|MessagePart[]|Message $systemInstruction) self
+            +usingTemperature(float $temperature) self
+            +usingTopP(float $topP) self
+            +usingTopK(int $topK) self
+            +usingStopSequences(...string $stopSequences) self
+            +usingCandidateCount(int $candidateCount) self
+            +usingOutputMime(string $mimeType) self
+            +usingOutputSchema(array< string, mixed > $schema) self
+            +usingOutputModalities(...AiModality $modalities) self
+            +generateResult() GenerativeAiResult
+            +generateOperation() GenerativeAiOperation
+            +generateTextResult() GenerativeAiResult
+            +streamGenerateTextResult() Generator< GenerativeAiResult >
+            +generateImageResult() GenerativeAiResult
+            +textToSpeechResult() GenerativeAiResult
+            +generateSpeechResult() GenerativeAiResult
+            +generateEmbeddingsResult() EmbeddingResult
+            +generateTextOperation() GenerativeAiOperation
+            +generateImageOperation() GenerativeAiOperation
+            +textToSpeechOperation() GenerativeAiOperation
+            +generateSpeechOperation() GenerativeAiOperation
+            +generateEmbeddingsOperation() EmbeddingOperation
+            +generateText() string
+            +streamGenerateText() Generator< string >
+            +generateImage() File
+            +textToSpeech() File
+            +generateSpeech() File
+            +generateEmbeddings() Embedding[]
+        }
+
+        class MessageFacade {
+            +usingRole(MessageRole $role) self
+            +withText(string $text) self
+            +withImageFile(File $file) self
+            +withAudioFile(File $file) self
+            +withVideoFile(File $file) self
+            +withFunctionCall(FunctionCall $functionCall) self
+            +withFunctionResponse(FunctionResponse $functionResponse) self
+            +withMessageParts(...MessagePart $part) self
+            +get() Message
+        }
+    }
+
+    AiEntrypoint .. PromptFacade : creates
+    AiEntrypoint .. MessageFacade : creates
+```
+
+### Overview: Traditional method call API for AI consumption
+
+This is a subset of the overall class diagram, purely focused on the traditional method call API for AI consumption.
+
+```mermaid
+---
+config:
+  class:
+    hideEmptyMembersBox: true
+---
+classDiagram
+direction LR
+    namespace Ai {
+        class AiEntrypoint {
             +generateResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
             +generateOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
             +generateTextResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
@@ -234,6 +299,26 @@ direction LR
             +textToSpeechOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
             +generateSpeechOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
             +generateEmbeddingsOperation(string[]|Message[] $input, AiModel $model) EmbeddingOperation$
+        }
+    }
+```
+
+### Overview: API for provider registration and implementation
+
+This is a subset of the overall class diagram, purely focused on the API for provider registration and implementation.
+
+```mermaid
+---
+config:
+  class:
+    hideEmptyMembersBox: true
+---
+classDiagram
+direction LR
+    namespace Ai {
+        class AiEntrypoint {
+            +defaultRegistry() AiProviderRegistry$
+            +isConfigured(AiProviderAvailability $availability) bool$
         }
     }
     namespace Ai.Providers {
@@ -251,7 +336,7 @@ direction LR
     AiEntrypoint "1" o-- "1..*" AiProviderRegistry
 ```
 
-### Class diagram zoomed in on AI consumption
+### Details: Class diagram for AI consumption
 
 ```mermaid
 ---
@@ -263,7 +348,9 @@ classDiagram
 direction LR
     namespace Ai {
         class AiEntrypoint {
-            +defaultRegistry() AiProviderRegistry
+            +prompt(?string $text) PromptFacade$
+            +message(?string $text) MessageFacade$
+            +defaultRegistry() AiProviderRegistry$
             +isConfigured(AiProviderAvailability $availability) bool$
             +generateResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
             +generateOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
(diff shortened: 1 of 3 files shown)


docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,673 @@
...
+            +generateImageResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
+            +textToSpeechResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
+            +generateSpeechResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
+            +generateEmbeddingsResult(string[]|Message[] $input, AiModel $model) EmbeddingResult$
+            +generateTextOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
+            +generateImageOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
+            +textToSpeechOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
+            +generateSpeechOperation(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiOperation$
+            +generateEmbeddingsOperation(string[]|Message[] $input, AiModel $model) EmbeddingOperation$
```
swissspidy: Personally I'd probably consolidate these into fewer methods to reduce the public API surface and make them more composable. For example, you can get a result easily via `generateOperation()`, no need for `generateResult()`.

Then, if starting with `GenerativeAiOperation` or `GenerativeAiResult`, there could be some `toText`, `toImage`, or `stream()` methods or so to transform the result into the desired shape.

Just thinking out loud though. Best to get some feedback out in the wild from developers building with it :)

felixarntz: A few thoughts on this, I'm partially in agreement, partially not, partially not sure.

* The `generate*Result()` vs `generate*Operation()` need to remain separate because they can fundamentally how they invoke an operation. Technically, you can wrap everything in an operation of course, but the implication of triggering an operation is very different from wanting a result right away - it's clear that an operation may take longer than you can wait for in this request, whereas wanting a result is explicitly waiting to get it right away.
* The same applies for streaming vs not streaming, it triggers a fundamentally different kind of request handling chain, so that needs to remain separate at the root too.
* For the methods `generateText()`, `generateImage()` etc., which technically would simply wrap `generateTextResult()`, `generateImageResult()` etc., I can see what you're saying would make sense. At this point the question is what is more intuitive and/or convenient for developers: `generateText()` or `generateTextResult()->text()`?
* Overall, most of these methods will be very brief wrappers of other methods. Pretty much all the heavy lifting will happen in `generateResult()` and `generateOperation()`, given the SDK is built with a multimodal first mindset. For example, `generateTextResult()` is basically just forwarding to `generateResult()` with an `outputModalities: [ 'text' ]` config arg injected. But passing that manually in would be very verbose, and while the API needs to be multimodal-first to be flexible, this would make usage unnecessary complex if you always have to think in that way - so calling `generateTextResult()` or `generateText()` feels way more intuitive if you only want to generate text for example.
* One exception is `streamGenerateTextResult()` which lives separately (not using `generateResult()` or `generateOperation()`), although we could even think there about how this could be abstracted to support multimodal streaming. That part goes a bit above my head right now, so it's not in here, but it certainly could be, if we want to support streaming beyond just text output.

TL;DR: For all the wrapper methods (which almost all of these are), we could consider handling them in another way. But I wouldn't say the current approach isn't ideal just because it's a large list of methods on the entrypoint object - it depends on what API developers consider more intuitive.

felixarntz: Obviously very limited due to character limit, but I just created https://x.com/felixarntz/status/1936116496658579717, maybe it'll give us at least a rough idea.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,673 @@
...
+    AiFeature::IMAGE_GENERATION
+);
+$imageFile = Ai::generateImage(
+    'Generate an illustration of the PHP elephant in the Carribean sea.',
+    Ai::defaultRegistry()->getProviderModel(
+        $providerModelsMetadata[0]->getProvider()->getId(),
+        $providerModelsMetadata[0]->getModels()[0]->getId()
+    )
+);
+```
```
JasonTheAdams: Looking at the Vercel AI SDK, the [ProviderV2](https://github.com/vercel/ai/blob/v5/packages/provider/src/provider/v2/provider-v2.ts) interface has some useful methods for grabbing models based on common capabilities — perhaps that's a simpler interface than a query based on the AiFeature and metadata? Maybe adding support for metadata-level queries could come in a later phase to simplify things?

docs/ARCHITECTURE.md
JasonTheAdams: Weighing in here, I'll focus on @swissspidy's comment with regards to reducing the public API surface.

_Personally,_ I prefer it when an API has a combination of robust and declarative methods. There's often a base public method with more imperative parameters which is used by a series of declarative methods. When possible, in my code, I'll prefer the declarative methods for readability and simplicity; when necessary, I'll use the base methods to construct my own declarative methods.

So I like the idea of having the declarative methods that can be used to simplify the implementing code for patterns that we know will be highly used. I like `generateTextResult()` which gives me the option to work with the object _and_ I like `generateText()` (and would likely use more often), which reduces the amount of times I'd need to do `generateTextResult()->text()`.

If a class has too many methods doing too many different things, then that's violating single responsibility. But if a class has many declarative methods for the same thing, that's useful, in my friendly opinion. 😄 

As a note, when writing documentation, you don't have to detail out every declarative method, or vice versa the base method — whichever is the more intended API.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,673 @@
...
+        class WithEmbeddingOperations {
+            +getOperation(string $operationId) EmbeddingOperation
+        }
+        class AiTextGenerationModel {
+            +generateTextResult(Message[] $prompt) GenerativeAiResult
+            +streamGenerateTextResult(Message[] $prompt) Generator< GenerativeAiResult >
+        }
+        class AiImageGenerationModel {
+            +generateImageResult(Message[] $prompt) GenerativeAiResult
+        }
```
JasonTheAdams, line 802: I like the idea of using the interface to infer the capabilities of the model. Should we also add a method for getting the metadata of said capability?

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,673 @@
...
+            +isProviderConfigured(string $idOrClassName) bool
+            +getProviderModel(string $idOrClassName, string $modelId, AiModelConfig|array $modelConfig) AiModel
+            +findProviderModelsMetadataForSupport(string $idOrClassName, AiFeature $feature, array<string, mixed > $capabilities) AiModelMetadata[]
+            +findModelsMetadataForSupport(AiFeature $feature, array<string, mixed > $capabilities) AiProviderModelMetadata[]
+        }
+    }
+    namespace Ai.Providers.Contracts {
+        class AiProvider {
+            +metadata() AiProviderMetadata$
+            +model(string $modelId, AiModelConfig|array< string, mixed > $modelConfig) AiModel$
```
JasonTheAdams, line 773: What does the second parameter do?

JasonTheAdams (commented):
A lot of good stuff here! I don't think I've fully wrapped my head around the entire API, but I've got a good idea. I left some notes. Also, this PR notes that it's imperative that this satisfies every must-have of the requirements doc, but it doesn't address the REST API. Perhaps that shouldn't be a must-have requirement based on our conversations surrounding that?

docs/ARCHITECTURE.md
felixarntz: Thanks for the feedback @JasonTheAdams, I personally agree with it completely.

That said, it's always easier to add additional API surfaces later than remove them (if they show to be undesirable), and the outcome of the high-level https://x.com/felixarntz/status/1936116496658579717 survey shows a preference to calling the nested method to "transform" the overall result object.

Based on this, for now I'm leaning towards following that hint and remove the additional _declarative_ methods in favor of "transform" methods on the result objects. Again, this isn't set in stone and could be changed in the future, but I think it's reasonable to go with what the majority of the survey respondents favored as a starting point.

docs/ARCHITECTURE.md
felixarntz: See my reply in https://github.com/WordPress/php-ai-client/issues/4#issuecomment-3020383379 for why I think including more than just high-level capability is important.

That said, I think we could provide some additional _declarative_ methods here (somewhat related to your other comment above) to make the code needed for this less verbose. For example:

```php
$imageFile = Ai::generateImage(
    'Generate an illustration of the PHP elephant in the Carribean sea.',
    Ai::defaultRegistry()->findProviderModelForSupport(
        AiFeature::IMAGE_GENERATION
    )
);
```

This would do the same as the above code example, but by bundling these two commonly sequenced steps into one, it would make for a more intuitive API for the use-case.

docs/ARCHITECTURE.md
felixarntz: My idea so far is that metadata would be handled centrally for each provider and stored in `AiModelMetadata` objects, which then the model API object (e.g. this `AiImageGenerationModel` here) would receive. So there should definitely be a method to get the metadata from the model API object, but it would be declared on the `AiModel` interface, which every model (including `AiImageGenerationModel`) would implement.

docs/ARCHITECTURE.md
felixarntz: This parameter would be for all the additional configuration of the model, e.g.:
* Parameters like `temperature`, `topK`, `topP`, etc.
* System instruction, if needed.
* Function declarations to provide for function calling, if needed.
* JSON schema to follow in case JSON output is requested.
* ...

felixarntz:
@JasonTheAdams 
> Also, this PR notes that it's imperative that this satisfies every must-have of the requirements doc, but it doesn't address the REST API. Perhaps that shouldn't be a must-have requirement based on our conversations surrounding that?

Maybe this is not clear from the current `REQUIREMENTS.md`, but I just double checked and didn't see the REST API as a must-have there. Can you point to what made you think it was? Probably something we can clarify in that doc.

docs/ARCHITECTURE.md
felixarntz: I've just updated this in https://github.com/WordPress/php-ai-client/pull/2/commits/42c1a83549d78c69bf284f959a3e1a54380e75ce, that diff shows the changes clearly. I'm personally not in love with this as it feels more verbose, but let's start with that.

Follow-up question: The new utility methods are all named `to...()`, e.g. `toText()`. This is in line with what was in the Twitter survey, and it provides a separation between simple getters that return a direct child object of `GenerativeAiResult` and these methods, which get more deeply nested pieces of data, including bulk-extracting from arrays in some cases (when there are multiple AI response candidates returned by the model). Now my question is: Does that separation help, or is it confusing?

I raise this because e.g. `generateTextResult()->getCandidates()` but `generateTextResult()->toTexts()`, or, if you go the very verbose route, `generateTextResult()->getCandidates()->toTexts()`.

Should all methods be called `get...()`, e.g. `getText()` instead, or is the current naming good?

JasonTheAdams:
> Maybe this is not clear from the current `REQUIREMENTS.md`, but I just double checked and didn't see the REST API as a must-have there. Can you point to what made you think it was? Probably something we can clarify in that doc.

You're totally right. It's mentioned, but not in the requirements. Apologies. 😆

felixarntz:
The architecture proposed here includes the concept of long-running operations, which leads to the distinction between "Operation" objects and "Result" objects for the various generative AI tasks.

@JasonTheAdams raised this on Slack, please see [this thread](https://wordpress.slack.com/archives/C08TJ8BPULS/p1751906636337599) for additional context about this approach.

swissspidy (approved):

JasonTheAdams:
Per my conversation @felixarntz, here are the examples reflected in a Fluent API method:

#### Generate text using any suitable model from any provider (most basic example)
```php
$text = Ai::prompt('Write a 2-verse poem about PHP.')->generateText();
```

#### Generate text using a Google model
```php
$text = Ai::prompt('Write a 2-verse poem about PHP.')
  ->usingModel('gemini-2.5-flash')
  ->generateText();
```

#### Generate multiple text candidates using an Anthropic model
```php
$text = Ai::prompt('Write a 2-verse poem about PHP.')
  ->usingModel('claude-3.7-sonnet')
  ->generateTexts(4);
```

#### Generate an image using any suitable OpenAI model
```php
$text = Ai::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
  ->usingProvider('openai')
  ->usingModelSupportingImages() // optional
  ->generateImageResult()
  ->toFile();
```

#### Generate an image using any suitable model from any provider
```php
$text = Ai::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
  ->usingModelSupportingImages() // optional
  ->generateImageResult()
  ->toFile();
```

#### Generate text using any suitable model from any provider
```php
$text = Ai::prompt('Write a 2-verse poem about PHP.')
  ->usingModelSupportingText()
  ->generateText();
```

#### Generate text with an image as additional input using any suitable model from any provider
```php
$text = Ai::prompt('Generate alternative text for this image.')
  ->withImage('image/png', $base64blob)
  ->generateText();
```

#### Generate text with chat history using any suitable model from any provider
```php
$text = Ai::prompt('Can you repeat that please?')
  ->withHistory(
    new UserMessage('Do you spell it WordPress or Wordpress?'),
    new AgentMessage('The correct spelling is WordPress.')
  )
  ->generateText();
```

#### Generate text with JSON output using any suitable model from any provider
```php
$text = Ai::prompt('Transform the following CSV content into a JSON array of row data.')
  ->asJsonResponse
  ->withOutputSchema(['name' => 'string', 'age' => 'integer'])
  ->generateText();
```

#### Generate embeddings using any suitable model from any provider
```php
$embeddings = Ai::prompt('A very long text.', 'Another very long text.', 'More long text.')
  ->generateEmbeddings();
```

felixarntz:
Note to self: This PR so far talks about "API for consumption" vs "API for provider registration and implementation".

Based on the target audiences clearly outlined in #11, we should probably update the wording here to align with that, e.g. "Implementer API" vs "Extender API".

cc @JasonTheAdams

felixarntz:
@JasonTheAdams @borkweb FYI https://github.com/WordPress/php-ai-client/pull/2/commits/639d6271a18919e46b14d3e26a1673a079c81a4c is required as a workaround because Mermaid doesn't like namespaces that have the same names as a class.

The actual namespace will of course be `AiClient`, as already defined in `composer.json`.

felixarntz:
@JasonTheAdams @borkweb https://github.com/WordPress/php-ai-client/pull/2/commits/f7e8f69965383da74d185bea331367d46373644e adds the child classes for `Message` that we had previously discussed to have for ease of use, plus they're already referenced in the examples (also see related fix https://github.com/WordPress/php-ai-client/pull/2/commits/3ff376eceff38ba5c0740cea74a13b8e9a7067e5).

I'm quite happy with this as is now, so I'd say please give it another review to see whether you are too!

I already spotted #21, and I like this direction quite a bit, but I'd say let's get this merged independently of the namespacing/directory structure and create a follow up PR to revise based on the discussion in #21.

borkweb (approved):
Looks great to me - excellent catch on the `UserMessage` and `ModelMessage` objects 🍸

JasonTheAdams (approved):
LGTM! 🎉

PR #10, opened 2025-07-10
Revise proposed architecture PR with fluent API class diagrams

felixarntz:
This PR simply updates the branch for #2 with class diagrams that encompass the fluent API (since I had separately explored them in its own branch prior to today's discussion).

A quick review shall be sufficient here, as all of this will still have to undergo a more comprehensive review as part of #2.

After merging this into #2, a good next step would be to update the code examples, ideally with a side-by-side comparison of both consumer / implementor facing APIs (fluent API and traditional method call API).

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -198,16 +198,9 @@ This section shows comprehensive class diagrams for the proposed architecture. F
 
 **Note:** The class diagrams are also not meant to be comprehensive in terms of any specific configuration keys or parameters which are or will be supported. For now, the relevant definitions don't include any specific parameter names or constants.
 
-### Zoomed out view
+### Overview: Fluent API for AI consumption
 
-Below you find the zoomed out overview class diagram, looking at the two entrypoints for the largely decoupled APIs for:
-
-- Consuming AI capabilities.
-    - This is what the vast majority of developers will use.
-- Registering and implementing AI providers.
-    - This is what only developers that implement additional models or custom providers will use.
-
-Zoomed in views with detailed specifications for both of the APIs are found in the subsequent sections.
+This is a subset of the overall class diagram, purely focused on the fluent API for AI consumption.
 
 ```mermaid
 ---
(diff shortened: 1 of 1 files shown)


docs/ARCHITECTURE.md
```diff
@@ -198,16 +198,83 @@ This section shows comprehensive class diagrams for the proposed architecture. F
...
+            +usingCandidateCount(int $candidateCount) self
+            +usingOutputMime(string $mimeType) self
+            +usingOutputSchema(array< string, mixed > $schema) self
+            +usingOutputModalities(...AiModality $modalities) self
+            +generateResult() GenerativeAiResult
+            +generateOperation() GenerativeAiOperation
+            +generateTextResult() GenerativeAiResult
+            +streamGenerateTextResult() Generator< GenerativeAiResult >
+            +generateImageResult() GenerativeAiResult
+            +textToSpeechResult() GenerativeAiResult
```
JasonTheAdams: Why aren't the `textToSpeech()` methods prefixed with `generate`?

docs/ARCHITECTURE.md
```diff
@@ -198,16 +198,83 @@ This section shows comprehensive class diagrams for the proposed architecture. F
...
+        }
+
+        class PromptBuilder {
+            +withText(string $text) self
+            +withImageFile(File $file) self
+            +withAudioFile(File $file) self
+            +withVideoFile(File $file) self
+            +withFunctionResponse(FunctionResponse $functionResponse) self
+            +withMessageParts(...MessagePart $part) self
+            +withHistory(Message[] $messages) self
```
JasonTheAdams:
```suggestion
            +withHistory(...Message $messages) self
```
Let's use variadic here as well. You can always do `withHistory(...$messages)` if you have an array already.

docs/ARCHITECTURE.md
```diff
@@ -198,16 +198,83 @@ This section shows comprehensive class diagrams for the proposed architecture. F
...
+            +prompt(?string $text) PromptBuilder$
+            +message(?string $text) MessageBuilder$
+        }
+
+        class PromptBuilder {
+            +withText(string $text) self
+            +withImageFile(File $file) self
+            +withAudioFile(File $file) self
+            +withVideoFile(File $file) self
+            +withFunctionResponse(FunctionResponse $functionResponse) self
```
JasonTheAdams, line 224: What does this do?

docs/ARCHITECTURE.md
```diff
@@ -198,16 +198,83 @@ This section shows comprehensive class diagrams for the proposed architecture. F
...
+            +generateEmbeddingsOperation() EmbeddingOperation
+            +generateText() string
+            +streamGenerateText() Generator< string >
+            +generateImage() File
+            +textToSpeech() File
+            +generateSpeech() File
+            +generateEmbeddings() Embedding[]
+        }
+
+        class MessageBuilder {
```
JasonTheAdams, line 258: What is a message as opposed to a prompt?

docs/ARCHITECTURE.md
```diff
@@ -198,16 +198,83 @@ This section shows comprehensive class diagrams for the proposed architecture. F
...
+---
+classDiagram
+direction LR
+    namespace Ai {
+        class AiEntrypoint {
+            +prompt(?string $text) PromptBuilder$
+            +message(?string $text) MessageBuilder$
+        }
+
+        class PromptBuilder {
```
JasonTheAdams, line 219: This will also have a bunch of getter methods so the object consumer can do it's thing, unless you're planning on having a separate DTO that goes the model?

JasonTheAdams (changes requested):

JasonTheAdams:
Excited to see this! Left a few thoughts and questions, @felixarntz!

docs/ARCHITECTURE.md
felixarntz: This mostly stems from that "text to speech" as a term denotes its own type of AI feature, which is not "generative AI" in the same sense as text generation, image generation, speech generation, video generation etc. are.

"text to speech" is more about what I'd call "translating text to audio", but it's not coming up with anything creative or original.

I'm definitely open to other naming suggestions here, mostly since this method is missing a verb. But I don't think "generate" is right.

docs/ARCHITECTURE.md
felixarntz: Great point, updated.

docs/ARCHITECTURE.md
felixarntz: It adds a function response as a message part, typically this needs to be used in a user message, in response to the LLM returning a function call message part.

Technically similar to the other methods here that add message parts.

docs/ARCHITECTURE.md
felixarntz: A `Message` is the main data transfer object to communicate between the user and the LLM. There are three types of `Message`s: user, model, and system messages.

A "prompt" in this SDK is not really an actual thing. It's a higher-level overarching term for what could be a single user message, or many user and model messages (in case of history), or a user message with a bunch of configuration arguments.

I used the name `PromptBuilder` to relate to that overarching term, but we could go with a different name. Maybe something like `AiRequestBuilder` could be used instead.

The reason there's a `MessageBuilder` in addition to the (what is called now) `PromptBuilder` is to easily be able to create `Message` objects, which is mostly relevant when using `PromptBuilder::withHistory()`.

docs/ARCHITECTURE.md
felixarntz: Yeah, the model will receive this in a different way, see the detailed class diagram. For the most part, it's 1 or more `Message` objects, and an `AiModelConfig` object.

docs/ARCHITECTURE.md
JasonTheAdams: I poetically used Gemini to get some thoughts here. Curious what you think!

image

felixarntz: Per our conversation, in https://github.com/WordPress/php-ai-client/commit/8fa8b210be7f90f26f5fbb5f8098b0732b37dbf5 I updated all the `textToSpeech` naming to use `convertTextToSpeech`.

JasonTheAdams (approved):
Love it! Great work! 🎉

PR #12, opened 2025-07-12
Add fluent API code examples to architecture documentation

borkweb:
This proposed set of changes to #2 and is doing a few things:

1. Adding Fluent API code examples (similar to those suggested by @JasonTheAdams [here](https://github.com/WordPress/php-ai-client/pull/2#issuecomment-3054047920)) alongside the Traditional API examples.
2. Proposing a handful of helper methods to simplify the usage of Candidate Count and avoid implementers from needing to understand what _Candidate Count_ means.
3. Representing @JasonTheAdams' suggestion of `AiClient::withImage(string $mimeType, string $base64Blob)` to mimic the input simplicity that the Traditional API supports.
4. Adding the `AiClient::asJsonResponse()` and `AiClient::asArrayResponse()` as a helper methods for `usingOutputMime('application/json')`
5. Adding helper methods for specifying models that support a certain capability or option.

Here's the [ARCHITECTURE.md file](https://github.com/borkweb/php-ai-client/blob/add/fluent-code-examples/docs/ARCHITECTURE.md) in a readable format.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -181,11 +181,13 @@ _Note: Similarly to the previous example, even without specifying the model here
 
 ##### Fluent API
 ```php
+$history = [
+    new UserMessage('Do you spell it WordPress or Wordpress?'),
+    new AgentMessage('The correct spelling is WordPress.'),
+];
+
 $text = AiClient::prompt('Can you repeat that please?')
-    ->withHistory(
-        new UserMessage('Do you spell it WordPress or Wordpress?'),
-        new AgentMessage('The correct spelling is WordPress.')
-    )
+    ->withHistory(...$history)
     ->generateText();
 ```
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -156,7 +156,7 @@ _Note: Since this omits the model parameter, the SDK will automatically determin
 ##### Fluent API
 ```php
 $text = AiClient::prompt('Generate alternative text for this image.')
-    ->withImageFile(new InlineFile('image/png', $base64blob))
+    ->withInlineImage($base64Blob, 'image/png')
     ->generateText();
 ```
 
@@ -328,6 +328,9 @@ direction LR
 
         class PromptBuilder {
             +withText(string $text) self
+            +withInlineImage(string $base64Blob, string $mimeType)
+            +withLocalImage(string $path, string $mimeType)
+            +withRemoteImage(string $uri, string $mimeType)
             +withImageFile(File $file) self
             +withAudioFile(File $file) self
             +withVideoFile(File $file) self
@@ -487,6 +490,9 @@ direction LR
 
         class PromptBuilder {
             +withText(string $text) self
+            +withInlineImage(string $base64Blob, string $mimeType)
+            +withLocalImage(string $path, string $mimeType)
+            +withRemoteImage(string $uri, string $mimeType)
             +withImageFile(File $file) self
             +withAudioFile(File $file) self
             +withVideoFile(File $file) self
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -362,14 +362,14 @@ direction LR
             +generateSpeechOperation() GenerativeAiOperation
             +generateEmbeddingsOperation() EmbeddingOperation
             +generateText() string
-            +generateTexts(int $candidateCount) string
+            +generateTexts(int $candidateCount) string[]
             +streamGenerateText() Generator< string >
             +generateImage() File
-            +generateImages(int $candidateCount) File
+            +generateImages(int $candidateCount) File[]
             +convertTextToSpeech() File
-            +convertTextToSpeeches(int $candidateCount) File
+            +convertTextToSpeeches(int $candidateCount) File[]
             +generateSpeech() File
-            +generateSpeeches(int $candidateCount) File
+            +generateSpeeches(int $candidateCount) File[]
             +generateEmbeddings() Embedding[]
             +getModelRequirements() AiModelRequirements
             +isSupported() bool
@@ -524,14 +524,14 @@ direction LR
             +generateSpeechOperation() GenerativeAiOperation
             +generateEmbeddingsOperation() EmbeddingOperation
             +generateText() string
-            +generateTexts(int $candidateCount) string
+            +generateTexts(int $candidateCount) string[]
             +streamGenerateText() Generator< string >
             +generateImage() File
-            +generateImages(int $candidateCount) File
+            +generateImages(int $candidateCount) File[]
             +convertTextToSpeech() File
-            +convertTextToSpeeches(int $candidateCount) File
+            +convertTextToSpeeches(int $candidateCount) File[]
             +generateSpeech() File
-            +generateSpeeches(int $candidateCount) File
+            +generateSpeeches(int $candidateCount) File[]
             +generateEmbeddings() Embedding[]
             +getModelRequirements() AiModelRequirements
             +isSupported() bool
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -321,8 +321,8 @@ classDiagram
 direction LR
     namespace Ai {
         class AiClient {
-            +prompt(string ...$text) PromptBuilder$
-            +message(string ...$text) MessageBuilder$
+            +prompt(string|Message ...$text) PromptBuilder$
+            +message(string $text) MessageBuilder$
         }
 
         class PromptBuilder {
@@ -468,8 +468,8 @@ classDiagram
 direction LR
     namespace Ai {
         class AiClient {
-            +prompt(string ...$text) PromptBuilder$
-            +message(string ...$text) MessageBuilder$
+            +prompt(string|Message ...$text) PromptBuilder$
+            +message(string $text) MessageBuilder$
             +defaultRegistry() AiProviderRegistry$
             +isConfigured(AiProviderAvailability $availability) bool$
             +generateResult(string|MessagePart|MessagePart[]|Message|Message[] $prompt, AiModel $model) GenerativeAiResult$
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -346,38 +346,27 @@ direction LR
             +usingOutputModalities(...AiModality $modalities) self
             +asJsonResponse(?array< string, mixed > $schema) self
             +generateResult() GenerativeAiResult
-            +generateResults(int $candidateCount) GenerativeAiResult[]
             +generateOperation() GenerativeAiOperation
-            +generateOperations(int $candidateCount) GenerativeAiOperation[]
             +generateTextResult() GenerativeAiResult
-            +generateTextResults(int $candidateCount) GenerativeAiResult[]
             +streamGenerateTextResult() Generator< GenerativeAiResult >
             +generateImageResult() GenerativeAiResult
-            +generateImageResults(int $candidateCount) GenerativeAiResult[]
             +convertTextToSpeechResult() GenerativeAiResult
-            +convertTextToSpeechResults(int $candidateCount) GenerativeAiResult[]
             +generateSpeechResult() GenerativeAiResult
-            +generateSpeechResults(int $candidateCount) GenerativeAiResult[]
             +generateEmbeddingsResult() EmbeddingResult
-            +generateEmbeddingsResults(int $candidateCount) EmbeddingResult[]
             +generateTextOperation() GenerativeAiOperation
-            +generateTextOperations(int $candidateCount) GenerativeAiOperation[]
             +generateImageOperation() GenerativeAiOperation
-            +generateImageOperations(int $candidateCount) GenerativeAiOperation[]
             +convertTextToSpeechOperation() GenerativeAiOperation
             +generateSpeechOperation() GenerativeAiOperation
-            +generateSpeechOperations(int $candidateCount) GenerativeAiOperation[]
             +generateEmbeddingsOperation() EmbeddingOperation
-            +generateEmbeddingsOperations(int $candidateCount) EmbeddingOperation[]
             +generateText() string
-            +generateTexts(int $candidateCount) string[]
+            +generateTexts(int $candidateCount) string
             +streamGenerateText() Generator< string >
             +generateImage() File
-            +generateImages(int $candidateCount) File[]
+            +generateImages(int $candidateCount) File
             +convertTextToSpeech() File
-            +convertTextToSpeeches(int $candidateCount) File[]
+            +convertTextToSpeeches(int $candidateCount) File
             +generateSpeech() File
-            +generateSpeeches(int $candidateCount) File[]
+            +generateSpeeches(int $candidateCount) File
             +generateEmbeddings() Embedding[]
             +getModelRequirements() AiModelRequirements
             +isSupported() bool
@@ -518,38 +507,27 @@ direction LR
             +usingOutputModalities(...AiModality $modalities) self
             +asJsonResponse(?array< string, mixed > $schema) self
             +generateResult() GenerativeAiResult
-            +generateResults(int $candidateCount) GenerativeAiResult[]
             +generateOperation() GenerativeAiOperation
-            +generateOperations(int $candidateCount) GenerativeAiOperation[]
             +generateTextResult() GenerativeAiResult
-            +generateTextResults(int $candidateCount) GenerativeAiResult[]
             +streamGenerateTextResult() Generator< GenerativeAiResult >
             +generateImageResult() GenerativeAiResult
-            +generateImageResults(int $candidateCount) GenerativeAiResult[]
             +convertTextToSpeechResult() GenerativeAiResult
-            +convertTextToSpeechResults(int $candidateCount) GenerativeAiResult[]
             +generateSpeechResult() GenerativeAiResult
-            +generateSpeechResults(int $candidateCount) GenerativeAiResult[]
             +generateEmbeddingsResult() EmbeddingResult
-            +generateEmbeddingsResults(int $candidateCount) EmbeddingResult[]
             +generateTextOperation() GenerativeAiOperation
-            +generateTextOperations(int $candidateCount) GenerativeAiOperation[]
             +generateImageOperation() GenerativeAiOperation
-            +generateImageOperations(int $candidateCount) GenerativeAiOperation[]
             +convertTextToSpeechOperation() GenerativeAiOperation
             +generateSpeechOperation() GenerativeAiOperation
-            +generateSpeechOperations(int $candidateCount) GenerativeAiOperation[]
             +generateEmbeddingsOperation() EmbeddingOperation
-            +generateEmbeddingsOperations(int $candidateCount) EmbeddingOperation[]
             +generateText() string
-            +generateTexts(int $candidateCount) string[]
+            +generateTexts(int $candidateCount) string
             +streamGenerateText() Generator< string >
             +generateImage() File
-            +generateImages(int $candidateCount) File[]
+            +generateImages(int $candidateCount) File
             +convertTextToSpeech() File
-            +convertTextToSpeeches(int $candidateCount) File[]
+            +convertTextToSpeeches(int $candidateCount) File
             +generateSpeech() File
-            +generateSpeeches(int $candidateCount) File[]
+            +generateSpeeches(int $candidateCount) File
             +generateEmbeddings() Embedding[]
             +getModelRequirements() AiModelRequirements
             +isSupported() bool
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -273,32 +273,6 @@ $jsonString = AiClient::generateTextResult(
 )->toText();
 ```
 
-#### Generate embeddings using any suitable model from any provider
-
-##### Fluent API
-```php
-$embeddings = AiClient::prompt('A very long text.', 'Another very long text.', 'More long text.')
-    ->generateEmbeddings();
-```
-
-##### Traditional API
-```php
-$providerModelsMetadata = AiClient::defaultRegistry()->findModelsMetadataForSupport(
-    new AiModelRequirements([AiCapability::EMBEDDING_GENERATION])
-);
-$embeddings = AiClient::generateEmbeddingsResult(
-    [
-        'A very long text.',
-        'Another very long text.',
-        'More long text.',
-    ],
-    AiClient::defaultRegistry()->getProviderModel(
-        $providerModelsMetadata[0]->getProvider()->getId(),
-        $providerModelsMetadata[0]->getModels()[0]->getId()
-    )
-)->getEmbeddings();
-```
-
 ## Class diagrams
 
 This section shows comprehensive class diagrams for the proposed architecture. For explanation on specific terms, see the [glossary](./GLOSSARY.md).
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -75,7 +75,6 @@ $texts = AiClient::generateTextResult(
 ```php
 $imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
     ->usingProvider('openai')
-    ->usingModelSupportingImages() // Optional.
     ->generateImage();
 ```
 
@@ -99,7 +98,6 @@ $imageFile = AiClient::generateImageResult(
 ##### Fluent API
 ```php
 $imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
-    ->usingModelSupportingImages() // Optional.
     ->generateImage();
 ```
 
@@ -123,8 +121,17 @@ _Note: This does effectively the exact same as [the first code example](#generat
 
 ##### Fluent API
 ```php
+$providerModelsMetadata = AiClient::defaultRegistry()->findModelsMetadataForSupport(
+    new AiModelRequirements([AiCapability::TEXT_GENERATION])
+);
+
 $text = AiClient::prompt('Write a 2-verse poem about PHP.')
-    ->usingModelSupportingText() // Optional.
+    ->withModel(
+        AiClient::defaultRegistry()->getProviderModel(
+            $providerModelsMetadata[0]->getProvider()->getId(),
+            $providerModelsMetadata[0]->getModels()[0]->getId()
+        )
+    )
     ->generateText();
 ```
 
@@ -320,20 +327,6 @@ direction LR
             +withMessageParts(...MessagePart $part) self
             +withHistory(...Message $messages) self
             +usingModel(AiModel $model) self
-            +usingModelSupporting(...AiCapability|AiOption $aiCapabilityOrOption) self
-            +usingModelSupportingCapability(...AiCapability $aiCapability) self
-            +usingModelSupportingOption(...AiOption $aiOption) self
-            +usingModelSupportingAudio() self
-            +usingModelSupportingHistory() self
-            +usingModelSupportingEmbeddings() self
-            +usingModelSupportingImages() self
-            +usingModelSupportingJsonOutput() self
-            +usingModelSupportingMusic() self
-            +usingModelSupportingOutputSchema() self
-            +usingModelSupportingSpeech() self
-            +usingModelSupportingText() self
-            +usingModelSupportingTextToSpeech() self
-            +usingModelSupportingVideo() self
             +usingSystemInstruction(string|MessagePart[]|Message $systemInstruction) self
             +usingMaxTokens(int $maxTokens) self
             +usingTemperature(float $temperature) self
@@ -507,20 +500,6 @@ direction LR
             +withMessageParts(...MessagePart $part) self
             +withHistory(...Message $messages) self
             +usingModel(AiModel $model) self
-            +usingModelSupporting(...AiCapability|AiOption $aiCapabilityOrOption) self
-            +usingModelSupportingCapability(...AiCapability $aiCapability) self
-            +usingModelSupportingOption(...AiOption $aiOption) self
-            +usingModelSupportingAudio() self
-            +usingModelSupportingHistory() self
-            +usingModelSupportingEmbeddings() self
-            +usingModelSupportingImages() self
-            +usingModelSupportingJsonOutput() self
-            +usingModelSupportingMusic() self
-            +usingModelSupportingOutputSchema() self
-            +usingModelSupportingSpeech() self
-            +usingModelSupportingText() self
-            +usingModelSupportingTextToSpeech() self
-            +usingModelSupportingVideo() self
             +usingSystemInstruction(string|MessagePart[]|Message $systemInstruction) self
             +usingMaxTokens(int $maxTokens) self
             +usingTemperature(float $temperature) self
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -19,6 +19,13 @@ The following examples indicate how this SDK could eventually be used.
 
 #### Generate text using any suitable model from any provider (most basic example)
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->generateText();
+```
+
+##### Traditional API
 ```php
 $text = AiClient::generateTextResult(
     'Write a 2-verse poem about PHP.'
@@ -27,6 +34,14 @@ $text = AiClient::generateTextResult(
 
 #### Generate text using a Google model
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModel('gemini-2.5-flash')
+    ->generateText();
+```
+
+##### Traditional API
 ```php
 $text = AiClient::generateTextResult(
     'Write a 2-verse poem about PHP.',
@@ -36,6 +51,14 @@ $text = AiClient::generateTextResult(
 
 #### Generate multiple text candidates using an Anthropic model
 
+##### Fluent API
+```php
+$texts = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModel('claude-3.7-sonnet')
+    ->generateTexts(4);
+```
+
+##### Traditional API
 ```php
 $texts = AiClient::generateTextResult(
     'Write a 2-verse poem about PHP.',
@@ -48,6 +71,15 @@ $texts = AiClient::generateTextResult(
 
 #### Generate an image using any suitable OpenAI model
 
+##### Fluent API
+```php
+$imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
+    ->usingProvider('openai')
+    ->usingModelSupportingImages() // Optional.
+    ->generateImage();
+```
+
+##### Traditional API
 ```php
 $modelsMetadata = AiClient::defaultRegistry()->findProviderModelsMetadataForSupport(
     'openai',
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
Thanks, @borkweb! I believe we're actually only going with the Fluent API at this point, with the idea we can add the other API style later if it makes sense to do so.

Is that right, @felixarntz?

borkweb:
Ha! Excellent. That was going to be a question I was going to ask - maintaining both would be laborious

felixarntz:
@JasonTheAdams @borkweb
> I believe we're actually only going with the Fluent API at this point, with the idea we can add the other API style later if it makes sense to do so.

I think the decision on Thursday was to go with both, in order to compare developer interest/usage in them and eventually decide on the final approach based on that.

The fluent API should be the baseline, and we should start with building that, but the traditional method call API should still be part of the planned architecture. I don't think it's a big lift to wrap the fluent API with it anyway.

I understand and even share your preference for the fluent API, but as discussed on Thursday we should properly consider both APIs with developers especially in the early stages of this SDK.

docs/ARCHITECTURE.md
```diff
@@ -27,6 +34,14 @@ $text = AiClient::generateTextResult(
 
 #### Generate text using a Google model
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModel('gemini-2.5-flash')
```
felixarntz: This needs to also specify the provider. Let's use the same syntax as the traditional method call API for this, which will tie in with how provider classes should be implemented:
```suggestion
    ->usingModel(Google::model('gemini-2.5-flash'))
```

docs/ARCHITECTURE.md
```diff
@@ -36,6 +51,14 @@ $text = AiClient::generateTextResult(
 
 #### Generate multiple text candidates using an Anthropic model
 
+##### Fluent API
+```php
+$texts = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModel('claude-3.7-sonnet')
```
felixarntz: See above:
```suggestion
    ->usingModel(Anthropic::model('claude-3.7-sonnet'))
```

docs/ARCHITECTURE.md
```diff
@@ -36,6 +51,14 @@ $text = AiClient::generateTextResult(
...
+    ->generateTexts(4);
```
felixarntz, line 58: Worth noting that having the parameter `4` here would be optional, and it would technically need to be a shorthand for something like `->usingCandidateCount(4)->generateTexts()`.

The reason that `usingCandidateCount()` must also exist as a method is for scenarios where you'll want the full result object instead of just the text strings: Calling `getTextResult(4)` wouldn't look right, because it would still get only a single result object, but with 4 candidates inside.

docs/ARCHITECTURE.md
```diff
@@ -48,6 +71,15 @@ $texts = AiClient::generateTextResult(
 
 #### Generate an image using any suitable OpenAI model
 
+##### Fluent API
+```php
+$imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
+    ->usingProvider('openai')
+    ->usingModelSupportingImages() // Optional.
```
felixarntz: Let's remove this. The other fluent API examples mostly focus on the most concise way to write things, and I'm not sure what the value of having `usingModelSupportingImages()` would be in the first place, as that would automatically be inferred from the input and the `generateImage()` call.
```suggestion
```

docs/ARCHITECTURE.md
```diff
@@ -64,6 +96,14 @@ $imageFile = AiClient::generateImageResult(
 
 #### Generate an image using any suitable model from any provider
 
+##### Fluent API
+```php
+$imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
+    ->usingModelSupportingImages() // Optional.
```
felixarntz: See above:
```suggestion
```

docs/ARCHITECTURE.md
```diff
@@ -81,6 +121,14 @@ $imageFile = AiClient::generateImageResult(
 
 _Note: This does effectively the exact same as [the first code example](#generate-text-using-any-suitable-model-from-any-provider-most-basic-example), but more verbosely. In other words, if you omit the model parameter, the SDK will do this internally._
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModelSupportingText() // Optional.
+    ->generateText();
```
felixarntz: See above, this shouldn't be here.

The code for the traditional API does the same thing as the previous code, but is explicitly outlined as the more verbose way to do that, but I'm not sure we'll need additional declarative methods like `usingModelSupportingText()` for this in the fluent API, given the advanced nature of this in the first place.

If anything, I think we should show here the equivalent of what the code example below does, but using the fluent API, clarifying similarly that this would be the verbose way of doing it:

```suggestion
$providerModelsMetadata = AiClient::defaultRegistry()->findModelsMetadataForSupport(
    new AiModelRequirements([AiCapability::TEXT_GENERATION])
);
$text = AiClient::prompt('Write a 2-verse poem about PHP.')
    ->usingModel(
        AiClient::defaultRegistry()->getProviderModel(
            $providerModelsMetadata[0]->getProvider()->getId(),
            $providerModelsMetadata[0]->getModels()[0]->getId()
        )
    )
    ->generateText();
```

Naturally, this is not an intuitive example of how the fluent API is used - but so isn't the traditional API example below. They are supposed to indicate what happens under the hood, and what you _could_ do instead, e.g. if you wanted some really custom model filtering logic (since `$providerModelsMetadata` contains multiple models, so you could apply custom logic to choose the model to actually use out of them).

docs/ARCHITECTURE.md
```diff
@@ -98,6 +146,14 @@ $text = AiClient::generateTextResult(
 
 _Note: Since this omits the model parameter, the SDK will automatically determine which models are suitable and use any of them, similar to [the first code example](#generate-text-using-any-suitable-model-from-any-provider-most-basic-example). Since it knows the input includes an image, it can internally infer that the model needs to not only support `AiCapability::TEXT_GENERATION`, but also `AiOption::INPUT_MODALITIES => ['text', 'image']`._
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Generate alternative text for this image.')
+    ->withImage('image/png', $base64blob)
```
felixarntz: I'm not sure this method is flexible enough - not every image would be provided as base64 data.

The architecture already caters for this by providing classes `InlineFile`, `LocalFile`, and `RemoteFile`, and I think that's what we should use here. For example:
```suggestion
    ->withImageFile(new InlineFile('image/png', $base64blob))
```

If you think there should be a variant that doesn't require instantiating a class like this, then I think this method will need to have a more concrete name, e.g. `withInlineImage`.

docs/ARCHITECTURE.md
```diff
@@ -116,6 +172,17 @@ $text = AiClient::generateTextResult(
...
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Can you repeat that please?')
+    ->withHistory(
+        new UserMessage('Do you spell it WordPress or Wordpress?'),
+        new AgentMessage('The correct spelling is WordPress.')
+    )
+    ->generateText();
+```
```
felixarntz: Just a thought: Out of all code examples, this is the only one where I find the fluent API one slightly worse than the traditional one. It's still shorter, but the fact that the order of the messages is not entirely sequenced is confusing to me (the order of the messages in the code here is "3, 1, 2" instead of "1, 2, 3").

Should we maybe provide a separate entry point in this case, like `AiClient::promptHistory`? Or alternatively, we leave out the string parameter of `AiClient::prompt` here (it needs to be optional anyway), and we could instead specify all three messages including the new one in the `withHistory` call?

docs/ARCHITECTURE.md
```diff
@@ -139,6 +206,21 @@ $text = AiClient::generateTextResult(
 
 _Note: Unlike the previous two examples, to require JSON output it is necessary to go the verbose route, since it is impossible for the SDK to detect whether you require JSON output purely from the prompt input. Therefore this code example contains the logic to manually search for suitable models and then use one of them for the task._
 
+##### Fluent API
+```php
+// Verbose.
+$text = AiClient::prompt('Transform the following CSV content into a JSON array of row data.')
+    ->asJsonResponse()
+    ->usingOutputSchema(['name' => 'string', 'age' => 'integer'])
```
felixarntz: I acknowledge that this might be a nice idea to "shorten" the more verbose structure of JSON schema, but that has nothing to do with whether the API is fluent or not, so I think it's misleading to use it here.

Let's put the same proper JSON schema array in here that is used in the traditional API call below. We can separately think later about a short-hand version of it.
```suggestion
    ->usingOutputSchema([
        'type'  => 'array',
        'items' => [
            'type'       => 'object',
            'properties' => [
                'name' => [
                    'type' => 'string',
                ],
                'age'  => [
                    'type' => 'integer',
                ],
            ],
        ],
    ])
```

docs/ARCHITECTURE.md
```diff
@@ -139,6 +206,21 @@ $text = AiClient::generateTextResult(
...
+    ->generateText();
+
+// Simple.
+$text = AiClient::prompt('Transform the following CSV content into a JSON array of row data.')
+    ->asJsonResponse(['name' => 'string', 'age' => 'integer'])
+    ->generateText();
```
felixarntz: IMO, we can skip this. Maybe a bit too much nuance for this list of examples to clarify here that you could also omit the `asJsonResponse()`. To be clear, I agree this should be possible, but not worth including here.

Also, if we kept this, this would also need to be updated to use proper JSON schema, in which case the `Simple` annotation doesn't feel entirely right anymore :)

docs/ARCHITECTURE.md
```diff
@@ -217,19 +306,34 @@ classDiagram
 direction LR
     namespace Ai {
         class AiClient {
-            +prompt(?string $text) PromptBuilder$
-            +message(?string $text) MessageBuilder$
+            +prompt(...string $text) PromptBuilder$
+            +message(...string $text) MessageBuilder$
```
felixarntz: Why should these support multiple text strings?

docs/ARCHITECTURE.md
```diff
@@ -217,19 +306,34 @@ classDiagram
...
         }
 
         class PromptBuilder {
             +withText(string $text) self
+            +withImage(string $mimeType, string $base64Blob) self
```
felixarntz: See my comment above about this method name being too vague.

docs/ARCHITECTURE.md
```diff
@@ -217,19 +306,34 @@ classDiagram
...
+            +usingModelSupportingHistory() self
+            +usingModelSupportingEmbeddings() self
+            +usingModelSupportingImages() self
+            +usingModelSupportingJsonOutput() self
+            +usingModelSupportingMusic() self
+            +usingModelSupportingOutputSchema() self
+            +usingModelSupportingSpeech() self
+            +usingModelSupportingText() self
+            +usingModelSupportingTextToSpeech() self
+            +usingModelSupportingVideo() self
```
felixarntz: See my comments above, I'm not sure these methods are needed since the required model capabilities and options are always inferred from the prompt. There is already a way to verbosely handle this "manually", IMO no need to overcomplicate the fluent API for something that's unnecessary.

docs/ARCHITECTURE.md
```diff
@@ -240,24 +344,41 @@ direction LR
             +usingOutputMime(string $mimeType) self
             +usingOutputSchema(array< string, mixed > $schema) self
             +usingOutputModalities(...AiModality $modalities) self
+            +asArrayResponse(?array< string, mixed > $schema) self
```
felixarntz: I think we should remove this - the response isn't necessarily an array in the sense of JSON - it'll often be a JSON object. In PHP that's still an (associative) array, but still IMO this is a confusing name. Since there's already `asJsonResponse`, I don't see a need for this alternative anyway.

docs/ARCHITECTURE.md
```diff
@@ -240,24 +344,41 @@ direction LR
...
+            +generateEmbeddingsResults(int $candidateCount) EmbeddingResult[]
             +generateTextOperation() GenerativeAiOperation
+            +generateTextOperations(int $candidateCount) GenerativeAiOperation[]
             +generateImageOperation() GenerativeAiOperation
+            +generateImageOperations(int $candidateCount) GenerativeAiOperation[]
             +convertTextToSpeechOperation() GenerativeAiOperation
             +generateSpeechOperation() GenerativeAiOperation
+            +generateSpeechOperations(int $candidateCount) GenerativeAiOperation[]
             +generateEmbeddingsOperation() EmbeddingOperation
+            +generateEmbeddingsOperations(int $candidateCount) EmbeddingOperation[]
```
felixarntz: Per my previous comment on the `generateTexts(4)` example, please keep in mind that the result will _always_ be a single `GenerativeAiResult` or a single `GenerativeAiOperation`. It's the _candidates_ within these objects that would be affected by the number of how many candidates you request.

In other words, all these `*Results()` and `*Operations()` methods would be invalid.

For the short-hand methods below (like `generateTexts()`) I like your idea though. Since the return value there is always the fully parsed result, having a plural short-hand to not have to separately call `usingCandidateCount(4)` makes a lot of sense to me.

docs/ARCHITECTURE.md
```diff
@@ -266,6 +387,7 @@ direction LR
         class MessageBuilder {
             +usingRole(MessageRole $role) self
             +withText(string $text) self
+            +withImage(string $mimeType, string $base64Blob) self
```
felixarntz: See above.

felixarntz (changes requested):
@borkweb Thank you for the PR! Overall the examples look great, and they very much clarify how to use the fluent API and how the two APIs differ.

I think there are a few things here that need to be revised, please see my comments below.

Note: For the class diagrams, I didn't leave all the same comments again on the duplicate instances of `AiClient`, `PromptBuilder`, and `MessageBuilder`, but please make sure to keep them in sync after making changes. :)

docs/ARCHITECTURE.md
borkweb: Ohhh, that makes sense. That was a fundamental misunderstanding on my part how the `usingCandidateCount(4)` would behave. I'll update the return types to reflect the actual return type as you mentioned above.

docs/ARCHITECTURE.md
borkweb: Aye, keeping `usingCandidateCount()` is what I was going for here as a shorthand of the above. Your comment in the class diagram clarifies the return type mixup I had.

docs/ARCHITECTURE.md
borkweb: You make a good point here! The termination methods themselves would add their requirements to the overall request, performing the above logic under the hood. I'll update the docs as you are suggesting for that and remove those `usingModelSupporting*()` methods.

docs/ARCHITECTURE.md
borkweb: As an example, yeah, that can read strangely (though it doesn't to me because it is a collection of instantiations within the `withHistory()` method). I think the following would also be valid:

```php
$text = AiClient::prompt()
    ->withHistory(
        new UserMessage('Do you spell it WordPress or Wordpress?'),
        new AgentMessage('The correct spelling is WordPress.'),
        new UserMessage('Can you repeat that please?')
    )
    ->generateText();
```

or

```php
$text = AiClient::prompt()
    ->withHistory(
        new UserMessage('Do you spell it WordPress or Wordpress?'),
        new AgentMessage('The correct spelling is WordPress.'),
    )
    ->withText('Can you repeat that please?')
    ->generateText();
```

or

```php
$history = [
    new UserMessage('Do you spell it WordPress or Wordpress?'),
    new AgentMessage('The correct spelling is WordPress.'),
];

$text = AiClient::prompt('Can you repeat that please?')
    ->withHistory(...$history)
    ->generateText();
```

I kind of like that last one the most.

docs/ARCHITECTURE.md
borkweb: I do like the simplicity of not requiring an instantiation for inline images like you are suggesting. Perhaps there should be something similar for all of the image types as well?

* `withLocalImage(string $path)`
* `withRemoteImage(string $uri)`

If handling the other image types similar to the above doesn't feel like a good idea, I'm inclined to stick with `new InlineFile()`!

docs/ARCHITECTURE.md
borkweb: As a shorthand for `withMessage()` similar to the fluent embeddings example I was providing where each text string would be a `UserMessage()` under the hood.

JasonTheAdams:
> I think the decision on Thursday was to go with both, in order to compare developer interest/usage in them and eventually decide on the final approach based on that.

Ah, then I misread what you were saying [in Slack](https://wordpress.slack.com/archives/C08TJ8BPULS/p1752166554115059). I interpreted that as:
1. Build this with a Fluent API
2. Assess how it would be to add a "traditional" API that uses the builder under the hood
3. Determine whether we do it now or later

docs/ARCHITECTURE.md
```diff
@@ -19,6 +19,13 @@ The following examples indicate how this SDK could eventually be used.
 
 #### Generate text using any suitable model from any provider (most basic example)
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->generateText();
```
JasonTheAdams, line 25:
```suggestion
$text = AiClient::prompt('Write a 2-verse poem about PHP.')->generateText();
```
I'm tempted to only break the chaining if we don't immediately terminate the builder. A single line just shows how awesome simple it is to do a basic prompt. Not a big deal either way, just a thought. 😄

docs/ARCHITECTURE.md
JasonTheAdams: I like these convenience methods!

docs/ARCHITECTURE.md
JasonTheAdams: I like the idea of the `AiClient::prompt()` method's text parameter being optional. It's effectively just applying `withText()` immediately for readability in most cases. So being able to begin a blank prompt feels nice and, per the examples above, gives the flexibility to construct the prompt however makes most sense to the dev. Makes the diversity of options feel like a feature not a bug. 😄

docs/ARCHITECTURE.md
JasonTheAdams: Ah yeah, I've wondered about this. Makes me wonder if 	`$text` should be a union of `string|Message` so that it does match `withHistory` and can be used that way. 🤔

docs/ARCHITECTURE.md
borkweb: Agreed! The spec that @felixarntz laid out already supports an empty `prompt()` and making it variadic still allows for that 🥇

JasonTheAdams: Do we want to include option 2 in the examples? Maybe even contrast the way we have now? I like having an example that shows the prompt text parameter as optional.

docs/ARCHITECTURE.md
felixarntz: IMO keeping separate lines is better for readability.

docs/ARCHITECTURE.md
felixarntz: Agreed, it would be nice to have those 3 convenience wrappers to not have to instantiate one of the image classes. Under the hood, these wrappers could do that. Just to clarify, I did like your original idea (in addition to the class based `with*File()` methods), but meant to suggest they become wrappers.

So +1 to add:
* `withInlineImage(string $base64blob, $mimeType)`
* `withLocalImage(string $path, $mimeType)`
* `withRemoteImage(string $uri, $mimeType)`

I'd suggest putting the MIME type last, because that is not the "key" piece of data, and in some cases it could also be optional. I see in your previous comment you omitted the `$mimeType`, but I think we should keep it. Yes, in principle it can typically be inferred from the given path or URI (file extension), but that's not guaranteed. FWIW if someone provides a data URI instead of the pure base64 data for `$base64blob`, even there the MIME type could be inferred.

Maybe we can make `$mimeType` an optional parameter on all three of these?

docs/ARCHITECTURE.md
```diff
@@ -98,6 +153,14 @@ $text = AiClient::generateTextResult(
 
 _Note: Since this omits the model parameter, the SDK will automatically determine which models are suitable and use any of them, similar to [the first code example](#generate-text-using-any-suitable-model-from-any-provider-most-basic-example). Since it knows the input includes an image, it can internally infer that the model needs to not only support `AiCapability::TEXT_GENERATION`, but also `AiOption::INPUT_MODALITIES => ['text', 'image']`._
 
+##### Fluent API
+```php
+$text = AiClient::prompt('Generate alternative text for this image.')
+    ->withImageFile(new InlineFile('image/png', $base64blob))
```
felixarntz: Based on https://github.com/WordPress/php-ai-client/pull/12#discussion_r2203525309, if we go with that, could you reverse the order here accordingly too?

docs/ARCHITECTURE.md
```diff
@@ -254,10 +360,14 @@ direction LR
             +generateSpeechOperation() GenerativeAiOperation
             +generateEmbeddingsOperation() EmbeddingOperation
             +generateText() string
+            +generateTexts(int $candidateCount) string
```
felixarntz: This will return an array of strings, for each individual candidate:
```suggestion
            +generateTexts(int $candidateCount) string[]
```

docs/ARCHITECTURE.md
```diff
@@ -254,10 +360,14 @@ direction LR
...
             +streamGenerateText() Generator< string >
             +generateImage() File
+            +generateImages(int $candidateCount) File
```
felixarntz: See above:
```suggestion
            +generateImages(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
```diff
@@ -254,10 +360,14 @@ direction LR
...
             +convertTextToSpeech() File
+            +convertTextToSpeeches(int $candidateCount) File
```
felixarntz: See above:
```suggestion
            +convertTextToSpeeches(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
```diff
@@ -254,10 +360,14 @@ direction LR
...
             +generateSpeech() File
+            +generateSpeeches(int $candidateCount) File
```
felixarntz: See above:
```suggestion
            +generateSpeeches(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
```diff
@@ -408,10 +519,14 @@ direction LR
             +generateSpeechOperation() GenerativeAiOperation
             +generateEmbeddingsOperation() EmbeddingOperation
             +generateText() string
+            +generateTexts(int $candidateCount) string
```
felixarntz:
```suggestion
            +generateTexts(int $candidateCount) string[]
```

docs/ARCHITECTURE.md
```diff
@@ -408,10 +519,14 @@ direction LR
...
             +streamGenerateText() Generator< string >
             +generateImage() File
+            +generateImages(int $candidateCount) File
```
felixarntz:
```suggestion
            +generateImages(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
```diff
@@ -408,10 +519,14 @@ direction LR
...
             +convertTextToSpeech() File
+            +convertTextToSpeeches(int $candidateCount) File
```
felixarntz:
```suggestion
            +convertTextToSpeeches(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
```diff
@@ -408,10 +519,14 @@ direction LR
...
             +generateSpeech() File
+            +generateSpeeches(int $candidateCount) File
```
felixarntz:
```suggestion
            +generateSpeeches(int $candidateCount) File[]
```

docs/ARCHITECTURE.md
felixarntz: +1, maybe simply replace this fluent API example with option 2? I don't think we need to keep both, and agree option 2 would show off the flexibility of the fluent API.

docs/ARCHITECTURE.md
felixarntz: A few thoughts:
* I think for `message()` this shouldn't be variadic, as the `MessageBuilder`'s purpose is to build a single message. It's supplementary to e.g. using `withHistory()`, so you could construct each message using `MessageBuilder`.
    * This makes me think: If we wanted to make it easier to construct histories, I wonder whether a `HistoryBuilder` would make sense. Or alternatively to `withHistory()`, you could do multiple `addMessage()` calls, and this method could accept a single message in a more flexible way (like I'm brainstorming for `prompt()` also in my last bullet point below).
* For `prompt()`, I'm not strongly opposed of this being variadic, but I'm also not fully convinced it's needed. If we wanted to support that for the entrypoint, maybe a separate entrypoint would be better?
    * Thinking out loud `historyPrompt()` or something like that? It would also return a `PromptBuilder`, but with a different entrypoint.
* For the data type, I agree with @JasonTheAdams that ideally it shouldn't only allow for `string`. That seems unnecessarily opinionated. Yes, a single string as a message is most certainly the most common, but it still feels wrong to center the API on only that. So I'd agree that at a minimum this should be `string|Message`.
    * We could also consider being _extremely_ flexible here and make this a super simple entrypoint. Throw a string at it? Sure, we parse it into a single-string message. Throw a `File` instance at it? Sure, we parse it into a single-file message. Throw an array of `MessagePart`s at it? Sure, we parse it into a message consisting of these parts. Or, you pass the entire `Message` object to it yourself. Curious what you think about this. Not a must-have to me, but worth thinking about.

I think for now, going with `string|Message` seems good enough for me to unblock this. We can separately think about further finetuning later IMO, but happy to discuss here.

docs/ARCHITECTURE.md
borkweb: Consider it done!

docs/ARCHITECTURE.md
borkweb: ### `message()`

The point of this method's purpose of building a _single_ message makes sense and multiple `$text` could be considered out of place. My original thought was to collapse the provided strings into a single message (like, if there are multiple variables that contain strings that the dev would want to make up a single message). I jumped to that use case due to the presence of the `withText()` in the `MessageBuilder`. Because technically a user could do the following (assuming `withText()` is additive as I assume it is):

```php
// Maybe these calls are grabbing from user submitted data.
$name = blahGetName();
$task = blahGetTaskToPerform();

$message = AiClient::message('My name is:')
    ->withText($name)
    ->withText('Based on my name, write a poem about:')
    ->withText($task)
    ->get();
```

Being variadic, the same could be done like so:

```php
$inputs = [
    'My name is:',
    blahGetName(),
    'Based on my name, write a poem about:',
    blahGetTaskToPerform(),
];

$message = AiClient::message(...$inputs)
    ->get();

// Of course, the same could be done like so:

$message = AiClient::message(implode("\n", $inputs))
    ->get();
```

All that being said, the case for the arguments for `message()` being super simple is a good case and we have `withText()` to handle the more complicated use case. 

**I'm not super passionate about whether we have variadic arguments for `message()`** so we can roll with whatever. `withText()` is, arguably, the more intuitive implementation anyway.

### `prompt()`

Similar to `message()` I could go either way and the case for `prompt()` being variadic is just as loose as `message()`. A `historyPrompt()` entry point could be interesting with a `HistoryBuilder` but that definitely feels like something we should stew on a bit before implementing. My gut reaction right now is that an array of messages is easier to build/understand than learning how a `HistoryBuilder` would/could work.

### Data type

Ooo, that's a really good point on being able to throw strings, Files, Messages, MessageParts etc at it. I kind of like it!

---

> I think for now, going with string|Message seems good enough for me to unblock this. We can separately think about further finetuning later IMO, but happy to discuss here.

Agreed, let's do `string|Message`; and keep `prompt()` variadic; and `message()` NOT variadic for now (updated already) so that we don't keep this PR in limbo 👍

felixarntz: > Agreed, let's do `string|Message`; and keep `prompt()` variadic

The one remaining concern I have with that is with the `string` piece being variadic, as it's ambiguous to what that would do:

* Does it create a single `Message` under the hood with all these strings as text parts?
* Or does it create multiple `Message`s under the hood, with each string being the sole text of each message? If so, how would it handle the respective message role? Automatically alternate between `USER` and `MODEL`?

I don't _love_ that ambiguity.

Of course for the `Message` type, making this variadic would be perfectly fine. But for `string` I'm not sure for the above reason. On the other hand, if we decide multiple `string`s will be used to construct a single message, then it would be confusing to have `Message` variadic. What do you think about this ambiguity? And which of the two things would you say it _should_ do if we keep it that way?

An alternative would be to keep `prompt()` non-variadic, but have a separate entrypoint focused on prompt with a history, which would be variadic. For that, it would be more obvious that every parameter would translate to another message.

docs/ARCHITECTURE.md
```diff
@@ -217,12 +321,15 @@ classDiagram
 direction LR
     namespace Ai {
         class AiClient {
-            +prompt(?string $text) PromptBuilder$
-            +message(?string $text) MessageBuilder$
+            +prompt(string|Message ...$text) PromptBuilder$
+            +message(string $text) MessageBuilder$
```
felixarntz: Regardless of what we decide for `prompt()` (see https://github.com/WordPress/php-ai-client/pull/12#discussion_r2205325982), I think for `message()` the parameter needs to remain optional, as not every message needs to start with a text, plus there's `withText()` as an alternative.
```suggestion
            +message(?string $text) MessageBuilder$
```

docs/ARCHITECTURE.md
```diff
@@ -356,8 +468,8 @@ classDiagram
...
```
felixarntz: See above:
```suggestion
            +message(?string $text) MessageBuilder$
```

docs/ARCHITECTURE.md
```diff
@@ -254,10 +362,14 @@ direction LR
...
             +generateEmbeddingsOperation() EmbeddingOperation
             +generateText() string
+            +generateTexts(int $candidateCount) string[]
             +streamGenerateText() Generator< string >
             +generateImage() File
+            +generateImages(int $candidateCount) File[]
             +convertTextToSpeech() File
+            +convertTextToSpeeches(int $candidateCount) File[]
             +generateSpeech() File
+            +generateSpeeches(int $candidateCount) File[]
```
felixarntz: Should we clarify the `int $candidateCount` as optional, i.e. `?int $candidateCount`? If someone already went with `usingCandidateCount($candidateCount)` in the builder before, there wouldn't be a need to specify it here again. In other words, it's only a short-cut, as discussed before.

On the other hand, if you didn't already use `usingCandidateCount`, it would be weird not to specify it here. Although, that's kind of similar if you did `generateTextResult()->candidates()` without having specified a candidate count. I think if you don't specify it, we'd just stick with the default of 1.

WDYT?

docs/ARCHITECTURE.md
```diff
@@ -408,10 +524,14 @@ direction LR
...
```
felixarntz: See above.

felixarntz (commented):
A few last follow up questions, but no real blockers. The key thing to decide on still is https://github.com/WordPress/php-ai-client/pull/12#discussion_r2205325982.

docs/ARCHITECTURE.md
borkweb: I think documentation helps a lot with removing ambiguity (not just architecture documentation) but documenting the builders themselves.

I was imagining a single Message would be constructed, but I guess this begs the question on how we plan to handle:

```php
AiClient::prompt('something')
    ->withText('something else')
    ->withText('something else entirely')
    ->generateText();
```

**The answer to that is likely the same answer we'd have for the handling of variadic `strings`.** My original thought was that given a collection of user-provided strings (either in a prompt or `withText()`) it all becomes a single `UserMessage()` under the hood.

However, I see where we'd have to make a decision if someone did something like:

```php
AiClient::prompt('yo', new Message('banana pancakes'), 'some people are squirrel handed')
    ->withText('i like butter')
    ->generateText();
```

With that all typed out, I see your point about variadic argument for `prompt()` being confusing to deal with as soon as we introduce/allow `Message` and `File` as well.

## TL;DR

I've come around to your way of thinking on `prompt()` NOT being variadic. But, the first snippet will need some discussion, I think.

docs/ARCHITECTURE.md
borkweb: I love that change. Optional where we return an array of 1 item if omitted and `usingCandidateCount()` was not specified.

docs/ARCHITECTURE.md
felixarntz: I agree however `prompt()` would handle variadic values would need to align with calling `withText()` multiple times (or also `withInlineImage()` etc., if/once we supported the `File` parameter type in `prompt()`).

+1 for now to keep the entire `PromptBuilder` focused on creating a single message, with the only API surface for multiple messages being `withHistory()`. I think that keeps things clear.

We can see how well `withHistory()` works in practice in combination with `MessageBuilder` or simply passing an array of manually constructed `Message` objects, and go from there. I'm sure there's ways to simplify that more, but it could be added later as we get a better sense of how people use it. Maybe there could be a `HistoryBuilder` that creates an array of messages - just thinking out loud. Anyway, let's think about it separately - maybe we can discuss further in a new issue.

felixarntz:
For reference, sharing here the directly related discussion that happened on Slack: https://wordpress.slack.com/archives/C08TJ8BPULS/p1752512797654719

docs/ARCHITECTURE.md
```diff
@@ -81,6 +119,23 @@ $imageFile = AiClient::generateImageResult(
...
+);
+
+$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->withModel(
+        AiClient::defaultRegistry()->getProviderModel(
+            $providerModelsMetadata[0]->getProvider()->getId(),
+            $providerModelsMetadata[0]->getModels()[0]->getId()
+        )
+    )
+    ->generateText();
```
JasonTheAdams, line 135: We can simplify this nicely!
```suggestion
$text = AiClient::prompt('Write a 2-verse poem about PHP.')
    ->withModelRequirements(AiCapability::TEXT_GENERATION)
    ->generateText();
```

JasonTheAdams (changes requested):

docs/ARCHITECTURE.md
borkweb: We continued the discussion [in Slack](https://wordpress.slack.com/archives/C08TJ8BPULS/p1752512797654719) for quicker back and forths!

docs/ARCHITECTURE.md
felixarntz: @JasonTheAdams I commented before that this is explicitly in a section that shows what happens under the hood / how to do it more verbosely. Since that's the intention here, I don't think we should simplify it again.

borkweb: That's related to this thread: https://github.com/WordPress/php-ai-client/pull/12#discussion_r2202918705

felixarntz (approved):
@borkweb Thanks for bearing with me here - this looks excellent! 🎉

docs/ARCHITECTURE.md
JasonTheAdams: Ahhh, I see. That's true. I'd like to change the wording in the example itself. I can do that in a subsequent PR.

JasonTheAdams (approved):
It's so fluent and beautiful! 🥹

PR #13, opened 2025-07-12
Simplify architecture usage down to the Fluent API

borkweb:
This is a presumptive change based on [the comment](https://github.com/WordPress/php-ai-client/pull/12#issuecomment-3065836517) by @JasonTheAdams in https://github.com/WordPress/php-ai-client/pull/12 regarding a potential decision to just go fluent. If that is the intent, then this PR is stripping out the Traditional approach for the implementer API.

**Note:** because I cannot push branches to this repository, this PR needs to be against the `add/architecture-definition` branch when, ideally, it would be against #12. (But a PR against my own fork isn't very helpful 😄).

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -19,6 +19,13 @@ The following examples indicate how this SDK could eventually be used.
 
 #### Generate text using any suitable model from any provider (most basic example)
 
+##### Fluent API
+```php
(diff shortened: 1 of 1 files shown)


felixarntz:
Thank you for the PR @borkweb! As per my reply in https://github.com/WordPress/php-ai-client/pull/12#issuecomment-3066039523, I think this is a bit premature since we'll want to assess both APIs to see what developers prefer. For this reason, I'm going to close this PR.

PR #14, opened 2025-07-12
Add additional glossary terms

borkweb:
Adds definitions for:
- Candidate Count
- Capability
- Extender API
- Implementer API
- Option/AI Option
- Registry/Provider Registry

### DIFF
--- a/docs/GLOSSARY.md
+++ b/docs/GLOSSARY.md
@@ -4,7 +4,7 @@ This glossary defines common terms relevant for the PHP AI Client and related pr
 
 * **Agent**: An autonomous system that can perceive its environment, make decisions, and take actions to achieve specific goals, often leveraging AI models.
 * **Candidate Count**: The number of different response options an LLM generates internally.
-* **Capability**: A specific skill, function, or type of task that an AI model can perform.
+* **Capability**: A specific skill, function, or type of task that an AI model can perform, e.g. text generation or image generation.
 * **Extender API**: The API used by developers that want to enable the use of additional _Providers_ or _Models_.
 * **Generative AI**: Overaching term describing AI models that generate content as requested in a prompt.
 * **Implementer API**: The API used by people that want to _implement_ AI features in their own software/products.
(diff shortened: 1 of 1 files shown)


docs/GLOSSARY.md
```diff
@@ -3,11 +3,17 @@
 This glossary defines common terms relevant for the PHP AI Client and related projects.
 
 * **Agent**: An autonomous system that can perceive its environment, make decisions, and take actions to achieve specific goals, often leveraging AI models.
+* **Candidate Count**: The number of different response options an LLM generates internally.
+* **Capability**: A specific skill, function, or type of task that an AI model can perform.
```
felixarntz: I think it would be good to include an example:
```suggestion
* **Capability**: A specific skill, function, or type of task that an AI model can perform, e.g. text generation or image generation.
```

docs/GLOSSARY.md
```diff
@@ -3,11 +3,17 @@
...
+* **Extender API**: The API used by developers that want to enable the use of additional _Providers_ or _Models_.
 * **Generative AI**: Overaching term describing AI models that generate content as requested in a prompt.
+* **Implementer API**: The API used by people that want to _implement_ AI features in their own software/products.
 * **MCP**: The "Model Context Protocol", a proposed standard for connecting AI assistants to the systems where data lives.
 * **Message**: A single message, either a user prompt, a model response, or a system prompt—optionally containing of multiple message parts.
 * **Message part**: A part of a message, such as a piece of text, a URL, a file, or a function call.
 * **Modality**: The type or format of input provided to, or output received from, an AI model. Examples include text, image, audio, and video.
 * **Model**: A specific AI model that supports arbitrary AI features and modalities. Examples include content generation, classification, embedding.
+* **Option** or **AI Option**: AI configuration and output options. Examples: `CANDIDATE_COUNT`, `MAX_TOKENS`, `TEMPERATURE`, `OUTPUT_SCHEMA`, etc.
```
felixarntz: I think we should choose whether we refer to this as "Option" or as "AI Option", not include both here. Related, we should then also use the same approach for "Capability" (or "AI Capability") above.

For now, I'll suggest the one without prefix (since almost all other terms here could otherwise also be prefixed with "AI").

For the examples, I'd prefer to keep things "colloquial" here, so that we don't mention specific identifiers here that may not exist in the codebase later, or that we would need to keep in sync.
```suggestion
* **Option**: AI configuration and output options for a model, e.g. temperature or output schema.
```

docs/GLOSSARY.md
```diff
@@ -3,11 +3,17 @@
...
 * **Prompt**: The input that a generative AI model uses to generate content—often text, but it can also be of other modalities.
 * **Provider**: An entity (company, organization, or platform) that offers access to one or more AI models or services via an API (e.g., Anthropic, Google, OpenAI, a locally hosted service).
+* **Registry** or **Provider Registry**: Manages the available AI _Providers_ and _Models_. Provides a mechanism to find models that match criteria for a given AI interaction.
```
felixarntz: See above, I'd say let's align on one term only.
```suggestion
* **Provider Registry**: Manages the available AI _Providers_ and _Models_. Provides a mechanism to find models that match criteria for a given AI interaction.
```

felixarntz (commented):
@borkweb Thanks for the PR, this looks great!

Just a few minor suggestions before merging.

felixarntz:
@borkweb One additional distinction between capabilities and options is that capabilities are always "just a string" whereas options can also have more concrete constraints in addition to the option name.

For example:
* For a capability like "text generation", you simply check: Does the model support this capability?
* For an option like "output MIME type", you check: Does the model support this option, and the concrete value specified?

The rationale for that distinction is that options may have different support between different models. For example, one model may support output of `image/png` and `image/jpeg`, while another model may support also `image/webp`. So checking _only_ whether specifying an output MIME type is supported in general would not be sufficient.

We don't necessarily have to clarify this distinction in the glossary, but it might be useful if we can keep it short. But in any case, I also wanted to share it here for context.

borkweb:
That additional context is great - thank you!

felixarntz (approved):
Great to have all these terms clarified and documented! 🎉

PR #15, opened 2025-07-12
Add .gitattributes for package distribution

borkweb:
Excludes development and documentation files from Composer package:
- Development configs (.editorconfig, .gitignore, .gitattributes)
- GitHub workflows
- Documentation folder
- Test files and configs
- Vendor directory

This is, of course, excluding things that are not yet part of the repository but will likely be part of it soon.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,7 @@
 /.github/              export-ignore
 .gitignore             export-ignore
 /*.md                  export-ignore
-/LICENSE.md            -export-ignore
+/LICENSE.md           -export-ignore
 /README.md            -export-ignore
 /docs/                 export-ignore
 /tests/                export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,5 +5,6 @@
 /docs/                 export-ignore
 /tests/                export-ignore
 /vendor/               export-ignore
+/phpcs.xml.dist        export-ignore
 /phpstan.neon.dist     export-ignore
 /phpunit.xml.dist      export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,9 @@
 .gitattributes         export-ignore
 /.github/              export-ignore
 .gitignore             export-ignore
+/*.md                  export-ignore
+/LICENSE.md            -export-ignore
+/README.md            -export-ignore
 /docs/                 export-ignore
 /tests/                export-ignore
 /vendor/               export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+.editorconfig          export-ignore
+.gitattributes         export-ignore
+/.github/              export-ignore
+.gitignore             export-ignore
+/docs/                 export-ignore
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
I'm a touch reluctant to preemptively include things like `/docs/`, as I'm not aware of that even being discussed yet. But others, like tests, are probably fine.

borkweb:
I can count on one hand the number of times I've personally looked at documentation in a vendor folder. That number is zero 😁 

Ultimately, I can modify it however we want - I was mimicking Guzzle's repo.

.gitattributes
```diff
@@ -0,0 +1,9 @@
+.editorconfig          export-ignore
+.gitattributes         export-ignore
+/.github/              export-ignore
+.gitignore             export-ignore
+/docs/                 export-ignore
+/tests/                export-ignore
+/vendor/               export-ignore
+/phpstan.neon.dist     export-ignore
```
felixarntz, line 12: What about PHPCS?
```suggestion
/phpcs.xml.dist        export-ignore
/phpstan.neon.dist     export-ignore
```

.gitattributes
```diff
@@ -0,0 +1,9 @@
+.editorconfig          export-ignore
+.gitattributes         export-ignore
+/.github/              export-ignore
+.gitignore             export-ignore
+/docs/                 export-ignore
```
felixarntz, line 8: If we exclude `docs`, shouldn't we also exclude any `.md` files in the root, other than `LICENSE.md` and `README.md`? At the moment, that's only `CONTRIBUTING.md`, but a wildcard exclusion with those two exceptions might work best.

felixarntz (commented):
Thanks @borkweb! Two points of feedback, plus a higher level question: What is the rationale for prefixing some root files with forward slashes while not prefixing others? What's the distinction?

borkweb:
> What is the rationale for prefixing some root files with forward slashes while not prefixing others? What's the distinction?

Any non prefixed item will apply to that file/directory _anywhere_ in the repo. Technically git observes `.gitignore` and `.gitattributes` within nested directories and the rules cascade down recursively. For `.editorconfig`, that gives a project the ability to set different standards in a different directory.

That being said, it is unlikely that those will occur in this repo and we can prefix if we want. I just left them unprefixed because of habit.

.gitattributes
```diff
@@ -0,0 +1,13 @@
+.editorconfig          export-ignore
+.gitattributes         export-ignore
+/.github/              export-ignore
+.gitignore             export-ignore
+/*.md                  export-ignore
+/LICENSE.md            -export-ignore
+/README.md            -export-ignore
```
felixarntz: Not sure about how you'd like to indent these, but it looks like it's off by 1 space on either one of these lines :)

felixarntz (approved):
@borkweb LGTM, just a tiny bit before merge.

.gitattributes
borkweb: Whoops! Gross. Fixed 🕺

PR #16, opened 2025-07-12
Add composer.json with PHP 7.4+ support, phpcs, phpstan, and phpunit

borkweb:
I made some assumptions in this PR that may be inaccurate.

1. I'm assuming we'll be using PHPStan. I'll PR a suggested `phpstan.neon.dist` in a follow-up PR.
2. ~Is `WordPress\AI` the correct namespace?~ Updated to `WordPress\AiClient`

The main bits:

- Configures package as wordpress/php-ai-client library
- Requires PHP 7.4 or higher
- Adds PHPUnit and PHPStan as dev dependencies and sets up initial configuration
- Sets up PSR-4 autoloading with WordPress\AI namespace
- Includes convenient scripts for testing and analysis
- Adds keywords for package discovery
- Add a dummy source file so phpstan and phpcs can be run

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "wordpress/php-ai-client",
-    "description": "A provider-agnostic PHP SDK for communicating with generative AI models",
+    "description": "A provider agnostic PHP AI client SDK to communicate with any generative AI models of various capabilities using a uniform API.",
     "license": "GPL-2.0-or-later",
     "type": "library",
     "keywords": [
--- a/composer.json
+++ b/composer.json
@@ -20,12 +20,12 @@
         "source": "https://github.com/WordPress/php-ai-client"
     },
     "require": {
-        "php": ">=7.4 || ^8.0",
+        "php": ">=7.4",
         "ext-json": "*"
     },
     "require-dev": {
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
-        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan": "~2.1",
         "phpunit/phpunit": "^9.5 || ^10.0",
         "squizlabs/php_codesniffer": "^3.7"
     },
--- a/composer.json
+++ b/composer.json
@@ -52,14 +52,13 @@
         "sort-packages": true
     },
     "scripts": {
-        "analyze": "phpstan analyze",
-        "check": [
-            "@test",
-            "@analyze",
-            "@cs"
+        "lint": [
+            "@phpcs",
+            "@phpstan"
         ],
-        "cs": "phpcs",
-        "cs-fix": "phpcbf",
-        "test": "phpunit"
+        "phpcbf": "phpcbf",
+        "phpcs": "phpcs",
+        "phpstan": "phpstan analyze",
+        "phpunit": "phpunit"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,10 @@
         "ext-json": "*"
     },
     "require-dev": {
+        "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
         "phpstan/phpstan": "^1.10",
-        "phpunit/phpunit": "^9.5 || ^10.0"
+        "phpunit/phpunit": "^9.5 || ^10.0",
+        "squizlabs/php_codesniffer": "^3.7"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
@@ -53,8 +55,11 @@
         "analyze": "phpstan analyze",
         "check": [
             "@test",
-            "@analyze"
+            "@analyze",
+            "@cs"
         ],
+        "cs": "phpcs",
+        "cs-fix": "phpcbf",
         "test": "phpunit"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,60 @@
+{
+    "name": "wordpress/php-ai-client",
+    "description": "A provider-agnostic PHP SDK for communicating with generative AI models",
+    "license": "GPL-2.0-or-later",
+    "type": "library",
+    "keywords": [
+        "ai",
+        "api",
+        "llm"
+    ],
+    "authors": [
+        {
+            "name": "WordPress AI Team",
+            "homepage": "https://make.wordpress.org/ai/"
+        }
+    ],
+    "homepage": "https://github.com/WordPress/php-ai-client",
+    "support": {
+        "issues": "https://github.com/WordPress/php-ai-client/issues",
+        "source": "https://github.com/WordPress/php-ai-client"
+    },
+    "require": {
+        "php": ">=7.4 || ^8.0",
+        "ext-json": "*"
+    },
+    "require-dev": {
+        "phpstan/phpstan": "^1.10",
+        "phpunit/phpunit": "^9.5 || ^10.0"
+    },
+    "minimum-stability": "dev",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -0,0 +1,65 @@
...
+    "require": {
+        "php": ">=7.4 || ^8.0",
+        "ext-json": "*"
+    },
+    "require-dev": {
+        "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
+        "phpstan/phpstan": "^1.10",
+        "phpunit/phpunit": "^9.5 || ^10.0",
+        "squizlabs/php_codesniffer": "^3.7"
+    },
```
felixarntz: Let's leave specific dependencies for another PR.

composer.json
```diff
@@ -0,0 +1,65 @@
...
+        "analyze": "phpstan analyze",
+        "check": [
+            "@test",
+            "@analyze",
+            "@cs"
+        ],
+        "cs": "phpcs",
+        "cs-fix": "phpcbf",
+        "test": "phpunit"
+    }
```
felixarntz: See above, let's leave these out until the PR that actually adds the relevant tooling completely.

composer.json
```diff
@@ -0,0 +1,65 @@
...
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "WordPress\\AI\\Tests\\": "tests/"
+        }
+    },
+    "config": {
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true
+        },
```
felixarntz, line 47: See above, let's add this separately in the PHPCS PR.

composer.json
```diff
@@ -0,0 +1,65 @@
+{
+    "name": "wordpress/php-ai-client",
+    "description": "A provider-agnostic PHP SDK for communicating with generative AI models",
```
felixarntz: Can we align this with the short description currently in `README.md`?
```suggestion
    "description": "A provider agnostic PHP AI client SDK to communicate with any generative AI models of various capabilities using a uniform API.",
```

composer.json
```diff
@@ -0,0 +1,65 @@
...
+    "autoload": {
+        "psr-4": {
+            "WordPress\\AI\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "WordPress\\AI\\Tests\\": "tests/"
+        }
+    },
```
felixarntz: The namespace will need to be something more concrete than just "AI", as this is only the AI _client_ SDK, not a general all-in-one AI SDK. I think `AiClient` makes most sense.
```suggestion
    "autoload": {
        "psr-4": {
            "WordPress\\AiClient\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "WordPress\\AiClient\\Tests\\": "tests/"
        }
    },
```

felixarntz (commented):
@borkweb Thanks for adding this, you beat me to it :D

Since the main purpose of this PR is to add `composer.json`, let's limit it to just that. The specific dependencies and their config should only be added together with their proper setup in other PRs (e.g. #17 that you already opened).

composer.json
borkweb: I've moved those dependencies into here based on your suggestion in #17

composer.json
```diff
@@ -0,0 +1,65 @@
...
+        },
+        "optimize-autoloader": true,
+        "platform": {
+            "php": "7.4"
+        },
+        "sort-packages": true
+    },
+    "scripts": {
+        "analyze": "phpstan analyze",
+        "check": [
```
felixarntz: Maybe better `lint`?
```suggestion
        "lint": [
```

composer.json
```diff
@@ -0,0 +1,65 @@
...
+            "dealerdirect/phpcodesniffer-composer-installer": true
+        },
+        "optimize-autoloader": true,
+        "platform": {
+            "php": "7.4"
+        },
+        "sort-packages": true
+    },
+    "scripts": {
+        "analyze": "phpstan analyze",
```
felixarntz: This may be a personal preference, but IMO for any scripts that purely execute a single specific tool it would be more intuitive to use the actual tool name for the script.

In other words, call the tool-specific scripts `phpstan`, `phpcs`, and `phpcbf`.

felixarntz (approved):
Looks great! Just two minor comments, but I'm preemptively approving.

Other than this, it would be great to add a GitHub workflow that runs these tools on every commit and for pull requests. Feel free to add it here, or we could do it in a follow up PR.

phpcs.xml.dist
```diff
@@ -0,0 +1,25 @@
...
+    <description>PSR-12 coding standards for PHP AI Client SDK</description>
+
+    <!-- Scan these files -->
+    <file>src</file>
+
+    <!-- Ignore vendor directory -->
+    <exclude-pattern>*/vendor/*</exclude-pattern>
+
+    <!-- Use PSR-12 standard -->
+    <rule ref="PSR12"/>
```
JasonTheAdams, line 12: As a note, here, we can encourage folks to use [PER-3.0](https://www.php-fig.org/per/coding-style/) in the CONTRIBUTING doc, as that extends PSR-12, but unfortunately [PHPCS doesn't support PER](https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/29).

composer.json
borkweb: Yeah, I see all kinds of approaches. In the projects that I've worked on most recently at https://github.com/stellarwp, I usually do `phpstan` and `phpcs` as well so I'm happy to make that change!

composer.json
borkweb: I can make that change, but removing `test`/`phpunit` is would make sense in that case

borkweb:
Let's do the workflows in another PR 🕺

composer.json
```diff
@@ -0,0 +1,64 @@
...
+            "homepage": "https://make.wordpress.org/ai/"
+        }
+    ],
+    "homepage": "https://github.com/WordPress/php-ai-client",
+    "support": {
+        "issues": "https://github.com/WordPress/php-ai-client/issues",
+        "source": "https://github.com/WordPress/php-ai-client"
+    },
+    "require": {
+        "php": ">=7.4 || ^8.0",
```
justlevine: Is the intent here to prevent PHP v9.x ? Not sure if this should be `^7.4 || ^8.0` or just `>=7.4`

composer.json
```diff
@@ -0,0 +1,64 @@
...
+        "ext-json": "*"
+    },
+    "require-dev": {
+        "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
+        "phpstan/phpstan": "^1.10",
```
justlevine: Any objections? We're already requiring PHP 7.4 as the minimum, and we don't even need to rewrite a config. 

```suggestion
        "phpstan/phpstan": "~2.1",
```

borkweb: No objections here - I'll make that change.

composer.json
borkweb: Haha not the intent. I'll simplify to `>=7.4`

JasonTheAdams (approved):
Thanks for putting this together, @borkweb!

composer.json
felixarntz: Good point. For consistency, I'd say it would be good to also have `test`, which for now would only invoke PHPUnit.

That way there are scripts for all:
* specific tools
* high-level workflows (lint and test)

PR #17, opened 2025-07-12
Add PHPStan and PHP CodeSniffer configuration

borkweb:
- Add phpstan.neon.dist with max level analysis
- Add phpcs.xml.dist with PSR-12 coding standards
- Configure both tools to scan src/ and tests/ directories
- Set up parallel processing and colored output for phpcs

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+<ruleset name="PHP AI Client Coding Standards">
+    <description>PSR-12 coding standards for PHP AI Client SDK</description>
+
+    <!-- Scan these files -->
+    <file>src</file>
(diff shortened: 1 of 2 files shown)


felixarntz (commented):
@borkweb Thanks!

See my related comments in #16 - I guess if we merge both PRs together, it's fine to keep all the `composer.json` changes in the other PR, but ideally they should be in here as the other PR doesn't add this tooling.

Other than that, this PR looks good to go. Maybe add a dummy PHP file in `src` with just a class and minimal test code, so that we can already verify that all of this works?

borkweb:
For simplicity's sake, I'll cherry-pick this into the other PR so we can do it all in one fell swoop. (along with the dummy file)

PR #22, opened 2025-07-14
Add PHP lint workflow

borkweb:
Adds GitHub Actions workflow to validate PHP syntax with PHPStan and PHPCS. The workflow runs on push and pull requests - the tools execute with the latest PHP version while PHPStan will validate the code against the version of PHP specified in `composer.json` (which is PHP 7.4).

(I used [WordPress/plugin-check](https://github.com/WordPress/plugin-check) for reference)

### DIFF
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -1,0 +1,59 @@
+name: PHP Code Linting
+
+on:
+  push:
+    branches:
+      - trunk
+      - 'feature/**'
+      - 'release/**'
+    # Only run if PHP-related files changed.
+    paths:
+      - '.github/workflows/php-lint.yml'
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpstan.neon.dist'
+      - 'composer.json'
+      - 'composer.lock'
+  pull_request:
+    # Only run if PHP-related files changed.
+    paths:
+      - '.github/workflows/php-lint.yml'
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpstan.neon.dist'
+      - 'composer.json'
+      - 'composer.lock'
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  php-lint:
+    name: PHP
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: 'latest'
+
+      - name: Validate Composer configuration
+        run: composer validate
+
+      - name: Install PHP dependencies
+        uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520
+        with:
+          composer-options: '--prefer-dist'
+
+      - name: PHPStan
+        run: composer phpstan
+
+      - name: PHP Code Sniffer
+        run: composer phpcs
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -48,7 +48,7 @@ jobs:
         run: composer validate
 
       - name: Install PHP dependencies
-        uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520
+        uses: ramsey/composer-install@v3
         with:
           composer-options: '--prefer-dist'
 


.github/workflows/php-lint.yml
```diff
@@ -0,0 +1,59 @@
+name: PHP Code Linting
+
+on:
+  push:
+    branches:
+      - trunk
+      - 'feature/**'
+      - 'release/**'
```
JasonTheAdams, line 8: @borkweb Any reason we wouldn't just lint for any branch? I'm not confident folks will follow the prefixed conventions here.

borkweb: I was matching what seems to be in place in a couple of other WordPress repos. I'm good with opening it up 👍

borkweb: That's section is just for `push` without a PR to save on action usage. The Pull Request section makes it run on any branch whether it is the original PR or a push (synchronization) of additional commits to that PR.

.github/workflows/php-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+    paths:
+      - '.github/workflows/php-lint.yml'
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpstan.neon.dist'
+      - 'composer.json'
+      - 'composer.lock'
+  pull_request:
+    # Only run if PHP-related files changed.
+    paths:
```
felixarntz, line 19: Somewhat related to https://github.com/WordPress/php-ai-client/pull/22/files#r2205986997, in other WordPress organization projects we also have these same branch restrictions under `pull_request`, to ensure that only pull requests _against_ those branches trigger the workflows.

I'm not strongly in favor of either way, but might be worth aligning. If you create some quick branch just to merge into some other secndary branch that still will require review anyway, I'd argue that running CI isn't needed.

.github/workflows/php-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: 'latest'
+
+      - name: Validate Composer configuration
+        run: composer validate
+
+      - name: Install PHP dependencies
+        uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520
```
felixarntz: Curious why this specific SHA version?

.github/workflows/php-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          composer-options: '--prefer-dist'
+
+      - name: PHPStan
+        run: composer phpstan
+
+      - name: PHP Code Sniffer
+        run: composer phpcs
```
felixarntz, line 59: We could instead invoke `composer lint`, to keep this simpler here, and be able to alter what "lint" means in a single place, instead of in two places. Just an idea.

felixarntz (approved):
@borkweb Thanks, looks great!

Just a few non-blocking thoughts.

.github/workflows/php-lint.yml
borkweb: I removed those branch restrictions under the `pull_request` section to help ensure that pull requesters (no matter where they are PRing into) have the burden of resolving issues rather than the target branch.

.github/workflows/php-lint.yml
borkweb: I got that directly from the `plugin-check` repo - I guessed it was for a reason and didn't dig in to learn why. :) I can go try to figure out why that's there and either remove or confirm that we should use it.

.github/workflows/php-lint.yml
borkweb: I find value in distinct steps that can fail independently in the GitHub Workflow so hunting down the problem area is simpler. Combining into `lint` would place the output of both commands into a single log.

felixarntz: Personally, I'd say it would still be easy to spot whether the error is from PHPCS or PHPStan, due to their different log output formatting and it all being sequenced under the one step.

But I'm okay to go with this if you prefer.

.github/workflows/php-lint.yml
felixarntz: Works for me 👍

.github/workflows/php-lint.yml
felixarntz: If we don't know why, I'd say let's try to pin the latest version instead - unless that would cause errors while this here works.

.github/workflows/php-lint.yml
borkweb: Cool - I'll stick with the separation as it keeps the failure point one fewer clicks away 🕺

.github/workflows/php-lint.yml
borkweb: Done!

Issue #21, opened 2025-07-14
Proposal: Use a Domain-Specific directory structure


borkweb:
I would like to propose a slight tweak to what we have cooking in #2 with regards to the organization of our classes, enums, interfaces, etc. If I am interpreting the namespaces correctly in that PR, we'd have a structure similar to:

```
src/
├── AiClient.php
├── MessageBuilder.php
├── PromptBuilder.php
|
├── Enums/
│   └── ...all of the non-provider enums
|
├── ProviderImplementations/
│   ├── Anthropic/
│   |   └── ...
│   ├── Google/
│   |   └── ...
│   └── OpenAI/
│       └── ...
|
├── Providers/
│   ├── Contracts/
│   |   └── ...all of the provider contracts
│   ├── Enums/
│   |   └── ...all of the provider enums
│   ├── Types/
│   |   └── ...all of the provider classes
|   └── Util/
│       └── ...provider utils
|
├── Types/
|   └── ...most of the classes and interfaces
|
└── Util/
    └── ...non-provider utils
```

## Proposed Structure: Domain-specific
A domain-specific structure is pretty good for reducing cognitive load in that it aligns nicely with mental models.

Here's what a domain-specific structure could look like:

```
src/
├── AiClient.php
|
├── Builders/
│   ├── MessageBuilder.php
│   └── PromptBuilder.php
│
├── Embeddings/
|   |
│   └── DTO/
│       └── Embedding.php
│
├── Files/
|   |
│   ├── Contracts/
│   |   └── File.php
|   |
│   └── DTO/
│       └── InlineFile.php
│       └── LocalFile.php
│       └── RemoteFile.php
|
├── ProviderImplementations/
│   ├── Anthropic/
│   |   └── ...
│   ├── Google/
│   |   └── ...
│   └── OpenAI/
│       └── ...
|
├── Messages/
|   |
│   └── DTO/
│   |   └── Message.php
│   |   └── MessagePart.php
|   |
│   └── Enums/
│       ├── MessageRole.php
│       └── MessagePartType.php
│
├── Operations/
|   |
│   ├── Contracts/
│   |   └── Operation.php
|   |
│   └── DTO/
│       └── EmbeddingOperation.php
│       └── GenerativeAiOperation.php
│
├── Providers/
│   ├── AiProviderRegistry.php
|   |
│   ├── Contracts/
│   |   ├── AiModelMetadataDirectory.php
│   |   ├── Authentication.php
[...]

Issue #20, opened 2025-07-13
Initial MVP task breakdown


felixarntz:
This issue proposes a breakdown of all the work based on the proposed architecture (see #2) to get to an MVP of this SDK, also taking into consideration the development roadmap (see #8).

What is called MVP here is largely equivalent with what is "Phase 1" in #8.

## Purpose

The primary purpose of this issue is to define the breakdown and later, once agreed upon, act as an overview issue to track the work until the MVP:

* The breakdown below is largely sequenced.
* Each checkbox should be considered an individual issue / pull request.
* Some things may be worked on at the same time, and if someone wants to tackle multiple subsequent steps in one session (without having to wait for PR approval), it's always okay to create a branch based on another branch for an unreviewed PR.
* Many tasks, especially in the "Foundation" section, are probably fine to start with a pull request directly. Some others are more complex and would likely benefit from some prior discussion in an issue. This can always be decided on a case-by-case basis.

Let's use the discussion in this issue to shape/refine this breakdown. As we discuss, iterations should be made directly on the issue description to keep things up to date in this source of truth.

### Foundation

_These almost entirely build on top of each other and therefore should mostly be sequenced._

* [x] Implement all enums for the implementer API and extender API (see #25 and #26)
    * See `AiClient.*.Enums` and `AiClient.Providers.*.Enums` namespaces in the class diagrams.
    * We won't be able to use PHP enums due to supporting PHP version support, but we can implement a very basic class foundation (e.g. `AbstractEnum`, with things like checking whether a given value is part of the enum) that effectively is our enum infrastructure.
* [x] Implement all data transfer objects for the implementer API, except omit any related to function calling and embeddings (see #28 and #30)
    * See `AiClient.*.DTO` namespaces in the class diagrams.
    * This also needs to include a few interfaces where appropriate, e.g. something to define the `getJsonSchema()` method commonly found.
    * Function calling will be implemented in a future milestone after the MVP.
* [x] Implement all data transfer objects for the extender API (see #31)
    * See `AiClient.Providers.DTO` and `AiClient.Providers.*.DTO` namespaces in the class diagrams.
* [x] Implement all interfaces (contracts) for the extender API, except the ones related to `HttpClient` and `Authentication` (see #35)
    * See `AiClient.Providers.Contracts` and `AiClient.Providers.*.Contracts` namespaces in the class diagrams.
[...]

PR #23, opened 2025-07-15
Update namespaces in architecture to domain-based

borkweb:
This PR is in draft mode while we contemplate #21.

Reorganized classes into more specific namespaces to match #21. Moved message-related classes, file types, models, and enums into their appropriate sub-namespaces to improve code organization and follow single responsibility principle.

You can see the [ARCHITECTURE.md here](https://github.com/WordPress/php-ai-client/blob/fix/namespace-references-in-architecture/docs/ARCHITECTURE.md).

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -21,13 +21,13 @@ The following examples indicate how this SDK could eventually be used.
 
 ##### Fluent API
 ```php
-$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+$text = Client::prompt('Write a 2-verse poem about PHP.')
     ->generateText();
 ```
 
 ##### Traditional API
 ```php
-$text = AiClient::generateTextResult(
+$text = Client::generateTextResult(
     'Write a 2-verse poem about PHP.'
 )->toText();
 ```
@@ -36,14 +36,14 @@ $text = AiClient::generateTextResult(
 
 ##### Fluent API
 ```php
-$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+$text = Client::prompt('Write a 2-verse poem about PHP.')
     ->usingModel(Google::model('gemini-2.5-flash'))
     ->generateText();
 ```
 
 ##### Traditional API
 ```php
-$text = AiClient::generateTextResult(
+$text = Client::generateTextResult(
     'Write a 2-verse poem about PHP.',
     Google::model('gemini-2.5-flash')
 )->toText();
@@ -53,18 +53,18 @@ $text = AiClient::generateTextResult(
 
 ##### Fluent API
 ```php
-$texts = AiClient::prompt('Write a 2-verse poem about PHP.')
+$texts = Client::prompt('Write a 2-verse poem about PHP.')
     ->usingModel(Anthropic::model('claude-3.7-sonnet'))
     ->generateTexts(4);
 ```
 
 ##### Traditional API
 ```php
-$texts = AiClient::generateTextResult(
+$texts = Client::generateTextResult(
     'Write a 2-verse poem about PHP.',
     Anthropic::model(
         'claude-3.7-sonnet',
-        [AiOption::CANDIDATE_COUNT => 4]
+        [OptionEnum::CANDIDATE_COUNT => 4]
     )
 )->toTexts();
 ```
@@ -73,20 +73,20 @@ $texts = AiClient::generateTextResult(
 
 ##### Fluent API
 ```php
-$imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
+$imageFile = Client::prompt('Generate an illustration of the PHP elephant in the Carribean sea.')
     ->usingProvider('openai')
     ->generateImage();
 ```
 
 ##### Traditional API
 ```php
-$modelsMetadata = AiClient::defaultRegistry()->findProviderModelsMetadataForSupport(
+$modelsMetadata = Client::defaultRegistry()->findProviderModelsMetadataForSupport(
     'openai',
-    new AiModelRequirements([AiCapability::IMAGE_GENERATION])
+    new ModelRequirements([CapabilityEnum::IMAGE_GENERATION])
 );
-$imageFile = AiClient::generateImageResult(
+$imageFile = Client::generateImageResult(
     'Generate an illustration of the PHP elephant in the Carribean sea.',
-    AiClient::defaultRegistry()->getProviderModel(
+    Client::defaultRegistry()->getProviderModel(
         'openai',
         $modelsMetadata[0]->getId()
     )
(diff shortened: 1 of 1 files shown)


JasonTheAdams:
@borkweb Regarding #21, are we just waiting for @felixarntz's response to https://github.com/WordPress/php-ai-client/issues/21#issuecomment-3071373631?

borkweb:
I was going to wait for the 👍 from both of you. There's also the question at the bottom of that regarding prefixes and suffixes that are a bit redundant. I'm not sure how y'all feel about that piece.

felixarntz:
I left some feedback in #21, let's iron that out first. Then we can update this PR as needed and review the details.

JasonTheAdams (approved):
Looking good!

docs/ARCHITECTURE.md
```diff
@@ -21,13 +21,13 @@ The following examples indicate how this SDK could eventually be used.
 
 ##### Fluent API
 ```php
-$text = AiClient::prompt('Write a 2-verse poem about PHP.')
+$text = Client::prompt('Write a 2-verse poem about PHP.')
```
felixarntz: Hmm, on this one I'm not sure whether we should drop the `Ai` prefix because it's the main entrypoint, and having to use `Client` as a name may easily conflict with other things, like an HTTP client.

I realize now that when I said only the `AiClient` namespace can retain the prefix, I probably meant to say the namespace and the class.

But curious to hear your thoughts on this.

docs/ARCHITECTURE.md
```diff
@@ -293,12 +293,14 @@ config:
 ---
 classDiagram
 direction LR
-    namespace AiClientNamespace {
-        class AiClient {
+    namespace AiClient {
+        class Client {
```
felixarntz: See above (in case we decide to go back to `AiClient`).

docs/ARCHITECTURE.md
```diff
@@ -523,79 +527,141 @@ direction LR
...
+            ERROR
+        }
+        class OperationStateEnum {
+            STARTING
+            PROCESSING
+            SUCCEEDED
+            FAILED
+            CANCELED
+        }
+    }
```
felixarntz: It's a good mental cross-check to see that this namespace currently remains relevant for the Implementer API (as this class diagram is only that part).

If we are going to be strict about it, we should probably move this to another namespace outside of `AiClient.Providers`. Since modalities are definitely relevant for implementers (as they can directly reference it when creating a prompt), and `FinishReason` belongs to `Result`, and `OperationStateEnum` belongs to `Operation`, I'd say it makes sense to move this to other `Enums` namespaces outside of `AiClient.Providers`.

docs/ARCHITECTURE.md
```diff
@@ -890,46 +902,32 @@ direction LR
...
         }
-        class AiSupportedOption {
+        class SupportedOption {
             +getName() string
             +isSupportedValue(mixed $value) bool
             +getSupportedValues() mixed[]
             +getJsonSchema() array< string, mixed >$
         }
-        class AiRequiredOption {
+        class RequiredOption {
```
felixarntz: Nit-pick: Reverse the order since otherwise I believe you sorted them alphabetically

docs/ARCHITECTURE.md
```diff
@@ -952,66 +950,133 @@ direction LR
...
+            +getFunctionDeclarations() FunctionDeclaration[]?
+            +getWebSearch() WebSearch?
+            +getJsonSchema() array< string, mixed >$
+        }
+        class WebSearch {
+            +getAllowedDomains() string[]
+            +getDisallowedDomains() string[]
+            +getJsonSchema() array< string, mixed >$
+        }
+    }
```
felixarntz: I'd say let's move (or at least copy) these also to the other class diagram, since they're relevant for the Implementer API too.

felixarntz (commented):
@borkweb Thank you for all the updates in accordance with #21!

Almost LGTM, just a few small notes.

docs/ARCHITECTURE.md
borkweb: It feels redundant, but I get your point and it _would_ cause less required aliasing via `use x as y;`. The change has been made!

docs/ARCHITECTURE.md
borkweb: Dang! I thought I got them all.

docs/ARCHITECTURE.md
borkweb: Finding a home for `ModalityEnum` is a little challenging. It _feels_ like `Models/Enums` is the right home, but we don't/won't have a top level `Models/` directory. So...I plopped it in a top level `Enums/` directory.

Since this is the last holdout of this body of work, if you have an idea that makes sense - definitely push a commit to this PR or merge if we're good here 🕺

JasonTheAdams: I had a thought this morning! What if we did this:
1. Make a `Prompts` namespace
2. Move the `PromptBuilder` to `Prompts`
3. Move the `MessagesBuilders` to Messages`
4. Put the common enums we want folks to be thinking of for prompts into the `Prompts` domain

That, or we move `Enums` to `Builders.Enums`. In any case, in keeping with the implementors thinking, we move the prompt builder and its commonly related parts close together.

borkweb: I like bullets 1-4 as an approach. I think it - at the very least - makes `PromptBuilder` more findable as `Builders` isn't as intuitive of a directory name. It also makes sense that the `MessageBuilder` would exist in `Messages`.

felixarntz: Hmm, I see how finding a home for `ModalityEnum` isn't as straightforward as the other two (which I see you already migrated, and LGTM 👍 ).

But I don't think we should add a `Prompts` namespace, primarily because a "prompt" is not an actual thing in this SDK. There's no `Prompt` class for example, and the only thing that uses the term is `PromptBuilder`. To me, "prompt" is the higher-level term for whatever is sent to the model altogether - but if we made that a namespace, it would mean almost every other root level namespace could fall under it, which I find confusing.

IMO, `ModalityEnum` fits best under `AiClientNamespace.Messages.Enums`. Every `Message` consists of parts of different modalities, and that applies to both prompts as well as responses/results (for both of which `Message` is used).

docs/ARCHITECTURE.md
```diff
@@ -523,79 +527,147 @@ direction LR
...
+        class FinishReasonEnum {
+            STOP
+            LENGTH
+            CONTENT_FILTER
+            TOOL_CALLS
+            ERROR
+        }
+    }
+
+    namespace AiClientNamespace.Results.DTO {
```
felixarntz: Nit-pick: The order of these two namespaces needs to be reversed due to alphanumeric order.

docs/ARCHITECTURE.md
```diff
@@ -523,79 +527,147 @@ direction LR
...
+
+    namespace AiClientNamespace.Enums {
+        class ModalityEnum {
+            TEXT
+            DOCUMENT
+            IMAGE
+            AUDIO
+            VIDEO
+        }
+    }
```
felixarntz: Per my reply in https://github.com/WordPress/php-ai-client/pull/23#discussion_r2214205222, I think this fits best in `AiClientNamespace.Messages.Enums`.

felixarntz (commented):
Except for the point on `ModalityEnum`, this LGTM.

I left some feedback on that comment thread. The more I think about it, the more intuitive I find for this to be under the `Messages` namespace.

felixarntz (approved):
Thanks @borkweb!

LGTM, this looks a lot cleaner for the structure than before. If additional ideas come up, we can further discuss or iterate in subsequent issues/PRs.

I'm going to update #20 accordingly based on this.

docs/ARCHITECTURE.md
borkweb: I think a case can be made in either direction, tbh. But ultimately, I'm not too picky and can get behind either option, so let's go with @felixarntz 🎈

PR #25, opened 2025-07-17
Adds all the enums

JasonTheAdams:
This adds all the enums within the architecture. I saw in #20 that it says to add the Implementor enums, but I wasn't sure which those were — perhaps the ones **not** in the `Providers` domain? Otherwise, they're all here!

I wasn't sure where to put the `AbstractEnum` class, so I made a `Common` domain that's strictly for things like this. Open to other ideas! Otherwise, I tried to make the `AbstractEnum` work like the PHP 8.1 enum type as closely as possible, as I felt that would make things feel more consistent and easier to upgrade in the future. Similarities:
* Use `$enum->value` to get the underlying value
* Use `$enum->name` to get the name of the enum (i.e. the constant name)
* Use `MyEnum::from()` make an enum, but get an exception if there's no match
* Use `MyEnum::tryFrom()` to make an enum, returns null if there's no match

There are other non-standard methods because you can't use the comparison operator with these, and this has instances whereas enums technically do not. But I tried to make it similar and easy to make child classes of.

## Singelton behvior?

One thing the Claude did when generating this that I thought was interesting was go the extra mile in making these truly act like 8.1 enums is by being strictly equivalent with its instances. That is, this is true:
```php
$enum1 = CMS::from('WordPress');
$enum2 = CMS::from('WordPress');

$equal = $enum1 === $enum2; // true
```

It does this by caching instances based on the value inside a class static property. It's pretty clever. Since enums are immutable, it isn't possible to change one and have it cause issues elsewhere. Curious if we like this or if it's too voodoo.


.gitignore
```diff
@@ -12,3 +12,4 @@ vendor/
 [Dd]esktop.ini
 *.DS_store
 .DS_store?
+.claude/
```
felixarntz, line 20: While I am aware you created this with the help of Claude, let's not add any Claude specific config, at least not yet. This kind of concern requires further discussion, and I acknowledge that we need to think about how to best solve it. The reason it requires discussion is that as an open source project we need to carefully consider which tooling we use, and thus indirectly encourage using.

Case in point: If we add `CLAUDE.md`, are we also going to add `GEMINI.md`? Are we also going to add Cursor rules? Are we also going to add Cline rules? Obviously adding all of these would feel excessive, so I don't think that's the solution. But we also shouldn't just add a single one without having discussed the direction. It's a tricky question: How can we set up the tooling in this project that it supports developers that use LLMs for coding assistance, but without forcing them to use one way? Or is it reasonable to go with just one setup?

Anyway, I would prefer that we don't discuss this here, but in a new issue. For now, I'd recommend you keep these Claude specific files in a local directory or separate branch, or you could even put them somewhere outside this project (e.g. a GitHub Gist) so that others could use them in case they also use Claude and want to contribute with that setup. But let's remove them from this PR so we can first collect more thoughts on how we want to approach that for this project.

CLAUDE.md
```diff
@@ -0,0 +1,65 @@
+# CLAUDE.md
```
felixarntz: See above, let's remove this.

_If_ you prefer to keep the file content itself as part of the repo, what would be reasonable _without_ having a discussion on the LLM provider specific approach would be to put this as another file in `docs`, while removing any Claude specific mentions etc from it. After all, everything that's in here can also be considered documentation.

You could even call it something like `LLM-OVERVIEW.md`, to clarify that developers could load this into their AI coding assistance tool of choice (or copy into a file like `CLAUDE.md` or `GEMINI.md` in their local checkout). I realize that that's not the ideal solution, but it's an acceptable intermediate solution until we have decided on proper AI coding assistance setup for this project.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+     *
+     * @param string|int $value The enum value
+     * @param string $name The constant name
+     * @return static
+     * @phpstan-return static
+     */
+    private static function getInstance($value, string $name): self
+    {
+        $className = static::class;
+        $key = $name;
```
felixarntz: Nit-pick: Seems unnecessary to have a `$key` variable here? Could just use `$name` throughout.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+            }
+
+            self::$cache[$className] = $enumConstants;
+        }
+
+        return self::$cache[$className];
+    }
+
+    /**
+     * Handle dynamic method calls for enum creation and checking
```
felixarntz: "enum _creation_ and checking"? As far as I can tell, this only considers checking, not creating.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    /**
+     * Magic getter to provide read-only access to properties
+     *
+     * @param string $property The property name
+     * @return mixed
+     * @throws BadMethodCallException If property doesn't exist
+     */
+    public function __get(string $property)
```
felixarntz: Since none of the methods in this class should ever be overwritten, but the class _does_ need to be abstract, I think it would help to make all public and protected methods in this class `final`, to clarify that interface and make sure developers can't mess with it in unexpected ways.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    /**
+     * Create an enum instance from a value, throws exception if invalid
+     *
+     * @param string|int $value The enum value
+     * @return static
+     * @throws InvalidArgumentException If the value is not valid
+     */
+    public static function from($value): self
```
felixarntz: Leaving a few thoughts here regarding documentation, which apply to the entire PR. Let's discuss them here for how we want to proceed.

While we follow PSR coding standards, to my knowledge they don't specify requirements for documentation. However, the WordPress Coding Standards do. I'm not saying we need to follow the WordPress Coding Standards for documentation, but I think we should discuss each of the requirements and whether we want to embrace them for this project or not. It's crucial that we define our documentation requirements for this project early on so that code added adheres to them without later requiring tedious refactors.

Here are the points to consider:
* WordPress requires a description for the `@return` tag. I think this is just as crucial and helpful as it is to have description for `@param` tags, so I think we should require that here too.
* WordPress requires a `@since` annotation for almost everything (e.g. methods, class properties, classes, interfaces, traits). I personally think this is really useful as it allows to know at a glance when a certain thing was added. So I would suggest to require that for this project too.
    * How we would handle it is a related question. What we've done in other projects (e.g. `WordPress/performance`) is to always put `@since n.e.x.t` on anything newly added, so that it can later be replaced with the concrete version as part of release preparation, as at that point the release number will be clear.
* WordPress requires multiple `@param` tags for a single method to have its individual parts (type, name, description) indented equally. I personally like that for readability, but obviously a matter of preferene. We do need to decide for yes or no though.
* WordPress requires pretty much any description to end in a period / full stop. I'm personally very used to that, so I prefer it, but again a matter of preference. Yes or no?
* WordPress requires method and function descriptions to start with a third-person verb (e.g. "Creates" instead of "Create"). Again a matter of preference. Yes or no?

Please let me know your thoughts. Whatever we decide for all of these points we should document somewhere in `docs` or `CONTRIBUTING.md`, related to the coding standards.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+     *
+     * @param string|int $value The enum value
+     * @return static
+     * @throws InvalidArgumentException If the value is not valid
+     * @deprecated Use from() method instead
+     */
+    public static function fromValue($value): self
+    {
+        return self::from($value);
+    }
```
felixarntz: We shouldn't add deprecated methods. No reason to worry here about backward compatibility, so let's get rid of this method.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Get or create a singleton instance for the given value and name
+     *
+     * @param string|int $value The enum value
+     * @param string $name The constant name
+     * @return static
+     * @phpstan-return static
```
felixarntz: Why is `@phpstan-return` needed here, given its value is the same as `@return`.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * @var array<string, array<string, string|int>> Cache for reflection data
+     */
+    private static $cache = [];
+
+    /**
+     * @var array<string, array<string, self>> Cache for enum instances
+     */
+    private static $instances = [];
```
felixarntz: Since PHP 7.4 supports typed properties, can you make use of that for these? Probably something that we should mandate/encourage throughout the project.

CLAUDE.md
```diff
@@ -0,0 +1,65 @@
...
+- `Tools`: Function calling and tool support
+- `Util`: Utility classes for common operations
+
+### Key Design Patterns
+
+- **Interface Segregation**: Separate interfaces for different model capabilities (TextGenerationModelInterface, ImageGenerationModelInterface, etc.)
+- **Composition over Inheritance**: Models compose capabilities through interfaces rather than inheritance
+- **DTO Pattern**: Data Transfer Objects for messages, results, and configurations with JSON schema support
+- **Builder Pattern**: Fluent builders for constructing prompts and messages
+
```
felixarntz: Directly related to this PR, if we want to keep this file, I wonder: Should we document how enums need to be used in this project?

felixarntz (changes requested):
@JasonTheAdams This looks great so far, thank you for putting it together!

> I saw in #20 that it says to add the Implementor enums, but I wasn't sure which those were — perhaps the ones **not** in the `Providers` domain?

Exactly, the idea is that everything related to the extender API only is in the `Providers` namespace, and everything else is considered part of the implementer API.

> I wasn't sure where to put the `AbstractEnum` class, so I made a `Common` domain that's strictly for things like this.

That sounds perfect to me 👍 

> One thing the Claude did when generating this that I thought was interesting was go the extra mile in making these truly act like 8.1 enums is by being strictly equivocal with its instances.

I love it! It does feel like a hack, and singleton has several problems generally, but this is a perfect example of where the general best practice does not apply. This is a really creative and pragmatic solution, which I don't see any problems with, since it's stateless enum classes anyway.

One higher-level question I have (which relates to the last point and overall implementation of this PR) is: Do we want to mandate use of the enum classes for implementers and extenders (they would use e.g. `ProviderTypeEnum::fromCloud()`)? Or do we want to encourage them to use the raw values and handle enum safety under the hood (they would use e.g. `ProviderTypeEnum::CLOUD`)?

.gitignore
JasonTheAdams: I align with your greater point of not including an agent-specific file to the project, but do we also want to omit that from the `.gitignore` file? Seems like it's not a big deal to have it there and helps ensure said files _don't_ get added? Otherwise, I have to keep making sure I don't add it when doing `git add`. 🤔

CLAUDE.md
JasonTheAdams: I agree with this and will remove it. As much as I like the idea of having an `LLM-OVERVIEW.md` file, the strength of the agent-specific files is that they're automatically included, whereas a file like that wouldn't be. So we can just make sure our existing CONTRIBUTING and such files remain up-to-date for the LLM to follow.

It really doesn't bother me to just have my own local CLAUDE.md file that's derived from all out other documentation. 👍

src/Common/AbstractEnum.php
JasonTheAdams: Hah! Amazing. Resolved in ebe9d2a

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in 970d794

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in 189c478

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in 71529cb

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in 71529cb

src/Common/AbstractEnum.php
JasonTheAdams: Agreed! Resolved in 71529cb

JasonTheAdams:
> Exactly, the idea is that everything related to the extender API only is in the `Providers` namespace, and everything else is considered part of the implementer API.

So should I remove the enums in `Providers` in this PR? Or let them be for now?

> I love it! It does feel like a hack, and singleton has several problems generally, but this is a perfect example of where the general best practice does not apply. This is a really creative and pragmatic solution, which I don't see any problems with, since it's stateless enum classes anyway.

Right?? I actually really like it! It's impressive how close we can actually get this to be like 8.1 enums!

> One higher-level question I have (which relates to the last point and overall implementation of this PR) is: Do we want to mandate use of the enum classes for implementers and extenders (they would use e.g. `ProviderTypeEnum::fromCloud()`)? Or do we want to encourage them to use the raw values and handle enum safety under the hood (they would use e.g. `ProviderTypeEnum::CLOUD`)?

My inclination would be that in the `Providers` namespace we require use of the enums. For the Implementor APIs we allow _either_ the enum or the underlying value. I've done that pattern in the past and like it. It allows for the more advanced devs to have clearly coupled code and simpler use cases to just go the primitives route.

src/Common/AbstractEnum.php
JasonTheAdams: When it comes to conventions such as `@since n.e.x.t`, I'm perfectly happy adding that to the CONTRIBUTING doc. I think requiring things like docs for `@return` also fit there.

For things like spacing the `@param`, this is a direct deviation from the PSR standard, which means we'd actually be fighting any tools that are trying to adhere to the standard. For these I think we'd be working against ourselves, even though I agree that the indenting looks nice.

In short, for things that PSR-12/PER don't care about (e.g. full stops at the ends of comments and `@since` conventions), I think it's worth documenting and doing. But I'd avoid contradicting the standards themselves.

CLAUDE.md
JasonTheAdams: I think it's worth documenting in CONTRIBUTING.md! I'll update it based on our other decision around how we require enums in method calls.

.gitignore
felixarntz: Agreed that the `.gitignore` point is not as important. But if we want to keep it, then maybe let's add at least a few more exclusions for other popular solutions, plus the `CLAUDE.md` file itself. Like:
* `.clinerules`
* `.cursor`
* `GEMINI.md`
* ...

Doesn't have to be all comprehensive, but at least that clarifies that for now we wouldn't want any of those files in there.

src/Common/AbstractEnum.php
felixarntz: That's fair. Let's go with what you're saying - I wasn't aware that indenting actively contradicts PSR.

In other words, we'll stick to the WordPress conventions for documentation, except where they conflict with PSR. That's probably the most intuitive way to summarize it.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,359 @@
...
+{
+    /**
+     * @var string|int The value of the enum instance
+     */
+    private $value;
+
+    /**
+     * @var string The name of the enum constant
+     */
+    private string $name;
```
felixarntz: I already was thinking about it during my initial review, but now it stands out to me even more. While I'm aware that PHP 8 enums support both integers and strings, I don't really see the need to support integers in our implementation. I'm sure that we wouldn't support integers if we weren't aware of how enums were implemented in PHP8.

* We won't use integers anyway, as they're not meaningful when seeing the raw values.
* Since we'll also have REST APIs later where data is passed around as JSON, using string values is more intuitive for that reason alone.
* Enforcing strings leads to type safe code.
* Not supporting integers doesn't lead to any conflicts. If we were to later transition to PHP8 enums, it would still be a smooth transition.

Based on this, I would suggest we consistently type the `$value` as `string` throughout this PR.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * Magic getter to provide read-only access to properties
+     *
+     * @since n.e.x.t
+     * @param string $property The property name
+     * @return mixed
+     * @throws BadMethodCallException If property doesn't exist
+     */
+    final public function __get(string $property)
```
felixarntz, line 80: See above, it would also allow to put a clear `string` return value on this, for example.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,359 @@
...
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Common;
+
+use BadMethodCallException;
+use InvalidArgumentException;
+use ReflectionClass;
+
+/**
+ * Abstract base class for enum-like behavior in PHP 7.4
```
felixarntz: Based on our conversation in https://github.com/WordPress/php-ai-client/pull/25#discussion_r2217913029, can you update the docs in this PR accordingly?

* Descriptions (of methods, parameters, return, ...) always end in a full stop.
* Method descriptions always start with a third-person verb (except special cases such as a constructor, or maybe magic methods).
* Return annotations have a description.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,359 @@
...
+        $this->name = $name;
+    }
+
+    /**
+     * Magic getter to provide read-only access to properties
+     *
+     * @since n.e.x.t
+     * @param string $property The property name
+     * @return mixed
+     * @throws BadMethodCallException If property doesn't exist
```
felixarntz: In WordPress, the `@since` annotation is always part of another "group" of annotations, I.e. separated with a blank line. Is that something that conflicts with PSR or not? If not, might I suggest we do that here too? I do personally think that makes it easier to read, and it makes sense to separate, since `@param`, `@return`, and `@throws` are arguably the most crucial things to understand about a method.
```suggestion
     * @since n.e.x.t
     *
     * @param string $property The property name
     * @return mixed
     * @throws BadMethodCallException If property doesn't exist
```

If you agree, let's do that throughout.

felixarntz (commented):
@JasonTheAdams Looks close! I think the missing pieces to update are to align with the documentation based on what we're discussing in https://github.com/WordPress/php-ai-client/pull/25#discussion_r2217913029 and add other popular AI integrations to `.gitignore` per https://github.com/WordPress/php-ai-client/pull/25#discussion_r2223061690.

> So should I remove the enums in `Providers` in this PR? Or let them be for now?

I think let's leave them in. Wasn't originally in the scope of this PR, but you already did the work, so we can be pragmatic about it - the code is very much similar to the other enums anyway :)

> > One higher-level question I have (which relates to the last point and overall implementation of this PR) is: Do we want to mandate use of the enum classes for implementers and extenders (they would use e.g. `ProviderTypeEnum::fromCloud()`)? Or do we want to encourage them to use the raw values and handle enum safety under the hood (they would use e.g. `ProviderTypeEnum::CLOUD`)?
> 
> My inclination would be that in the `Providers` namespace we require use of the enums. For the Implementor APIs we allow _either_ the enum or the underlying value. I've done that pattern in the past and like it. It allows for the more advanced devs to have clearly coupled code and simpler use cases to just go the primitives route.

That sounds great to me, let's do that.

Could you maybe also add the points about documentation requirements and enum usage (per the above) to existing docs? Or, if you prefer, feel free to open a separate PR for that.

src/Common/AbstractEnum.php
JasonTheAdams: Sure! We can always make an `AbstractIntEnum` later if we want.

Resolved in c78973e

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in c78973e

src/Common/AbstractEnum.php
JasonTheAdams: Resolved in [863a07c](https://github.com/WordPress/php-ai-client/pull/25/commits/863a07cb20486891c1884411eabbbb45d18dfb79)

src/Common/AbstractEnum.php
JasonTheAdams: Sure! Resolved in [5cc061f](https://github.com/WordPress/php-ai-client/pull/25/commits/5cc061f0b33b9a9ab8f65db753547a2823ccb43d)

JasonTheAdams:
Back to you, @felixarntz! I resolved all the feedback and made #29 to add the documentation standards to the docs. I'll add the enum stuff to the docs in another PR.

felixarntz:
@JasonTheAdams Reviewing now... I think the point in https://github.com/WordPress/php-ai-client/pull/25#discussion_r2223061690 is still missing in the code.

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,375 @@
...
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Common;
+
+use BadMethodCallException;
+use InvalidArgumentException;
+use ReflectionClass;
+
+/**
+ * Abstract base class for enum-like behavior in PHP 7.4
```
felixarntz:
```suggestion
 * Abstract base class for enum-like behavior in PHP 7.4.
```

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,375 @@
...
+ * $enum->equals('first'); // Returns true
+ * $enum->is(PersonEnum::firstName()); // Returns true
+ * PersonEnum::cases(); // Returns array of all enum instances
+ *
+ * @since n.e.x.t
+ */
+abstract class AbstractEnum
+{
+    /**
+     * @var string The value of the enum instance
```
felixarntz:
```suggestion
     * @var string The value of the enum instance.
```

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,375 @@
...
+     */
+    private string $value;
+
+    /**
+     * @var string The name of the enum constant
```
felixarntz:
```suggestion
     * @var string The name of the enum constant.
```

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,375 @@
...
+     */
+    private string $name;
+
+    /**
+     * @var array<string, array<string, string>> Cache for reflection data
```
felixarntz:
```suggestion
     * @var array<string, array<string, string>> Cache for reflection data.
```

src/Common/AbstractEnum.php
```diff
@@ -0,0 +1,375 @@
...
+     */
+    private static array $cache = [];
+
+    /**
+     * @var array<string, array<string, self>> Cache for enum instances
```
felixarntz:
```suggestion
     * @var array<string, array<string, self>> Cache for enum instances.
```

src/Providers/Enums/ProviderTypeEnum.php
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Server-side AI provider (e.g., self-hosted models).
+     */
+    public const SERVER = 'server';
+
+    /**
+     * Client-side AI provider (e.g., browser-based models).
+     */
+    public const CLIENT = 'client';
```
felixarntz: I think it's better here to not put concrete provider names for the cloud based providers - partially for neutrality, partially because we don't do that for the other two types either.

Nit-pick, I don't think there needs to be a comma after `e.g.`?

```suggestion
    /**
     * Cloud-based AI provider (e.g. models available via external REST APIs).
     */
    public const CLOUD = 'cloud';

    /**
     * Server-side AI provider (e.g. self-hosted models).
     */
    public const SERVER = 'server';

    /**
     * Client-side AI provider (e.g. browser-based models).
     */
    public const CLIENT = 'client';
```

felixarntz (approved):
Almost good to go, just a few last nit-picks. Can you please address these, and more importantly https://github.com/WordPress/php-ai-client/pull/25#discussion_r2223061690?

After that, consider it approved.

.gitignore
JasonTheAdams: Resolved in 2cf2ea9

PR #28, opened 2025-07-22
Adds Implementor Data Transfer Objects

JasonTheAdams:
Per the second task of #20, this adds the DTOs for the Implementor portion of the codebase based on the defined architecture.

Note, I purposely have not added the unit tests. I figure we'll want to look this over thoughtfully first to make sure it architecturally aligns with our thinking. Once we're happy with what we're seeing then I'll add all the unit tests.


src/Embeddings/DTO/Embedding.php
JasonTheAdams, line 1: @felixarntz Am I recalling correctly that we're going to remove all embedding things for now?

justlevine: related: https://wordpress.slack.com/archives/C08TJ8BPULS/p1752772302665399

JasonTheAdams: To be clear, @justlevine, this isn't meant to imply removing them indefinitely, just from the MVP. But I could be remembering wrong, too! 😄

felixarntz: +1, I think it makes sense to leave out all the embeddings code for now, as that will only be a priority after the MVP. And since we're not entirely sure about their implementation, I think it makes sense to leave them for when we get there.

justlevine: @JasonTheAdams yes aware and agreed, just keeping all the context in one place

JasonTheAdams: Resolved in 5414454

src/Files/DTO/InlineFile.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $mimeType The MIME type of the file.
+     * @param string $base64Data The base64-encoded file data.
+     */
+    public function __construct(string $mimeType, string $base64Data)
```
felixarntz: Can we reverse the order of these? The data is more substantial than the MIME type, plus this is in line with how methods like `PromptBuilder::withInlineImage` are already outlined.

src/Files/DTO/InlineFile.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->mimeType = $mimeType;
+        $this->base64Data = $base64Data;
```
felixarntz: We should sanitize the `$base64data` here, to make sure at the root that it's always _only_ the data. Some providers may return base64 data _URLs_, and some developers may pass those around, hence we need to sanitize here to strip the respective prefix if it is one.

For example:
```suggestion
        $this->mimeType = $mimeType;
        $this->base64Data = preg_replace( '/^data:([a-z0-9-]+\/[a-z0-9-]+);base64,/', '', $base64Data );
```

Potentially taking this further, @borkweb and I had discussed in an earlier architecture PR to make the `$mimeType` parameter optional. Ideally the developer would still pass it just to be safe, but if they passed a full data _URL_, that alone would be sufficient, as we would be able to parse out the MIME type. WDYT?

src/Files/DTO/InlineFile.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return $this->base64Data;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function getJsonSchema(): array
```
felixarntz: Something that we didn't chat about yet, but I think we should be doing for all DTOs to support these `getJsonSchema()` methods is to provide easy ways to transform a DTO to and from the raw data conforming to the respective JSON schema.

I'm thinking we should define another interface in `Common/Contracts` like `WithRawDataRepresentation` or something like that. It should have methods like:
* `public function toRawData(): array`
* `public static function fromRawData(array $data): static`

We could also combine it somehow with PHP's native `JsonSerializable` interface, basically the `jsonSerialize()` method would be the same as `toRawData()`. That said, I kind of like having both since you may want to get the raw data representation in some cases without necessarily wanting to immediately JSON encode it - i.e. it's not _only_ relevant in context of JSON data. Alternatively, if y'all feel strongly about not having two methods that do the same thing, we could consider calling `fromRawData($data)` something like `jsonUnserialize($data)` - the main thing I'd want to make sure is that it would be clear that method would be the opposite of that other method. If we do the latter, maybe it could be called `WithJsonRepresentation`? And then that would automatically extend `JsonSerializable`?

Let's discuss in this thread how to implement this. Once we've agreed on the approach, feel free to add it as part of this PR, or as a separate follow up.

src/Files/DTO/LocalFile.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $mimeType The MIME type of the file.
+     * @param string $path The local filesystem path to the file.
+     */
+    public function __construct(string $mimeType, string $path)
```
felixarntz: See above, let's put `$path` first.

And similar to `InlineFile` (though here even more), we should probably make `$mimeType` optional. Here it's more or less guaranteed that we can infer it, from the file extension of `$path` (unless it has none, which is extremely unlikely).

Since we'd need the same for `RemoteFile`, maybe you could implement this in some other util class just for that, like `MimeTypeUtil::getMimeTypeForExtension(string $ext): string`. Thoughts?

src/Files/DTO/RemoteFile.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $mimeType The MIME type of the file.
+     * @param string $url The URL to the remote file.
+     */
+    public function __construct(string $mimeType, string $url)
```
felixarntz: See above, regarding putting `$url` first and making `$mimeType` optional.

src/Messages/DTO/Message.php
```diff
@@ -0,0 +1,107 @@
...
+     * @since n.e.x.t
+     *
+     * @param MessageRoleEnum $role The role of the message sender.
+     * @param string $text The text content.
+     * @return self
+     */
+    public static function fromText(MessageRoleEnum $role, string $text): self
+    {
+        return new self($role, [MessagePart::text($text)]);
+    }
```
felixarntz: I don't think this should live here. We'll already implement a `MessageBuilder` and `PromptBuilder` to make this simple, no need for a random static method on this class. It kind of opens a can of worms, like should we also add `fromImageFile` etc.

I'd say let's focus on the foundation for now. We can always add simplifications later as we go, but better to do that when we have a holistic overview and initial code examples to assess where things are more complicated than they need to be.

src/Messages/DTO/Message.php
```diff
@@ -0,0 +1,107 @@
...
+     * @since n.e.x.t
+     */
+    public static function getJsonSchema(): array
+    {
+        return [
+            'type' => 'object',
+            'properties' => [
+                'role' => [
+                    'type' => 'string',
+                    'enum' => ['user', 'model', 'system'],
```
felixarntz: Can you refer to the values from `MessageRoleEnum` here instead?
```suggestion
                    'enum' => array_values( MessageRoleEnum::getValues() ),
```

Writing out this suggestion, I wonder whether `AbstractEnum::getValues()` would be better to include the `array_values()` itself. Not sure what would be the use-case to also need the constant names in a map.

src/Messages/DTO/MessagePart.php
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * @param FunctionResponse $functionResponse The function response.
+     * @return self
+     */
+    public static function functionResponse(FunctionResponse $functionResponse): self
+    {
+        $part = new self(MessagePartTypeEnum::functionResponse());
+        $part->functionResponse = $functionResponse;
+        return $part;
+    }
```
felixarntz: See above, let's not add shortcuts like these just yet.

What I think this class would benefit from instead would be a more intuitive constructor, where you pass in any of the supported data structures, and the class will figure it out.

src/Messages/DTO/MessagePart.php
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * Private constructor to enforce factory method usage.
+     *
+     * @since n.e.x.t
+     *
+     * @param MessagePartTypeEnum $type The type of this message part.
+     */
+    private function __construct(MessagePartTypeEnum $type)
+    {
+        $this->type = $type;
```
felixarntz: See my other comment, instead of passing in the type here, we should allow passing in the actual data structure, and then the class can automatically determine the type value based on that. That way the class works equivalent to the other DTOs, allowing to pass in all necessary data in the constructor.

So basically replace the parameter with something like `$data`, which can be either a `string`, `InlineFile`, `RemoteFile`, etc. (validation would need to happen inside the constructor, since we can't use union types). We can throw an `InvalidArgumentException` if it's neither of them.

Then, based on what `$data` is, set it in the respective class property, and set the `$type` property accordingly.

src/Messages/DTO/MessagePart.php
```diff
@@ -0,0 +1,264 @@
...
+                    'description' => 'Function call request (when type is function_call).',
+                ],
+                'functionResponse' => [
+                    'oneOf' => [
+                        ['type' => 'null'],
+                        FunctionResponse::getJsonSchema(),
+                    ],
+                    'description' => 'Function response (when type is function_response).',
+                ],
+            ],
```
felixarntz: I think the more intuitive schema to put the `oneOf` on the outside. In other words, one of:
* `'type' => 'text', 'text' => ...`
* `'type' => 'inlineFile', 'inlineFile' => ...`
* `'type' => 'remoteFile', 'remoteFile' => ...`

The `jsonSerialize()` function (see other comment) can handle the correct formatting.

src/Messages/DTO/ModelMessage.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param string $text The text content.
+     * @return self
+     */
+    public static function text(string $text): self
+    {
+        return new self([MessagePart::text($text)]);
+    }
```
felixarntz: See above, let's remove this.

src/Messages/DTO/SystemMessage.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param string $text The text content.
+     * @return self
+     */
+    public static function text(string $text): self
+    {
+        return new self([MessagePart::text($text)]);
+    }
```
felixarntz: See above.

src/Messages/DTO/UserMessage.php
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param string $text The text content.
+     * @return self
+     */
+    public static function text(string $text): self
+    {
+        return new self([MessagePart::text($text)]);
+    }
```
felixarntz: See above.

src/Operations/DTO/GenerativeAiOperation.php
```diff
@@ -0,0 +1,114 @@
...
+                    'enum' => ['starting', 'processing', 'succeeded', 'failed', 'canceled'],
+                    'description' => 'The current state of the operation.',
+                ],
+                'result' => [
+                    'oneOf' => [
+                        ['type' => 'null'],
+                        GenerativeAiResult::getJsonSchema(),
+                    ],
+                    'description' => 'The result once the operation completes.',
+                ],
```
felixarntz: Somewhat related to my other `oneOf` feedback, similarly applies here: Better to put the `oneOf` on the outside, to clarify:
* Either this is an object with `state` always set to `succeeded`, and `result` containing a `GenerativeAiResult`.
* Or this is an object with `state` to one of the other values, and `result` being `null`. Or you could even omit `result` in that case.

src/Results/DTO/Candidate.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param Message $message The generated message.
+     * @param FinishReasonEnum $finishReason The reason generation stopped.
+     * @param int $tokenCount The number of tokens in this candidate.
+     */
+    public function __construct(Message $message, FinishReasonEnum $finishReason, int $tokenCount)
+    {
+        $this->message = $message;
```
felixarntz, line 53: Should we include a validation check to make sure that it uses the `model` role? For a `Candidate`, that's really the only kind of message that should ever be in it.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @param string $id Unique identifier for this result.
+     * @param Candidate[] $candidates The generated candidates.
+     * @param TokenUsage $tokenUsage Token usage statistics.
+     * @param array<string, mixed> $providerMetadata Provider-specific metadata.
+     */
+    public function __construct(string $id, array $candidates, TokenUsage $tokenUsage, array $providerMetadata = [])
+    {
+        $this->id = $id;
+        $this->candidates = $candidates;
```
felixarntz, line 60: Instead of validating in all the individual output methods, shouldn't we validate in here? `$candidates` must always contain at least 1 entry.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+     */
+    public function toText(): string
+    {
+        if (empty($this->candidates)) {
+            throw new \RuntimeException('No candidates available');
+        }
+
+        $message = $this->candidates[0]->getMessage();
+        foreach ($message->getParts() as $part) {
+            if ($part->getType()->equals(MessagePartTypeEnum::text()) && $part->getText() !== null) {
```
felixarntz: Technically, the first check here is unnecessary. The type is always inferred from what kind of data is set in the part, so no need to check for that too.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        $message = $this->candidates[0]->getMessage();
+        foreach ($message->getParts() as $part) {
+            if ($part->getType()->equals(MessagePartTypeEnum::inlineFile()) && $part->getInlineFile() !== null) {
+                return $part->getInlineFile();
+            }
+            if ($part->getType()->equals(MessagePartTypeEnum::remoteFile()) && $part->getRemoteFile() !== null) {
+                return $part->getRemoteFile();
+            }
```
felixarntz: These should also check for the MIME type of the respective file to ensure it's an image?

I mentioned in another comment adding a `MimeTypeUtil` class, I think that class could also be used here, e.g. with a method `isImage(FileInterface $file): bool`.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @return FileInterface The audio file.
+     * @throws \RuntimeException If no candidates or no audio content.
+     */
+    public function toAudioFile(): FileInterface
+    {
+        // Similar implementation to toImageFile, but checking for audio MIME types
+        return $this->toImageFile(); // Simplified for now
```
felixarntz: See above, ideally this verifies that it's an audio file.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @return FileInterface The video file.
+     * @throws \RuntimeException If no candidates or no video content.
+     */
+    public function toVideoFile(): FileInterface
+    {
+        // Similar implementation to toImageFile, but checking for video MIME types
+        return $this->toImageFile(); // Simplified for now
```
felixarntz: See above.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @return string[] Array of text content.
+     */
+    public function toTexts(): array
+    {
+        $texts = [];
+        foreach ($this->candidates as $candidate) {
+            $message = $candidate->getMessage();
+            foreach ($message->getParts() as $part) {
+                if ($part->getType()->equals(MessagePartTypeEnum::text()) && $part->getText() !== null) {
```
felixarntz: Similar feedback as above applies here and in the subsequent methods for multiple candidates - I'm not going to repeat it below :)

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -0,0 +1,112 @@
...
+                'name' => [
+                    'type' => 'string',
+                    'description' => 'The name of the function.',
+                ],
+                'description' => [
+                    'type' => 'string',
+                    'description' => 'A description of what the function does.',
+                ],
+                'parameters' => [
+                    'type' => 'object',
```
felixarntz: Elsewhere you have this as `mixed`, but here it's an object. We'll probably need to align it here and allow any type (although realistically in 95% of scenarios it's probably going to be an object).

src/Tools/DTO/FunctionResponse.php
```diff
@@ -0,0 +1,111 @@
...
+                    'type' => 'string',
+                    'description' => 'The ID of the function call this is responding to.',
+                ],
+                'name' => [
+                    'type' => 'string',
+                    'description' => 'The name of the function that was called.',
+                ],
+                'response' => [
+                    'description' => 'The response data from the function.',
+                ],
```
felixarntz: Is it allowed in JSON schema to omit `type`? Otherwise, you may need to specify here that any type (object, array, or scalar values) is allowed here.

src/Tools/DTO/Tool.php
```diff
@@ -0,0 +1,144 @@
...
+     * Private constructor to enforce factory method usage.
+     *
+     * @since n.e.x.t
+     *
+     * @param ToolTypeEnum $type The type of tool.
+     */
+    private function __construct(ToolTypeEnum $type)
+    {
+        $this->type = $type;
+    }
```
felixarntz: See my feedback on `MessagePart`, it applies here similarly: We should allow passing in any of the supported data structures and infer the type from that.

src/Tools/DTO/Tool.php
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @param WebSearch $webSearch The web search configuration.
+     * @return self
+     */
+    public static function webSearch(WebSearch $webSearch): self
+    {
+        $tool = new self(ToolTypeEnum::webSearch());
+        $tool->webSearch = $webSearch;
+        return $tool;
+    }
```
felixarntz: Similar to my feedback on `MessagePart`, let's remove these in favor of consistent constructor parameter support.

src/Tools/DTO/Tool.php
```diff
@@ -0,0 +1,144 @@
...
+                    'description' => 'Function declarations (when type is function_declarations).',
+                ],
+                'webSearch' => [
+                    'oneOf' => [
+                        ['type' => 'null'],
+                        WebSearch::getJsonSchema(),
+                    ],
+                    'description' => 'Web search configuration (when type is web_search).',
+                ],
+            ],
```
felixarntz: See my feedback on `MessagePart`, which applies here too.

src/Tools/DTO/Tool.php
```diff
@@ -0,0 +1,144 @@
...
+            'type' => 'object',
+            'properties' => [
+                'type' => [
+                    'type' => 'string',
+                    'enum' => ['function_declarations', 'web_search'],
+                    'description' => 'The type of tool.',
+                ],
+                'functionDeclarations' => [
+                    'type' => ['array', 'null'],
+                    'items' => FunctionDeclaration::getJsonSchema(),
```
felixarntz: Is this valid? Combing `type=null` with `items` seems like a problem to me. Don't we need to use a `oneOf` instead?

felixarntz (changes requested):
@JasonTheAdams Lots of good stuff in here! Overall this is looking great, though I also have a lot of feedback - simply because this is a lot of code and introducing many paradigms that we need to iron out.

There are probably a few things we'll need to discuss more, so let's just do that within the respective feedback threads below.

Let's take the time to get this right - once we've done that, I think we're set up for the future (e.g. the subsequent PR) to get things done fast.

src/Files/DTO/InlineFile.php
JasonTheAdams: Resolved in cb20ce1

src/Files/DTO/InlineFile.php
JasonTheAdams: Good thoughts! I went about this slightly differently. The most notable difference, and I'm happy to discuss this, is that I'm a bigger fan of validation for this sort of thing than sanitization. Since this is programmatic, an invalid base64 value should not be based; if it is, we want to catch where it's happening rather than push it through. I believe sanitization is based used for raw user input.

From there, the mime type is determined in one of three ways:
1. Uses the `$mimeType` if explicitly set
2. Parses the mime type from the base64 data if present
3. Defaults to `text/plain` per [RFC 2397](https://www.rfc-editor.org/rfc/rfc2397.html) if absent

Tentatively resolved in ac3c9e7

src/Files/DTO/LocalFile.php
JasonTheAdams: Good idea! Resolved in ec93ff4

src/Files/DTO/RemoteFile.php
JasonTheAdams: Resolved in cb20ce1 and f9d19af

src/Messages/DTO/Message.php
JasonTheAdams: Resolved in ab0cc93

src/Messages/DTO/UserMessage.php
JasonTheAdams: Resolved in 6f51c80

src/Messages/DTO/SystemMessage.php
JasonTheAdams: Resolved in 6f51c80

src/Messages/DTO/ModelMessage.php
JasonTheAdams: Resolved in 6f51c80

src/Messages/DTO/MessagePart.php
JasonTheAdams: Resolved in eb386cf

src/Messages/DTO/MessagePart.php
JasonTheAdams: Resolved in eb386cf

src/Messages/DTO/MessagePart.php
JasonTheAdams: Resolved in 109c17f

src/Messages/DTO/Message.php
JasonTheAdams: Resolved in ab0cc93

src/Results/DTO/Candidate.php
JasonTheAdams: Resolved in c213c49

src/Operations/DTO/GenerativeAiOperation.php
JasonTheAdams: Does this fit what you're thinking? 4e04fb5

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in d51a35f

src/Tools/DTO/Tool.php
TimothyBJacobs: This is allowed. JSON Schema keywords that don't apply to the actual type get ignored.

https://json-schema.org/draft/2020-12/json-schema-core#name-assertions-and-instance-pri

src/Tools/DTO/FunctionResponse.php
TimothyBJacobs: The spec allows it, but we don't in Core. Whether that should matter here, I don't know.

src/Files/DTO/InlineFile.php
felixarntz: I see what you're saying, but in practice the way you implemented it this now leads to the `$base64Data` property to not only contain the base64 data, but also that data URL prefix - in other words, it contains a data URL, not the actual data, which I think is confusing.

I think we should allow both here and sanitize, because there is no need to prohibit either of the two ways. Thinking about it further, the only actual problem is when someone passes _only_ the base64 data while omitting the MIME type, which would then be optional here too.

So how about this?
* If `$base64Data` is provided without data URL prefix _and_ there is no `$mimeType` specified, we have to fail with an exception.
* Otherwise, we can either take the specified `$mimeType` or infer it from the prefixed `$base64Data`.
* In any case, the `$base64Data` should then be sanitized to be purely that data.

We could add another "extra" method to the class like `getDataUrl()` which returns the data URL, i.e. the combination of MIME type and base64 data.

WDYT?

src/Operations/DTO/GenerativeAiOperation.php
felixarntz: Looks great!

src/Results/DTO/Candidate.php
```diff
@@ -0,0 +1,118 @@
...
+     */
+    public static function getJsonSchema(): array
+    {
+        return [
+            'type' => 'object',
+            'properties' => [
+                'message' => Message::getJsonSchema(),
+                'finishReason' => [
+                    'type' => 'string',
+                    'enum' => ['stop', 'length', 'content_filter', 'tool_calls', 'error'],
```
felixarntz:
```suggestion
                    'enum' => FinishReasonEnum::getValues(),
```

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    /**
+     * Gets the total number of candidates.
+     *
+     * @since n.e.x.t
+     *
+     * @return int The total number of candidates.
+     */
+    public function getTotalCandidates(): int
```
felixarntz: Maybe call this `getCandidateCount()`? This would match the `CANDIDATE_COUNT` model option, which controls how many candidates to generate.

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,405 @@
...
+     * Checks if the result has multiple candidates.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if there are multiple candidates, false otherwise.
+     */
+    public function hasMultipleCandidates(): bool
+    {
+        return count($this->candidates) > 1;
+    }
```
felixarntz: Neat, I like having this method in here for convenience 👍

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * Converts all candidates to image files.
+     *
+     * @since n.e.x.t
+     *
+     * @return FileInterface[] Array of image files.
+     */
+    public function toImageFiles(): array
```
felixarntz: You added a `toFile()` above, shouldn't we also have a `toFiles()` then for parity? Arguably you could then also use that in the other methods like `toImageFiles()` and then only do the MIME type filtering in there.

src/Tools/DTO/FunctionResponse.php
felixarntz: Better to include it then. Could use a `type` with an array containing all reasonably possible types (`object`, `array`, scalar values).

felixarntz (commented):
A few additional minor points of feedback, but this is shaping nicely.

The main thing to iron out is how to deal with MIME types (which we're discussing on Slack too), plus address some outstanding feedback regarding the tool DTOs from my previous review.

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in c5b632e

src/Results/DTO/GenerativeAiResult.php
JasonTheAdams: Resolved in 0540ea8

src/Tools/DTO/Tool.php
JasonTheAdams: Resolved in a80601e

src/Tools/DTO/Tool.php
JasonTheAdams: Resolved in a80601e

src/Tools/DTO/Tool.php
JasonTheAdams: Resolved in a80601e

src/Tools/DTO/FunctionResponse.php
JasonTheAdams: Good to know regarding WP. Resolved in b5bb8ee

src/Tools/DTO/FunctionDeclaration.php
JasonTheAdams: Resolved in 849558c

src/Results/DTO/Candidate.php
JasonTheAdams: Resolved in a1867dc

src/Files/DTO/InlineFile.php
JasonTheAdams: @felixarntz Ah, I see what you're saying. Somehow I got it in my head that there would always be the "data:" suffix. I can fix that.

Question, if they _do_ provide the `data:` suffix, using the RFC 2397 spec, should we fall back on `text/plain` as the mime type if it is omitted from the string — per the specification's instruction to do so? Or throw an exception if it's omitted and `$mimeType` is null?

felixarntz: Sorry, do you mean if they _don't_ provide a data URL? In that case they must provide `$mimeType`, otherwise we don't know, so will need to throw an exception. It's probably not a reasonable fallback to just assume `text/plain`.

JasonTheAdams: Tentatively resolved in af53eaf

src/Files/ValueObjects/MimeType.php
```diff
@@ -0,0 +1,259 @@
...
+
+        return in_array($this->value, $documentTypes, true);
+    }
+
+    /**
+     * Checks if this MIME type equals another.
+     *
+     * @since n.e.x.t
+     *
+     * @param mixed $other The other MIME type to compare.
```
JasonTheAdams: As a note, this used to be `@param self|string`, but PHPStan was thrown a false positive on this as it viewed the exception as unreachable. I documented this in a bug report to PHPStan: https://github.com/phpstan/phpstan/issues/13311

src/Files/DTO/InlineFile.php
JasonTheAdams: @felixarntz This is what I mean:
```mermaid
flowchart TD
      Start([Input: base64Data, mimeType])

      Start --> IsDataUri{Is data URI format?}

      IsDataUri -->|Yes| ExtractFromUri[Extract base64 and<br/>MIME from URI]
      ExtractFromUri --> HasProvidedMime1{Was MIME type<br/>provided?}

      IsDataUri -->|No| IsPlainBase64{Is plain base64?}

      IsPlainBase64 -->|No| InvalidFormat[Error: Invalid format]

      IsPlainBase64 -->|Yes| HasProvidedMime2{Was MIME type<br/>provided?}
      HasProvidedMime2 -->|No| MimeRequired[Error: MIME type required<br/>for plain base64]
      HasProvidedMime2 -->|Yes| UseProvided2[Use provided MIME type]

      HasProvidedMime1 -->|Yes| UseProvided1[Use provided MIME type]
      HasProvidedMime1 -->|No| HasExtracted{Does URI contain<br/>MIME type?}

      HasExtracted -->|Yes| UseExtracted[Use extracted MIME type]
      HasExtracted -->|No| UseDefault[Use text/plain]

      style Start fill:#90EE90
      style InvalidFormat fill:#FFB6C1
      style MimeRequired fill:#FFB6C1
```

The focus here being that a Data URI defaults to `plain/text` per the spec. So, technically, a plain base64 value **needs** a specified MIME type, whereas a Data URI does not. But that's per the spec. If we want a Data URI to require an explicit MIME type that's fine — that's the question.

JasonTheAdams: From call with @felixarntz:

We are going to have the Data URI require explicit MIME type declaration, either through the parameter or inside of the Data URI itself.

src/Files/ValueObjects/MimeType.php
justlevine: @JasonTheAdams I'm pretty sure this is expected behavior for our current neonfile.
The behavior you're describing is `treatPhpDocTypesAsCertain: false`

src/Files/DTO/InlineFile.php
JasonTheAdams: This is a great idea that we will add in a subsequent PR. 👍

src/Files/ValueObjects/MimeType.php
JasonTheAdams: Ah, that's true. I hadn't thought of that. This is the first and only time we've run into this so far, I wondering if it's worth setting that to false. Would it open up leniency and potentially lose type strength elsewhere?

JasonTheAdams: Yeah, the PHPStan folks are already acknowledging this as a bug in my Issue, so I'm going to leave it as is.

JasonTheAdams:
Alright, @felixarntz! We've done **a lot** of great stuff in this PR! Back to you for review. If it looks good then I'll add some unit tests to this PR and get it merged! 🙌

src/Files/DTO/File.php
```diff
@@ -0,0 +1,376 @@
...
+     * @throws \RuntimeException If the file is not remote.
+     */
+    public function getUrl(): string
+    {
+        if (!$this->fileType->isRemote()) {
+            throw new \RuntimeException('Cannot get URL for non-remote file.');
+        }
+
+        return $this->data;
+    }
```
felixarntz: I think we should handle these methods equivalently to how we do it in `MessagePart` and `Tool`: return the value, or `null` if it's not set (in accordance with the file type).

In other words, simply return the property value as is. Based on the internal logic of the class it'll be guaranteed that exactly one of the properties is set.

src/Files/DTO/File.php
```diff
@@ -0,0 +1,376 @@
...
+
+    /**
+     * @var FileTypeEnum The type of file storage.
+     */
+    private FileTypeEnum $fileType;
+
+    /**
+     * @var string The file data (URL for remote, base64 for inline).
+     */
+    private string $data;
```
felixarntz: See my other comment, I'd suggest to break this into two class properties, one for each `FileTypeEnum` value. This is in line with how `MessagePart` and `Tool` work.

So one property `?string $url` and another property `?string $base64Data`.

src/Files/DTO/File.php
```diff
@@ -0,0 +1,376 @@
...
+     * @throws \RuntimeException If the file is not inline.
+     */
+    public function getDataUrl(): string
+    {
+        if (!$this->fileType->isInline()) {
+            throw new \RuntimeException('Cannot get data URL for non-inline file.');
+        }
+
+        return sprintf('data:%s;base64,%s', $this->getMimeType(), $this->data);
+    }
```
felixarntz: Related to my other comment, here I would as well suggest to return `?string`, instead of throwing an exception. If `$base64Data` is set, transform it to a data URL, otherwise return `null`.

src/Files/DTO/File.php
```diff
@@ -0,0 +1,376 @@
...
+                            'type' => 'string',
+                            'description' => 'The MIME type of the file.',
+                            'pattern' => '^[a-zA-Z0-9][a-zA-Z0-9!#$&\\-\\^_+.]*\\/[a-zA-Z0-9]'
+                                . '[a-zA-Z0-9!#$&\\-\\^_+.]*$',
+                        ],
+                        'base64Data' => [
+                            'type' => 'string',
+                            'description' => 'The base64-encoded file data.',
+                        ],
+                    ],
```
felixarntz: Maybe we should also include the `fileType` as a `const` for both of these? Also similar to `MessagePart` and `Tool`.

src/Files/ValueObjects/MimeType.php
```diff
@@ -0,0 +1,259 @@
...
+            'application/pdf',
+            'application/msword',
+            'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+            'application/vnd.ms-excel',
+            'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+            'application/vnd.ms-powerpoint',
+            'application/vnd.openxmlformats-officedocument.presentationml.presentation',
+            'application/vnd.oasis.opendocument.text',
+            'application/vnd.oasis.opendocument.spreadsheet',
+        ];
```
felixarntz: Maybe store this in a private static class property too?

src/Files/ValueObjects/MimeType.php
felixarntz: I'd personally be in favor of setting `treatPhpDocTypesAsCertain: false`. The reality is that PHP doc types _aren't_ certain since they're just that, docs. Only actual PHP type hints and/or code validation can guarantee correctness.

For this reason, I've been setting `treatPhpDocTypesAsCertain: false` in most of my projects, and I'd be in favor here too.

src/Tools/DTO/FunctionCall.php
```diff
@@ -0,0 +1,113 @@
...
+                    'type' => 'string',
+                    'description' => 'The name of the function to call.',
+                ],
+                'args' => [
+                    'type' => 'object',
+                    'description' => 'The arguments to pass to the function.',
+                    'additionalProperties' => true,
+                ],
+            ],
+            'required' => ['id', 'name', 'args'],
```
felixarntz: This one is a bit tricky. Only some providers use function call IDs, others purely use names.

We have two options:
* We also require all, and for providers that under the hood don't use IDs, we just use the name for the ID as well.
* We go with a flexible `oneOf` schema of these three choices:
    * Either require `id` and `args`.
    * Or require `name` and `args`.
    * Or require all three properties.

Also worth thinking about whether `args` should be required at all. Not every function has parameters, and _having_ to provide an empty `args` object seems unnecessary in that case?

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -0,0 +1,112 @@
...
+                'description' => [
+                    'type' => 'string',
+                    'description' => 'A description of what the function does.',
+                ],
+                'parameters' => [
+                    'type' => ['string', 'number', 'boolean', 'object', 'array', 'null'],
+                    'description' => 'The JSON schema for the function parameters.',
+                ],
+            ],
+            'required' => ['name', 'description', 'parameters'],
```
felixarntz: Related to my other comment, should `parameters` be required? Not every function has parameters, so this would make sense for me to be optional.

src/Tools/DTO/FunctionResponse.php
```diff
@@ -0,0 +1,112 @@
...
+                'name' => [
+                    'type' => 'string',
+                    'description' => 'The name of the function that was called.',
+                ],
+                'response' => [
+                    'type' => ['string', 'number', 'boolean', 'object', 'array', 'null'],
+                    'description' => 'The response data from the function.',
+                ],
+            ],
+            'required' => ['id', 'name', 'response'],
```
felixarntz, line 109: See my `FunctionCall` comment regarding `id` and `name` nuances, the same consideration applies here too.

felixarntz (commented):
@JasonTheAdams Looks great, thank you for all the iterations!

A few last follow up comments, but nothing major. Definitely in a good place to add unit tests!

src/Files/DTO/File.php
JasonTheAdams: Good catch! Resolved in e93e2ee

src/Files/DTO/File.php
JasonTheAdams: Resolved in e93e2ee

src/Files/DTO/File.php
JasonTheAdams: Resolved in b7b7b63

src/Files/DTO/File.php
JasonTheAdams: Resolved in e7a58d0

src/Files/ValueObjects/MimeType.php
justlevine: Here it's not even about the nature of PHPDocs vs union type signatures IMO. The existence of a `throws` here means we're intentionally coding for that eventuality, and as such we want PHPStan to know so it's _other_ checks are more accurate.

(For example, if we _didnt_ handle an ambiguous type before returning on a strict signature, PHPStan wouldn't warn us unless `treatPhpDocTypesAsCertain: false` )

src/Files/ValueObjects/MimeType.php
JasonTheAdams: Resolved in f87fcde

src/Files/ValueObjects/MimeType.php
JasonTheAdams: You both convinced me. Hahah! Resolved in e4d5c3f

src/Tools/DTO/FunctionCall.php
JasonTheAdams: Yeah, I see no reason to require everything from a validation perspective, given our conversation earlier. It seems like it's up to the dev to know what the model and function needs. So it seems like we should require _either_ and id or name be present, and that's it?

src/Tools/DTO/FunctionDeclaration.php
JasonTheAdams: I don't think so. I'm guessing I should also make it optional in the constructor?

src/Tools/DTO/FunctionCall.php
felixarntz: Yes, _either_ has to be present. We should however also _permit_ for both to be present. But at least one of them MUST be.

src/Tools/DTO/FunctionDeclaration.php
felixarntz: Yeah, that would make sense to me.

src/Files/DTO/File.php
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Gets the URL for remote files.
+     *
+     * @since n.e.x.t
+     *
+     * @return string|null The URL, or null if not a remote file.
+     */
+    public function getUrl(): ?string
+    {
+        return $this->url;
```
felixarntz, line 170: The classic question: Should these be named `url` or `uri`?

src/Files/Enums/FileTypeEnum.php
```diff
@@ -0,0 +1,34 @@
...
+     * @var string
+     */
+    public const INLINE = 'inline';
+
+    /**
+     * Remote file referenced by URL.
+     *
+     * @var string
+     */
+    public const REMOTE = 'remote';
```
felixarntz, line 33: Do we want to rename these to `base64Data` and `url` (or `uri`, if we change that)? To me those names would be a lot more obvious in what they mean rather than `inline` and `remote`.

It would also match the type values with the corresponding other field in the JSON data (which is either `base64Data` or `url`), nicely aligned with how that is in `MessagePart` and `Tool`.

Last but not least, `inline` could be a bit too vague for a name, as it doesn't mandate for it to be a base64 encoded file, but that's what it is here.

These ideas are definitely not blockers for this PR. We can also think about this later - as long as there's no release we can rename and break things as much as we want :)

JasonTheAdams: I kind of like the names not being quite so in the weeds of what the data literally is. Having it one level higher feels closer to the intent.

What if `inline` was instead `encoded`, `embedded`, or something like that?

That said, I really don't have a strong opinion about this. I think it's close enough where it's hard to know what's "right" here. I suspect more people would mean more opinions on this.

src/Tools/DTO/FunctionDeclaration.php
JasonTheAdams: Resolved in 62b8776

src/Tools/DTO/FunctionCall.php
JasonTheAdams: Resolved in 62b8776

src/Files/DTO/File.php
JasonTheAdams: Hahah! Good question! _Does_ this work with URIs that are not URLs?

felixarntz: Not sure - happy to leave this as is.

src/Files/Enums/FileTypeEnum.php
felixarntz: To me, it's simply that `inline` is not as clear to me as to what it is compared to saying `base64`. Similarly, `remote` is not as clear to me as saying `url`.

I get your point on expressing at a higher level being desirable, but to me here it simply makes it less clear what it means.

But I'm happy to leave this as is for now, and we can reevaluate later.

JasonTheAdams:
Just added the unit tests, @felixarntz! All is passing. Back to you for review!

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -38,9 +38,9 @@ class FunctionDeclaration implements WithJsonSchemaInterface
      *
      * @param string $name The name of the function.
      * @param string $description A description of what the function does.
-     * @param mixed $parameters The JSON schema for the function parameters.
+     * @param mixed|null $parameters The JSON schema for the function parameters.
```
felixarntz, line 41: Isn't `null` a subset of `mixed`?

tests/unit/Files/DTO/FileTest.php
```diff
@@ -0,0 +1,270 @@
...
+        $dataUri = 'data:text/plain;base64,' . $base64Data;
+        $overrideMimeType = 'text/html';
+        
+        $file = new File($dataUri, $overrideMimeType);
+        
+        $this->assertEquals(FileTypeEnum::inline(), $file->getFileType());
+        $this->assertEquals($base64Data, $file->getBase64Data());
+        $this->assertEquals($overrideMimeType, $file->getMimeType());
+        $this->assertEquals('data:text/html;base64,' . $base64Data, $file->getDataUri());
+    }
```
felixarntz, line 90: Seeing this now, I wonder whether we should even allow for such an inconsistency. I think it's okay to consider it as an "override MIME type", but it could also be considered an accident that should throw an exception, when the MIME types differ.

felixarntz (approved):
@JasonTheAdams Awesome work!

I didn't review all the individual tests, since it's a lot and I trust Claude Code's thoroughness based on a high-level review, especially with all of them passing.

Two small details of feedback, but nothing critical. I'm super excited to see this land! 🎉

tests/unit/Files/DTO/FileTest.php
JasonTheAdams: Feels like a niche situation. I suspect folks will generally try without providing the MIME type, only providing it if they're giving a Base64 or something that requires it. To take your point from earlier, this feels like the kind of situation where our validation could be interpreted as wrong.

JasonTheAdams:
There are two open comments I'm happy to resolve in a subsequent PR! Merging this bad boy! Thanks for all the great input, folks! 🎉

src/Tools/DTO/FunctionDeclaration.php
JasonTheAdams: Oops! I did fix this but forgot to push. I'll fix this in the next PR.

PR #30, opened 2025-07-28
Adds DTO array transformation and JSON serialization

JasonTheAdams:
Follow up from: https://github.com/WordPress/php-ai-client/pull/28#discussion_r2227228823

This adds support for serializing and unserializing DTOs to and from JSON. It introduces a `WithJsonSerialization` interface that all DTOs implement. This interfaces extends the built-in [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) interface so that `json_encode()` can be used on DTOs. It also adds a `fromJson()` static method so DTOs can be constructed from the same JSON data.


src/Common/Contracts/WithJsonSerialization.php
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Creates an instance from JSON data.
+     *
+     * @since 1.0.0
+     *
+     * @param array<string, mixed> $json The JSON data.
+     * @return self The created instance.
+     */
+    public static function fromJson(array $json);
+}
```
felixarntz: A few thoughts on this approach:
* I think relying on calls to the `jsonSerialize()` method to get an associative array representation of the DTOs is not very intuitive. Keep in mind that calling `json_encode( $dto )` is not the only use-case for that. For that reason, I think we should also have a separate method that uses a more intuitive name (e.g. `toJson(): array`). It could do exactly the same as `jsonSerialize()` (almost, see below), but it makes the API more intuitive.
* I am not sure we should ground our interface in `Json`. Really, it's primary purpose is to transform to and from the "raw data representation". This is then usually used for JSON, but it could also be used for any other things. I would suggest something like `WithDataRepresentation` or `WithRawDataRepresentation`, and the methods would be `toData` and `fromData`, or `toRawData` and `fromRawData`.
* In any case, extending the `JsonSerializable` interface makes sense because that allows to easily JSON encode.
* Here's one nuance about a potential difference that needs to be considered between a `toData` / `toRawData` and `jsonSerialize`: An empty PHP array in JSON can mean either `[]` or `{}`. The DTO has to ensure that it returns the data so that this is correct. For DTOs which allow any empty objects within its data, they must transform empty PHP arrays that technically represent an object to e.g. `new \stdClass()` - this way it's ensured to be an empty object when JSON encoded. This distinction is _only_ relevant for encoding to JSON - not to get raw data which could be used in PHP too. Hence another reason why I think we shouldn't ground the entire interface in JSON, and only `jsonSerialize` should be directly associated with JSON.

felixarntz (commented):
@JasonTheAdams Some early thoughts here before I look more closely. FWIW, the feedback is probably the most crucial for us to discuss, to shape the remainder of the PR accordingly.

src/Common/Contracts/WithJsonSerialization.php
JasonTheAdams: @felixarntz Appreciate the input. I'm not personally a big fan of "data" (e.g. `toData`, `fromData`) as a term in methods because I find it ends up being too vague to be particularly helpful. What if we went the `toArray`, `fromArray` route just to open up its use to thinking about type conversion? Folks can use `(array) $object` easily enough to convert `stdClass` to an array if they want, too.

My only thought with `jsonSerializable` in this direction is that it then depends on each concrete class to have that function that uses `toArray()`. Not a big deal, but a small note.

felixarntz: @JasonTheAdams That's fair, `toArray` and `fromArray` works for me - in fact that's what I'm using in https://github.com/felixarntz/ai-services :)

So then the interface will be `WithArrayRepresentation`?

My only caveat would be that, in terms of JSON / JSON schema, in most cases it would return `object` rather than `array`, but since we are working in PHP here, I think those names still make sense.

For `jsonSerializable`, how about we add a simple trait that is meant for the (common) case of `jsonSerializable` simply calling `toArray`? This way, most DTOs will be able to use that, and any DTOs that need to juggle empty objects can instead manually implement it.

felixarntz: Here's another idea: Since we already have the static `getJsonSchema()`, we could consider a trait with a central implementation of `jsonSerialize` that solves it without DTOs having to worry about it:
* The implementation could call `toArray`, as well as `getJsonSchema`.
* It could then sanitize the resulting array, checking for any arrays within that are empty, and for those check the schema whether it's supposed to be `object`. If so, transform the entry to `new \stdClass()`.

That would solve this concern centrally - which is probably not a bad idea, because of how unintuitive that distinction between JSON `object` and `array` is (easy to forget!).

JasonTheAdams: Nice! I changed to the array-oriented terminology in 18ffdb4

I just dropped the `JsonSerializable` stuff for now, as I'm not quite understanding your second idea.

JasonTheAdams: Ok, I think this is all wrapped up. I added an `AbstractDataTransferObject` which brings this all together, including JSON serialization with empty object handling and array transformation.

src/Common/Contracts/WithArrayTransformationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+     * @since 1.0.0
+     *
+     * @return TArrayShape The array representation.
+     */
+    public function toArray(): array;
+
+    /**
+     * Creates an instance from array data.
+     *
+     * @since 1.0.0
```
felixarntz:
```suggestion
 * @since n.e.x.t
 *
 * @template TArrayShape of array<string, mixed>
 */
interface WithArrayTransformationInterface
{
    /**
     * Converts the object to an array representation.
     *
     * @since n.e.x.t
     *
     * @return TArrayShape The array representation.
     */
    public function toArray(): array;

    /**
     * Creates an instance from array data.
     *
     * @since n.e.x.t
```

src/Common/Contracts/WithArrayTransformationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Common\Contracts;
+
+/**
+ * Interface for objects that support array transformation.
+ *
+ * @since 1.0.0
+ *
+ * @template TArrayShape of array<string, mixed>
```
felixarntz, line 12: What does this do? Never seen it 🤔

src/Files/DTO/File.php
```diff
@@ -377,4 +386,49 @@ public static function getJsonSchema(): array
...
+        $data = [
+            'fileType' => $this->fileType->value,
+            'mimeType' => $this->getMimeType(),
+        ];
+
+        if ($this->fileType->isRemote() && $this->url !== null) {
+            $data['url'] = $this->url;
+        } elseif (!$this->fileType->isRemote() && $this->base64Data !== null) {
+            $data['base64Data'] = $this->base64Data;
+        }
```
felixarntz: The respective second checks here are overkill. It is guaranteed that `$url` is set if `isRemote()`, and it is guaranteed that `$base64Data` is set if `!isRemote()`. If it wasn't, we'd have a problem in our logic.

This is probably a case of Claude Code being too verbose :)

src/Files/DTO/File.php
```diff
@@ -377,4 +386,49 @@ public static function getJsonSchema(): array
...
+            if (!isset($array['url']) || !isset($array['mimeType'])) {
+                throw new \InvalidArgumentException('Remote file requires url and mimeType.');
+            }
+            return new self($array['url'], $array['mimeType']);
+        } else {
+            if (!isset($array['mimeType']) || !isset($array['base64Data'])) {
+                throw new \InvalidArgumentException('Inline file requires mimeType and base64Data.');
+            }
+            return new self($array['base64Data'], $array['mimeType']);
+        }
```
felixarntz: Nit-pick: I think any class references (like `InvalidArgumentException`) should have `use` references on top of the file. It's a best practice to see what's used in a file, and there's also a common PHPCS sniff we could include in the project to enforce that.

src/Messages/DTO/Message.php
```diff
@@ -90,4 +99,47 @@ public static function getJsonSchema(): array
...
+            }, $this->parts),
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     *
+     * @return UserMessage|ModelMessage|SystemMessage
```
felixarntz: Why not just `Message`? Seems odd here to return the individual values when `Message` would be perfectly accurate.

src/Messages/DTO/MessagePart.php
```diff
@@ -202,4 +216,64 @@ public static function getJsonSchema(): array
...
+
+        if ($this->type->isText() && $this->text !== null) {
+            $data['text'] = $this->text;
+        } elseif ($this->type->isFile() && $this->file !== null) {
+            $data['file'] = $this->file->toArray();
+        } elseif ($this->type->isFunctionCall() && $this->functionCall !== null) {
+            $data['functionCall'] = $this->functionCall->toArray();
+        } elseif ($this->type->isFunctionResponse() && $this->functionResponse !== null) {
+            $data['functionResponse'] = $this->functionResponse->toArray();
+        }
```
felixarntz: See my similar comment above on either the respective 1st or the respective 2nd checks here being overkill.

src/Messages/DTO/MessagePart.php
```diff
@@ -202,4 +216,64 @@ public static function getJsonSchema(): array
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): MessagePart
+    {
+        $type = MessagePartTypeEnum::from($array['type']);
```
felixarntz: Since you're already validating, I'd argue the approach this method takes is a bit in the wrong direction, for two reasons:
1. There's no check here whether `$array['type']` is set.
2. I don't think we should check the `type` value anyway, since we can generate the `MessagePart` from purely the respective other key. In other words, let's instead check `isset( $array['text'] )` first, if not check `isset( $array['file'] )`, etc.
3. If no `$array['type']` is set, we can probably ignore that, since it's inferred from the input data anyway. If it's set, I think we should only check that it matches, because if not, that would be somewhat invalid.

src/Messages/DTO/ModelMessage.php
```diff
@@ -14,7 +15,7 @@
  *
  * @since n.e.x.t
  */
-class ModelMessage extends Message
+final class ModelMessage extends Message
```
felixarntz: Why are these classes being made final here?

src/Operations/DTO/GenerativeAiOperation.php
```diff
@@ -15,8 +16,14 @@
...
  * @since n.e.x.t
+ *
+ * @phpstan-import-type GenerativeAiResultArrayShape from GenerativeAiResult
+ *
+ * @phpstan-type GenerativeAiOperationArrayShape array{id: string, state: string, result?: GenerativeAiResultArrayShape}
+ *
+ * @extends AbstractDataValueObject<GenerativeAiOperationArrayShape>
  */
-class GenerativeAiOperation implements OperationInterface
+final class GenerativeAiOperation extends AbstractDataValueObject implements OperationInterface
```
felixarntz: See above, why is this class made final here?

src/Results/DTO/Candidate.php
```diff
@@ -15,8 +15,14 @@
...
  * @since n.e.x.t
+ *
+ * @phpstan-import-type MessageArrayShape from Message
+ *
+ * @phpstan-type CandidateArrayShape array{message: MessageArrayShape, finishReason: string, tokenCount: int}
+ *
+ * @extends AbstractDataValueObject<CandidateArrayShape>
  */
-class Candidate implements WithJsonSchemaInterface
+final class Candidate extends AbstractDataValueObject
```
felixarntz: See above, same question applies here and throughout.

src/Operations/DTO/GenerativeAiOperation.php
```diff
@@ -132,4 +139,41 @@ public static function getJsonSchema(): array
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): GenerativeAiOperation
+    {
+        $state = OperationStateEnum::from($array['state']);
```
felixarntz: To take a random example, here you don't have any validation in place at all. I'm not saying we should validate everything, but I think we should take a consistent approach and apply it throughout the `fromArray` methods.

felixarntz (changes requested):
@JasonTheAdams I think at a high level this looks great, but there are many things in this PR which shows the typical weaknesses of AI assisted code generation - such as overkill checks, or random patterns applied in some places but not others.

I left some of the concrete feedback below, let's discuss as needed, but most importantly decide for an approach and stick with it. This is the type of thing where AI generated code still requires a lot of revision.

src/Common/Contracts/WithArrayTransformationInterface.php
JasonTheAdams: This is how [generics work in PHP](https://phpstan.org/blog/generics-in-php-using-phpdocs). It doesn't have native support, but PhpStan and IDEs support this syntax. I used generics for the array transformation so that PhpStan and IDEs know the expected shape of the array in `toArray()` and `fromArray()`.

src/Files/DTO/File.php
JasonTheAdams: This isn't actually Claude Code, mostly, it's PhpStan. I simplified this a little bit here: f5d4251

The null checks are still necessary from the eyes of PhpStan strict mode because it doesn't recognize that the file type being one thing implies which property is set. May even be worth just ignoring the type and checking which isn't null.

src/Messages/DTO/Message.php
JasonTheAdams: Another PhpStan moment. This is a `Message` function that's returning something more specific, not less. If it was the other way around, `UserMessage` returning a `Message` then that would work.

src/Messages/DTO/MessagePart.php
JasonTheAdams: Same thing as the other thread. This is logical overkill, but from a static analysis perspective it's not obvious that the type is indicative of the property that's set. We could drop the type checks and simply check which property is not null — which I'm fine with. That work for you?

src/Messages/DTO/MessagePart.php
JasonTheAdams: Yeah, in light of the other comments it seems like we're leaning away from checking the type and just checking the properties and array elements? That makes sense to me and does clean things up.

src/Messages/DTO/ModelMessage.php
JasonTheAdams: This may no longer be necessary. It has to do [unsafe usage of static](https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static), which we are using as the return type in the `WithArrayTransformationInterface`. At one point in building this I was using `return static` in a trait, which I'm not now, so I can probably release this constriction.

src/Operations/DTO/GenerativeAiOperation.php
JasonTheAdams: I definitely went happy path on this one. Admittedly I was so focused on getting the types working right that I didn' think about `fromArray` being used more broadly. I'll add some validation rules to the `fromArray()` methods. 👍

src/Messages/DTO/MessagePart.php
felixarntz: +1 from my end to that approach!

src/Common/Contracts/WithArrayTransformationInterface.php
felixarntz: I don't love that you have to redeclare the "raw data" array shape somewhere else in the code for every DTO, but I guess there's nothing you can do about that. +1 for this strict approach regarding array syntax.

src/Files/DTO/File.php
JasonTheAdams: Agreed! I actually prefer this but recently worked with folks who don't, so I reversed my habit. 😆 

Resolved in 0a5ed46

src/Messages/DTO/MessagePart.php
JasonTheAdams: Resolved in 7c13ac1

src/Messages/DTO/MessagePart.php
JasonTheAdams: Resolved in 7c13ac1

src/Messages/DTO/Message.php
felixarntz: Hmm, that doesn't really explain why we cannot hint `Message` here.

It's certainly a nit-pick, but we don't want to get into having to use type hints for `UserMessage|ModelMessage|SystemMessage` - we should always type hint `Message`. Technically, we know that the two are the same, but to PHPStan `UserMessage|ModelMessage|SystemMessage` is more specific, which is precisely why we should hint towards the more generic.

src/Files/DTO/File.php
felixarntz: Yeah, let's do the latter.

Technically this should even be an `if else` though. If PHPStan complains about that, I think we should mark it to be ignored? We know that it's never possible for both of these to be `null`.

src/Messages/DTO/ModelMessage.php
felixarntz: Yeah, let's try to get rid of it. PHPStan is great, but it has limitations. I think we should use caution with some of its recommendations, especially when the only way to address them would be to make actual architecture/API changes.

Sometimes we just have to ignore PHPStan, knowing where it's wrong and we know better :)

src/Files/DTO/File.php
JasonTheAdams: I just updated this here: 5be740a

As a note, I left the `if...elseif...else` for two reasons:
1. Since we know these are mutually exclusive we're reducing unnecessary condition checks
2. Having the exception thrown, even though it should be logically impossible, at least gives us a "something is fundamentally wrong, here" catch.

src/Messages/DTO/Message.php
JasonTheAdams: I kept feeling wrong about this. So I tested it again and it worked. Can we just pretend this conversation didn't happen? 😂 

Resolved in 0569ad4

JasonTheAdams:
Pardon the force push. I accidentally added commits to the wrong branch. 😆

src/Operations/DTO/GenerativeAiOperation.php
JasonTheAdams: Resolved in 25f0491

src/Messages/DTO/ModelMessage.php
JasonTheAdams: Resolved in 59e586c

src/Results/DTO/Candidate.php
JasonTheAdams: Resolved in 59e586c

src/Operations/DTO/GenerativeAiOperation.php
JasonTheAdams: Resolved in 59e586c

JasonTheAdams:
This is ready for review, again, @felixarntz!

src/Files/DTO/File.php
```diff
@@ -15,9 +17,22 @@
  * and handles them appropriately.
  *
  * @since n.e.x.t
+ *
+ * @phpstan-type FileArrayShape array{
+ *     fileType: string,
+ *     url?: string,
+ *     mimeType?: string,
```
felixarntz: I don't think this is optional?

src/Messages/DTO/MessagePart.php
```diff
@@ -151,55 +172,109 @@ public static function getJsonSchema(): array
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_TYPE]);
+
```
felixarntz: Per a previous comment, I don't think we need to validate that this is present. It's inferred from whichever of the other keys is present anyway.

src/Operations/DTO/GenerativeAiOperation.php
```diff
@@ -126,10 +137,49 @@ public static function getJsonSchema(): array
...
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_ID, self::KEY_STATE]);
+
+        $state = OperationStateEnum::from($array[self::KEY_STATE]);
+        $result = null;
+        if (isset($array[self::KEY_RESULT])) {
+            $result = GenerativeAiResult::fromArray($array[self::KEY_RESULT]);
+        }
```
felixarntz: Could validate here that if `KEY_RESULT` is not set but state is "completed", this is wrong and should throw an exception.

src/Results/DTO/Candidate.php
```diff
@@ -101,18 +110,52 @@ public static function getJsonSchema(): array
...
                     'description' => 'The reason generation stopped.',
                 ],
-                'tokenCount' => [
+                self::KEY_TOKEN_COUNT => [
                     'type' => 'integer',
                     'description' => 'The number of tokens in this candidate.',
                 ],
             ],
-            'required' => ['message', 'finishReason', 'tokenCount'],
+            'required' => [self::KEY_MESSAGE, self::KEY_FINISH_REASON, self::KEY_TOKEN_COUNT],
```
felixarntz, line 124: I'm not sure every provider returns a `token_count` on an individual candidate? Should we leave that optional? Or would we prefer consistency and either set to `0` or try some estimation if it's not provided?

src/Tools/DTO/FunctionCall.php
```diff
@@ -98,31 +105,71 @@ public static function getJsonSchema(): array
...
+                    'required' => [self::KEY_ID],
                 ],
                 [
-                    'required' => ['name'],
+                    'required' => [self::KEY_NAME],
                 ],
                 [
-                    'required' => ['id', 'name'],
+                    'required' => [self::KEY_ID, self::KEY_NAME],
                 ],
```
felixarntz: I realize this was added in the previous PR, but I think it's unnecessary, or could even cause problems?
* Unnecessary because the previous two already express requirements - if one is required, the other can still be there but is optional.
* Possibly problematic because the previous two are a subset of this one, so schema validation logic may get confused because it's not clear which of the `oneOf` "applies" if both `id` and `name` are present in the data.

src/Tools/DTO/FunctionResponse.php
```diff
@@ -93,20 +100,52 @@ public static function getJsonSchema(): array
...
-                'response' => [
+                self::KEY_RESPONSE => [
                     'type' => ['string', 'number', 'boolean', 'object', 'array', 'null'],
                     'description' => 'The response data from the function.',
                 ],
             ],
-            'required' => ['id', 'name', 'response'],
+            'required' => [self::KEY_ID, self::KEY_NAME, self::KEY_RESPONSE],
+        ];
+    }
```
felixarntz: I must have missed it in the prior PR, but this needs to be a `oneOf`, just like for `FunctionCall`:
* Either `id` or `name` is required, not both. Both _can_ be present, but required is only one of them.
* Only `response` is always required.

We also need to add corresponding validation to the constructor, like in `FunctionCall`.

src/Tools/DTO/FunctionResponse.php
```diff
@@ -93,20 +100,52 @@ public static function getJsonSchema(): array
...
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_ID, self::KEY_NAME, self::KEY_RESPONSE]);
```
felixarntz: See above regarding requirements.

src/Tools/DTO/Tool.php
```diff
@@ -104,32 +118,78 @@ public static function getJsonSchema(): array
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_TYPE]);
```
felixarntz: See my similar note on `MessagePart`, I think we can omit this, since it's inferred from the other data below.

felixarntz (commented):
@JasonTheAdams A few last points, but LGTM mostly.

src/Files/DTO/File.php
JasonTheAdams: It is because the MIME type can be extracted from the Data URI. If it's not provided in the Data URI and is null then it throws an exception.

src/Messages/DTO/MessagePart.php
JasonTheAdams: Ah! Good catch! Resolved in 342c07a

src/Operations/DTO/GenerativeAiOperation.php
JasonTheAdams: Ah yup! Resolved in 8ded94d

src/Files/DTO/File.php
felixarntz: Right. But I think it does always need to be present in the array representation. It's similar to e.g. `type` for `MessagePart`: You don't have to provide it as input for the DTO to work, but the full array representation MUST have it.

Also because we sanitize the potential data URI to become _just_ the base64 data - at that point not having MIME type would obviously be wrong.

The thing about allowing to omit the MIME type (or to omit the `type` when constructing a `MessagePart`) is purely about being "forgiving" towards developers. No need to validate something that also works correctly without validation. But whenever the SDK itself is responsible (e.g. transforming a DTO into its array representation), these fields MUST be present - the same way that we allow passing a data URI into `File`, but the "sanitized `base64Data` value will never be a data URI.

src/Results/DTO/Candidate.php
JasonTheAdams: Ohh, good question. How would token count be used downstream? Would folks be more likely to prefer to have something? Or would it be important to know that the  model didn't provide one? The use case isn't clear enough for me to make a suggestion. 🤔

src/Tools/DTO/FunctionCall.php
JasonTheAdams: Yeah, I think this is unnecessary. All we're really trying to validate is that at least one exists. If one does it's fine for the other to be present. I don't think we need that. Good catch.

Resolved in 6b48bef

src/Tools/DTO/FunctionResponse.php
JasonTheAdams: Resolved in 3eb5279

src/Tools/DTO/FunctionResponse.php
JasonTheAdams: Resolved in 3eb5279

src/Tools/DTO/Tool.php
JasonTheAdams: Good catch! Resolved in 7d53157

src/Files/DTO/File.php
JasonTheAdams: Honestly, this particular array shape is not a big deal; it's only used for PhpStan to analyze `toArray()` and `fromArray()`. What's most important is our validation and the property is required. I just checked and it passes PhpStan either way, so I'll make it required. 👍 

Resolved in 75a8034

src/Results/DTO/Candidate.php
felixarntz: I'm not sure yet myself 😆 
Let's leave this for a later discussion, if it becomes more relevant.

felixarntz (approved):
Awesome stuff 🎉

PR #31, opened 2025-07-28
Adds Extender API DTOs

JasonTheAdams:
Depends on #30 

This introduces the DTOs distinct to the Extender API. In the domains, this is the Providers and Models areas.

### DIFF
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -23,7 +23,7 @@ use WordPress\AiClient\Providers\Enums\ProviderTypeEnum;
  *
  * @extends AbstractDataValueObject<ProviderMetadataArrayShape>
  */
-final class ProviderMetadata extends AbstractDataValueObject
+class ProviderMetadata extends AbstractDataValueObject
 {
     public const KEY_ID = 'id';
     public const KEY_NAME = 'name';
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -25,6 +25,10 @@ use WordPress\AiClient\Providers\Enums\ProviderTypeEnum;
  */
 final class ProviderMetadata extends AbstractDataValueObject
 {
+    public const KEY_ID = 'id';
+    public const KEY_NAME = 'name';
+    public const KEY_TYPE = 'type';
+
     /**
      * @var string The provider's unique identifier.
      */
@@ -102,21 +106,21 @@ final class ProviderMetadata extends AbstractDataValueObject
         return [
             'type' => 'object',
             'properties' => [
-                'id' => [
+                self::KEY_ID => [
                     'type' => 'string',
                     'description' => 'The provider\'s unique identifier.',
                 ],
-                'name' => [
+                self::KEY_NAME => [
                     'type' => 'string',
                     'description' => 'The provider\'s display name.',
                 ],
-                'type' => [
+                self::KEY_TYPE => [
                     'type' => 'string',
                     'enum' => ProviderTypeEnum::getValues(),
                     'description' => 'The provider type (cloud, server, or client).',
                 ],
             ],
-            'required' => ['id', 'name', 'type'],
+            'required' => [self::KEY_ID, self::KEY_NAME, self::KEY_TYPE],
         ];
     }
 
@@ -130,9 +134,9 @@ final class ProviderMetadata extends AbstractDataValueObject
     public function toArray(): array
     {
         return [
-            'id' => $this->id,
-            'name' => $this->name,
-            'type' => $this->type->value,
+            self::KEY_ID => $this->id,
+            self::KEY_NAME => $this->name,
+            self::KEY_TYPE => $this->type->value,
         ];
     }
 
@@ -143,12 +147,12 @@ final class ProviderMetadata extends AbstractDataValueObject
      */
     public static function fromArray(array $array): self
     {
-        static::validateFromArrayData($array, ['id', 'name', 'type']);
+        static::validateFromArrayData($array, [self::KEY_ID, self::KEY_NAME, self::KEY_TYPE]);
 
         return new self(
-            $array['id'],
-            $array['name'],
-            ProviderTypeEnum::from($array['type'])
+            $array[self::KEY_ID],
+            $array[self::KEY_NAME],
+            ProviderTypeEnum::from($array[self::KEY_TYPE])
         );
     }
 }
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -143,6 +143,8 @@ final class ProviderMetadata extends AbstractDataValueObject
      */
     public static function fromArray(array $array): self
     {
+        static::validateFromArrayData($array, ['id', 'name', 'type']);
+
         return new self(
             $array['id'],
             $array['name'],
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -1,0 +1,152 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Providers\DTO;
+
+use WordPress\AiClient\Common\AbstractDataValueObject;
+use WordPress\AiClient\Providers\Enums\ProviderTypeEnum;
+
+/**
+ * Represents metadata about an AI provider.
+ *
+ * This class contains information about an AI provider, including its
+ * unique identifier, display name, and type (cloud, server, or client).
+ *
+ * @since n.e.x.t
+ *
+ * @phpstan-type ProviderMetadataArrayShape array{
+ *     id: string,
+ *     name: string,
+ *     type: string
+ * }
+ *
+ * @extends AbstractDataValueObject<ProviderMetadataArrayShape>
+ */
+final class ProviderMetadata extends AbstractDataValueObject
+{
+    /**
+     * @var string The provider's unique identifier.
+     */
+    protected string $id;
+
+    /**
+     * @var string The provider's display name.
+     */
+    protected string $name;
+
+    /**
+     * @var ProviderTypeEnum The provider type.
+     */
+    protected ProviderTypeEnum $type;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $id The provider's unique identifier.
+     * @param string $name The provider's display name.
+     * @param ProviderTypeEnum $type The provider type.
+     */
+    public function __construct(string $id, string $name, ProviderTypeEnum $type)
+    {
+        $this->id = $id;
+        $this->name = $name;
+        $this->type = $type;
+    }
+
+    /**
+     * Gets the provider's unique identifier.
+     *
+     * @since n.e.x.t
+     *
+     * @return string The provider ID.
+     */
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    /**
+     * Gets the provider's display name.
+     *
+     * @since n.e.x.t
+     *
+     * @return string The provider name.
+     */
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    /**
+     * Gets the provider type.
+     *
+     * @since n.e.x.t
+     *
+     * @return ProviderTypeEnum The provider type.
+     */
+    public function getType(): ProviderTypeEnum
+    {
+        return $this->type;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function getJsonSchema(): array
+    {
+        return [
+            'type' => 'object',
+            'properties' => [
+                'id' => [
+                    'type' => 'string',
+                    'description' => 'The provider\'s unique identifier.',
+                ],
+                'name' => [
+                    'type' => 'string',
+                    'description' => 'The provider\'s display name.',
+                ],
+                'type' => [
+                    'type' => 'string',
+                    'enum' => ProviderTypeEnum::getValues(),
+                    'description' => 'The provider type (cloud, server, or client).',
+                ],
+            ],
+            'required' => ['id', 'name', 'type'],
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     *
+     * @return ProviderMetadataArrayShape
+     */
+    public function toArray(): array
+    {
+        return [
+            'id' => $this->id,
+            'name' => $this->name,
+            'type' => $this->type->value,
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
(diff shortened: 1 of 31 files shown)


src/Providers/DTO/ProviderMetadata.php
```diff
@@ -0,0 +1,152 @@
...
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
+        return new self(
+            $array['id'],
+            $array['name'],
+            ProviderTypeEnum::from($array['type'])
+        );
+    }
```
felixarntz: Still not sure what you were leaning towards in #30, but similar would apply here: Validate presence of the keys or not?

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,721 @@
...
+
+    /**
+     * @var Tool[]|null Tools available to the model.
+     */
+    protected ?array $tools = null;
+
+    /**
+     * @var array<string, mixed>|null Custom provider-specific options.
+     */
+    protected ?array $customOptions = null;
```
felixarntz: I think this should never be `null`, since it's just a catch-all array for unknown options.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,721 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     */
+    public function __construct()
+    {
+        // All properties are optional and set via setters
+    }
```
felixarntz: Why not omit the constructor then? :)

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,721 @@
...
+     * @since n.e.x.t
+     *
+     * @param float $temperature The temperature.
+     * @return self For method chaining.
+     */
+    public function setTemperature(float $temperature): self
+    {
+        $this->temperature = $temperature;
+        return $this;
+    }
```
felixarntz: Let's not use chaining for all these setters. This is not a builder class :)

If we need a `ModelConfigBuilder` (which I'm not sure we do, since it's handled via `PromptBuilder`, but it's worth thinking about), then let's use that. But the `ModelConfig` DTO should behave similar to the other DTOs.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,721 @@
...
+
+        if (isset($array['tools'])) {
+            $config->setTools(array_map(static function (array $toolData): Tool {
+                return Tool::fromArray($toolData);
+            }, $array['tools']));
+        }
+
+        if (isset($array['customOptions'])) {
+            $config->setCustomOptions($array['customOptions']);
+        }
```
felixarntz: Not sure whether I prefer to use a `customOptions` sub key, or rather simply all unknown keys in `$array`.

src/Providers/Models/DTO/RequiredOption.php
```diff
@@ -0,0 +1,131 @@
...
+
+namespace WordPress\AiClient\Providers\Models\DTO;
+
+use WordPress\AiClient\Common\AbstractDataValueObject;
+
+/**
+ * Represents a required configuration option for an AI model.
+ *
+ * This class defines an option that must be set when using a model,
+ * including its name and the required value.
```
felixarntz: This documentation is incorrect, likely a result of Claude Code misunderstanding `RequiredOption`. This is not an option that _must_ be set when using a model, it's an option that the implementing code requires the model to support. We should clarify that.

I think there are other instances in the docs here as well where we simply refer to "required value" etc., would be good to clarify what that actually means.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,721 @@
...
+                'systemInstruction' => [
+                    'type' => 'string',
+                    'description' => 'System instruction for the model.',
+                ],
+                'candidateCount' => [
+                    'type' => 'integer',
+                    'minimum' => 1,
+                    'description' => 'Number of response candidates to generate.',
+                ],
+                'maxTokens' => [
```
felixarntz: I'm only realizing this now, but I think we should add class constants to every DTO for each key used in the DTO's array representation.

Having constants makes it easier to work with the array data, no chance for typos and so on.

For `ModelConfig` this is especially important, because those string values will also be relevant for the `RequiredOption` / `SupportedOption` functionality.

But I think we should add those constants to all DTOs.

felixarntz (commented):
@JasonTheAdams This looks great! A few notes, most of which should be relatively quick to resolve though.

Some feedback may still apply here from #30, and I'd say let's merge #30 first, so then we can finalize this here.

JasonTheAdams:
Pardon the force push. I accidentally added commits to the wrong branch. 😆

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Oh my goodness. Hahah! I apologize, I should not have told you to review this last night before I did a thorough review. I was just excited to have this first draft together before I took off. 😆 

Resolved in 1fbd566

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Resolved in 22b0a11

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Fluent APIs everywhere!!! 🔥 

...resolved in 22b0a11

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Since there are a number of keys already and the potential for that list to grow over time, I'd prefer to keep this to avoid potential collisions.

src/Providers/DTO/ProviderMetadata.php
JasonTheAdams: Resolved in 18092d2

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: I see your point. If this was PHP 8 compatible we could use attributes here. But constants certainly work!

~Any reason we'd make the constants public? Or is private/protected sufficient?~

Ah, you're saying they'd be used elsewhere, so public it is!

JasonTheAdams: Resolved in e1c9e5d

I'll add the DTOs not introduce in this PR in the other one.

src/Providers/Models/DTO/RequiredOption.php
JasonTheAdams: I _believe_ I improved this, but let me know if you have even clearer documentation in mind.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    /**
+     * Sets the custom options.
+     *
+     * @since n.e.x.t
+     *
+     * @param array<string, mixed> $customOptions The custom options.
+     */
+    public function setCustomOptions(array $customOptions): void
```
felixarntz, line 530: Can we also add a `setCustomOption(string $key, mixed $value): void`, as a simple way to set one custom option (without overriding any other custom options that may already be set)?

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,699 @@
...
+ *     frequencyPenalty?: float,
+ *     logprobs?: bool,
+ *     topLogprobs?: int,
+ *     tools?: array<int, ToolArrayShape>,
+ *     customOptions?: array<string, mixed>
+ * }
+ *
+ * @extends AbstractDataValueObject<ModelConfigArrayShape>
+ */
+class ModelConfig extends AbstractDataValueObject
```
felixarntz, line 43: One thing I'm not sure I like about this implementation is that all other DTOs are immutable (no setters, no way to alter after instantiating), but this one isn't.

I suppose it's partially due to this having _so many_ parameters, so it makes sense to take a different approach. But technically the same implications apply, ideally this would be immutable so that classes that process it (like provider implementations) can't modify it.

Let's leave this as is for this PR, but I think we should iterate in a separate PR (not a P0, could also happen a bit later in the near future). My proposal would be:
* Make `ModelConfig` an immutable DTO like the others which receives all its data as an array in the constructor. Basically, you could move the code of the `fromArray` method to the constructor, and make all the setters protected.
* Introduce a `ModelConfigBuilder` which allows to create the config declaratively. Under the hood, it would put together the associative array, and its terminate method would instantiate the `ModelConfig` based on that array.

LMKWYT. But regardless, let's keep it as is for this PR and iterate separately.

src/Providers/Models/DTO/ModelMetadata.php
```diff
@@ -0,0 +1,208 @@
...
+            self::KEY_ID => $this->id,
+            self::KEY_NAME => $this->name,
+            self::KEY_SUPPORTED_CAPABILITIES => array_values(array_map(
+                static fn(CapabilityEnum $capability): string => $capability->value,
+                $this->supportedCapabilities
+            )),
+            self::KEY_SUPPORTED_OPTIONS => array_values(array_map(
+                static fn(SupportedOption $option): array => $option->toArray(),
+                $this->supportedOptions
+            )),
```
felixarntz: Why do these use `array_values()`? The two properties are documented to be indexed arrays anyway.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -0,0 +1,699 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @return Tool[]|null The tools.
+     */
+    public function getTools(): ?array
+    {
+        return $this->tools;
+    }
+
```
felixarntz, line 453: This class is missing support for the following options which are specified in the architecture diagram:
* `outputMimeType: string`
* `outputSchema: array< string, mixed >` (JSON schema --> if provided, should automatically set `outputMimeType: "application/json"`)

src/Providers/Models/DTO/SupportedOption.php
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $name The option name.
+     * @param mixed[] $supportedValues The supported values for this option.
+     */
+    public function __construct(string $name, array $supportedValues)
```
felixarntz: `$supportedValues` should be optional, as it doesn't apply for any option. Some options, like `outputMimeType` support specific values, so basically they're an enum. Other options (e.g. `temperature`) are numeric, so for that this is not relevant.

We should probably think about additional validation mechanisms when we get there, e.g. for numeric options they could specify a min and max value they support. But not sure whether we want to get into the weeds of that right here just yet.

For now, I'd say let's make `$supportedValues` optional, and in `isSupportedValue` simply return `true` if that array is empty / not specified.

src/Providers/Models/DTO/SupportedOption.php
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_NAME, self::KEY_SUPPORTED_VALUES]);
```
felixarntz: See above regarding the latter parameter.

felixarntz (commented):
@JasonTheAdams Looks great - a few last bits and then I think we can merge this!

src/Providers/Models/DTO/ModelMetadata.php
JasonTheAdams: I wondered the same thing and even argued with Claude about this. Hahah! Claude's argument is that because the user passes the arrays directly we can't actually guarantee that they're lists and not associative. One option here would be to add an [array_is_list](https://www.php.net/manual/en/function.array-is-list.php) polyfill and validate the arrays in the constructor.

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: I agree. This should be immutable and I like the idea of a `ModelConfigBuilder`. Happy to add this in a subsequent PR!

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Resolved in 7025568

src/Providers/Models/DTO/ModelMetadata.php
felixarntz: Are you sure those are the only two places in the codebase that are supposed be indexed arrays but we don't validate them? I guess my point is, for now let's just go with the simple approach: no validation and expect indexed arrays to be indexed arrays.

I think polyfills for things like that would be great (also `str_contains` etc.), but let's add those via a separate PR purely focused on setting them up. We could also combine that with adding validation for any input arrays that should be indexed arrays.

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Tentatively resolved in 02d3433

Only question is whether we want `setOutputSchema()` to set the MIME type to `application/json` _regardless_ of whether it's been previously set, or to only set it if not previously set?

felixarntz: Good point, let's only set it when it wasn't previously set.

src/Providers/Models/DTO/SupportedOption.php
JasonTheAdams: Hmm, good point. Some things are flags, others have specific validation rules for the value. It kind of feels like this may be a better Interface or Abstract Class than a generic class. That way each option can provide it's own rules. 🤔 

Tentatively resolved in 07df1e7 — simply makes it optional

src/Providers/Models/DTO/SupportedOption.php
JasonTheAdams: Tentatively resolved in 07df1e7

src/Providers/Models/DTO/SupportedOption.php
felixarntz: Agreed. Let's revise this later when we get towards that problem space.

src/Providers/Models/DTO/ModelMetadata.php
JasonTheAdams: I'm good with that. In any case we'll remove these and handle the validation elsewhere. 👍

JasonTheAdams: Ok, here's the fix: fb76c99

I had to suppress some phpstan errors for now because it's angry that we can't guarantee that the arrays are lists. I'll address this in a subsequent PR.

JasonTheAdams: Subsequent PR: #33

felixarntz (approved):
@JasonTheAdams Thank you for the updates, LGTM!

PR #32, opened 2025-07-29
Include contributing docs about branch naming conventions

felixarntz:
This adds a brief documentation sections on branch naming, most importantly which names _not_ to use due to being reserved for special branches.


JasonTheAdams (approved):
Thanks for adding this!

PR #34, opened 2025-08-04
Add isType() fallback method to MimeType value object

JosephGabito:
### What's changed

- New method: isType(string $prefix): bool added to `WordPress\AiClient\Files\ValueObjects\MimeType`.
- Refactor: `isImage()`, `isAudio()`, `isVideo()`, and `isText()` now delegate to isType() instead of repeating strpos() logic.
- Tests: Added `testIsType()` to cover positive and negative cases; all existing tests remain green.

### Why it matters
- Keeps the convenient is*() helpers for common categories while introducing a general-purpose fallback for any non-enumerated MIME groups (e.g., fonts, scripts) without bloating the API.
- Centralizes prefix matching, reducing duplication and clarifying intent.
- Lays groundwork for future category checks with no breaking changes.
- All tests including the new MimeTypeTest::testIsType() should pass.

Open to feedback, happy to adjust if we'd like any tweaks

### DIFF
--- a/src/Files/ValueObjects/MimeType.php
+++ b/src/Files/ValueObjects/MimeType.php
@@ -179,7 +179,7 @@ final class MimeType
      * @param string $mimeType The MIME type to check.
      * @return bool True if the MIME type has the specified MIME type.
      */
-    public function hasMimeType(string $mimeType): bool
+    public function isType(string $mimeType): bool
     {
         return strpos($this->value, $mimeType . '/') === 0;
     }
@@ -193,7 +193,7 @@ final class MimeType
      */
     public function isImage(): bool
     {
-        return $this->hasMimeType('image');
+        return $this->isType('image');
     }
 
     /**
@@ -205,7 +205,7 @@ final class MimeType
      */
     public function isAudio(): bool
     {
-        return $this->hasMimeType('audio');
+        return $this->isType('audio');
     }
 
     /**
@@ -217,7 +217,7 @@ final class MimeType
      */
     public function isVideo(): bool
     {
-        return $this->hasMimeType('video');
+        return $this->isType('video');
     }
 
     /**
@@ -229,7 +229,7 @@ final class MimeType
      */
     public function isText(): bool
     {
-        return $this->hasMimeType('text');
+        return $this->isType('text');
     }
 
     /**
--- a/src/Files/ValueObjects/MimeType.php
+++ b/src/Files/ValueObjects/MimeType.php
@@ -170,6 +170,21 @@ final class MimeType
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Files/ValueObjects/MimeType.php
```diff
@@ -169,6 +169,21 @@ public static function isValid(string $mimeType): bool
...
+     *
+     * @since n.e.x.t
+     *
+     * @param string $mimeType The MIME type to check.
+     * @return bool True if the MIME type has the specified MIME type.
+     */
+    public function hasMimeType(string $mimeType): bool
+    {
+        return strpos($this->value, $mimeType . '/') === 0;
+    }
```
JasonTheAdams: The VO is intended to indicate a single value, so it's more accurate to say that the value _is_ a certain type rather than "has" a type.

Also, the internal value is stored as lowercase, so let's make sure the `$mimeType` is also made lowercase. Good to add this to the tests, too.
```suggestion
    /**
     * Checks if this MIME type is a specific MIME type, the type being the prefixed part (e.g. "audio/").
     *
     * @since n.e.x.t
     *
     * @param string $mimeType The MIME type to check.
     * @return bool True if the MIME type is the specified MIME type.
     */
    public function isType(string $mimeType): bool
    {
        return strpos($this->value, strtolower($mimeType) . '/') === 0;
    }
```

JasonTheAdams: Please note that I didn't adjust this in all other places.

JasonTheAdams (changes requested):
Thanks, @JosephGabito! Added a suggestion.

src/Files/ValueObjects/MimeType.php
JosephGabito: Thanks for the review, @JasonTheAdams 

✅ Renamed hasMimeType() to isType().
✅ Added strtolower() for the argument, and updated the test.

Let me know if you'd like any other tweaks!

src/Files/ValueObjects/MimeType.php
```diff
@@ -169,6 +169,21 @@ public static function isValid(string $mimeType): bool
         );
     }
 
+    /**
+     * Checks if this MIME type has a specific MIME type.
```
JasonTheAdams:
```suggestion
     * Checks if this MIME type belongs to a type group.
```

JasonTheAdams (approved):
This is great! Nice improvement! Thank you!

JosephGabito:
@JasonTheAdams 

Appreciate the fast review and the pointer on naming. Glad this helps tighten things up. 💪

PR #35, opened 2025-08-04
Adds Provider interfaces

JasonTheAdams:
This adds the contracts (interfaces) for the Providers domain. It omits:
* Embeddings, since we're deferring this feature
* HTTP interface(s)
* Authentication interface(s)

The latter two will be coming in a subsequent PR.


JasonTheAdams:
I did add some methods to `ModelMetadata` for comparing against requirements. Perhaps better to move into its own PR... but also happy to just sneak in here. 😁

src/Providers/Contracts/ProviderInterface.php
```diff
@@ -0,0 +1,60 @@
...
+     * Creates a model instance.
+     *
+     * @since n.e.x.t
+     *
+     * @param string                          $modelId     Model identifier.
+     * @param ModelConfig|array<string,mixed> $modelConfig Model configuration.
+     * @return ModelInterface Model instance.
+     * @throws InvalidArgumentException If model not found or configuration invalid.
+     */
+    public static function model(string $modelId, $modelConfig = []): ModelInterface;
```
felixarntz: I think we should revise this to only accept `ModelConfig` (even though in the architecture it is so far outlined to accept both). It'll typically be passed from the `PromptBuilder` instance anyway, so allowing an easy workaround of an array here isn't needed, and it would put an additional (albeit small) burden on the provider implementation to handle transforming it - which seems unnecessary.

We could adjust that in the architecture diagrams as well.

```suggestion
     * @param ?ModelConfig $modelConfig Model configuration.
     * @return ModelInterface Model instance.
     * @throws InvalidArgumentException If model not found or configuration invalid.
     */
    public static function model(string $modelId, ?ModelConfig $modelConfig = null): ModelInterface;
```

src/Providers/Models/Contracts/WithGenerativeAiOperationsInterface.php
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Interface for models that support generative AI operations.
+ *
+ * Provides methods to retrieve and manage long-running generative AI
+ * operations such as text, image, or speech generation.
+ *
+ * @since n.e.x.t
+ */
+interface WithGenerativeAiOperationsInterface
```
felixarntz: I'm starting to question whether this approach makes sense: Looking up an existing operation is technically subject to the provider support, not model-specific, so maybe this shouldn't be an interface for _models_ to implement?

Maybe we could instead have a `ProviderOperationsHandlerInterface` (something like that) with the relevant instance methods (like `getOperation`), and another `ProviderWithOperationsInterface` which the actual provider class would implement, with a single static method `operationsHandler()` which would return the provider-specific instance. That would align well with how the rest works, but keep it optional, as not every provider would need this.

LMKWYT. If this needs more thinking and more discussion, we could leave it out of this PR entirely, as we'll only implement long-running operations later anyway. The key features for the MVP and the big three providers don't need it.

src/Providers/Models/DTO/ModelMetadata.php
```diff
@@ -189,6 +189,60 @@ public function toArray(): array
...
+     * @since n.e.x.t
+     *
+     * @param ModelRequirements $requirements The requirements to check against.
+     * @return bool True if the model meets all requirements, false otherwise.
+     */
+    public function meetsRequirements(ModelRequirements $requirements): bool
+    {
+        // Check if all required capabilities are supported
+        foreach ($requirements->getRequiredCapabilities() as $requiredCapability) {
+            if (!in_array($requiredCapability, $this->supportedCapabilities, true)) {
```
felixarntz: To speed up this lookup, can we use a map? Either by adjusting the existing internal class property, or creating another one just for the lookup. Using `isset` on a bunch of items is more efficient than `in_array`.

src/Providers/Models/DTO/ModelMetadata.php
```diff
@@ -189,6 +189,60 @@ public function toArray(): array
...
+     *
+     * @since n.e.x.t
+     *
+     * @param string $name The option name to find.
+     * @return SupportedOption|null The supported option, or null if not found.
+     */
+    private function findSupportedOption(string $name): ?SupportedOption
+    {
+        foreach ($this->supportedOptions as $supportedOption) {
+            if ($supportedOption->getName() === $name) {
```
felixarntz: See above. Maybe we can have a lookup map by name for these internally in the class?

src/Providers/Models/TextGeneration/Contracts/TextGenerationModelInterface.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Streams text generation from a prompt.
+     *
+     * @since n.e.x.t
+     *
+     * @param list<Message> $prompt Array of messages containing the text generation prompt.
+     * @return Generator<GenerativeAiResult> Generator yielding partial results.
+     */
+    public function streamGenerateTextResult(array $prompt): Generator;
```
felixarntz, line 38: Since we don't _likely_ want to implement streaming for the MVP yet (due to additional complexity), we may want to leave this out for now.

That being said, having this already in place clarifies the intention and leaves the decision to implement it or not at the provider level, so I kind of like keeping it despite the above. Initially we could just have it throw a "Not implemented" exception, until we get to it.

WDYT?

felixarntz (commented):
@JasonTheAdams This looks almost good to go, just a few small points of feedback.

I like the addition of `ModelMetadata::meetsRequirements` 👍

src/Providers/Contracts/ProviderInterface.php
JasonTheAdams: Agreed! 👍

src/Providers/Models/Contracts/WithGenerativeAiOperationsInterface.php
JasonTheAdams: You know more than I do in this case, but are we confident that looking up operations will always be on a per-provider and not per-model basis? I'd hate to make assumptions that back us into a corner where a provider exists with varying model operational abilities.

I do agree, though, that we shouldn't spend too much time on this  right now.

src/Providers/Models/DTO/ModelMetadata.php
JasonTheAdams: Agreed! Since the DTO is immutable, I went with extra map properties which keeps the code clean.

Resolved in c37dabb

src/Providers/Models/DTO/ModelMetadata.php
JasonTheAdams: Resolved in c37dabb

src/Providers/Models/TextGeneration/Contracts/TextGenerationModelInterface.php
JasonTheAdams: I'm good with leaving it and throwing an exception. It will be asked about, otherwise. 😆

src/Providers/Models/Contracts/WithGenerativeAiOperationsInterface.php
felixarntz: All existing long-running operation implementations in APIs I've seen are model independent - i.e. tied to the entire provider, not specific models. It also makes sense conceptually because the model is only relevant once you trigger/start the operation. At that point you pass all the same information you'd pass to a regular request (model ID, model config etc.), but after that you just get an operation ID back that you can look up provider-wide - no need to remember which model was used at that point.

From that perspective, I'd suggest we go with what I mentioned in the previous comment:
* `ProviderOperationsHandlerInterface` with (non-static) method `getOperation` (just like the current interface)
* `ProviderWithOperationsHandlerInterface` with static method `operationsHandler(): ProviderOperationsHandlerInterface`

Would you be okay replacing this current interface with that? This way we have a starting point in place for operations, which IMO makes sense for a comprehensive picture (similar to https://github.com/WordPress/php-ai-client/pull/35#discussion_r2255093346), even if we'll only use it later.

felixarntz (commented):
@JasonTheAdams Looks all good, pending only https://github.com/WordPress/php-ai-client/pull/35#discussion_r2255083630.

src/Providers/Models/Contracts/WithGenerativeAiOperationsInterface.php
JasonTheAdams: Gotcha! Appreciate the insight! I resolved this in 5b46445

felixarntz (approved):
Thanks, LGTM!

PR #37, opened 2025-08-05
Replace mixed constructor with abstract class pattern for MessagePart

JosephGabito:
## Overview

This PR refactors the `MessagePart` DTO to improve type safety, simplify the codebase, and make the system easier to extend. It replaces the previous mixed constructor approach with a value object-based architecture, while keeping everything backward compatible. All relevant tests have been updated.

---

## Problem

- The constructor accepted `mixed $content`, requiring manual type checks
- Different content types were stored in separate properties (`$text`, `$file`, etc.)
- Branching logic (`if/else`) was needed to determine the content type
- Adding new types meant editing several areas of the code

---

## Solution

- A base `MessageContent` class defines common behavior and default accessors
- Concrete content types like `TextContent`, `FileContent`, etc. implement only what they need
- `MessagePart` delegates to a single content object, rather than managing multiple properties
- A factory in `fromArray()` handles content deserialization
- All existing tests have been updated to reflect the new structure

---

### Before
```php
public function __construct($content)
{
    if (is_string($content)) {
        $this->type = MessagePartTypeEnum::text();
        $this->text = $content;
    } elseif ($content instanceof File) {
        $this->type = MessagePartTypeEnum::file();
        $this->file = $content;
    } elseif ($content instanceof FunctionCall) {
        $this->type = MessagePartTypeEnum::functionCall();
        $this->functionCall = $content;
    } elseif ($content instanceof FunctionResponse) {
// ...
````

### After

```php
public function __construct(MessageContent $content)
{
    $this->content = $content;
}
```

## Why Abstract Class Instead of Interface?

Using an interface would require every content type to implement all possible accessors (`getText()`, `getFile()`, etc.), even if they return `null`. This creates:

- Unnecessary boilerplate in each implementation
- Tight coupling between unrelated content types
- Interface changes every time a new content type is added

By using an abstract class, we:

- Provide default `null` implementations for optional methods
- Allow each content type to override only what it needs
- Avoid touching shared contracts when extending the system

This keeps the design simpler, more flexible, and easier to maintain.

---

## Benefits

* ✅ Type-safe constructor
* ✅ Simpler, smaller classes with clear responsibilities
* ✅ Easier to add new content types in the future
* ✅ No more branching logic or duplicated getters
* ✅ Fully backwards compatible via `fromArray()`
* ✅ All tests passing (531 tests, 2,371 assertions)
* ✅ All PHPStan warnings resolved

---

### Added

* `TextContent.php`
* `FileContent.php`
* `FunctionCallContent.php`
* `FunctionResponseContent.php`
* `MessageContent.php` (base class)

### Modified

* `MessagePart.php` - updated to use new structure
* Relevant test classes — updated for new architecture

---

## Notes

Appreciate any feedback on this change. Happy to adjust anything to better fit the overall direction or future plans.


felixarntz:
@JosephGabito Thank you for opening this PR. I'm not sure yet that this is something we'll want to do though. The current architecture is built with portability in mind, with a focus on DTOs that can easily be expressed in other formats like JSON or other languages like TypeScript, all to have a uniform API across the different layers.

I generally understand and agree with the points you're making regarding use of abstract classes, but I think for this concrete case the implementation as it is is more aligned with the above philosophy and easier to follow. The manual type checks are certainly something you can argue against from a software architecture "cleanliness" perspective, but it works flawlessly, is pragmatic and simple.

I think introducing separate classes just to wrap the different types of message part content, while looking more elegant, adds more complexity in practice.

Aside: In general, for larger proposed changes like this opening an issue to discuss first would be ideal, so that you don't spend too much time writing code without there being consensus yet.

JosephGabito:
Thanks for the thoughtful feedback, Felix 🙌

That all makes sense. I understand the focus on portability and consistency across layers. I approached this from a pure PHP design lens, but I see now that alignment across systems needs to come first.

I'll close this out for now. Next time, I'll raise an issue first before proposing changes at this level. 

Appreciate the perspective.

Issue #4, opened 2025-06-18
Abstracting away the need to manually maintain tons of provider and model metadata


felixarntz:
One pain point I've had while maintaining [AI Services](https://wordpress.org/plugins/ai-services/) is to keep up with provider details and their model metadata, especially as the available models and their metadata (e.g. maximum supported input and output tokens, capabilities like function calling) change all the time.

Unfortunately, the majority of providers do not provide structured information about these aspects in their APIs. Most providers barely provide any model information other than the identifier and human readable name - not even something as high level as whether a model is for text generation or image generation, for example.

This means that, the more providers and models we'll eventually support, the worse the maintenance burden becomes. At some point it's probably unreasonable for the maintainers of this project, however many it will be, to be able to keep up with all the model changes across all the supported providers.

When chatting to [Jamie](https://www.drupal.org/u/yautja_cetanu) and [Marcus](https://www.drupal.org/u/marcus_johansson) from the Drupal AI side, they also raised this as a severe problem they've been facing. They shared https://models.litellm.ai/, which might be the only source of granular model information across providers that's publicly available. If we could somehow implement infrastructure to consume that data to help with model discovery based on capabilities, it would severely reduce the maintenance burden.

Let's think about how we could make that a reality. Potentially this is something that could be built into this SDK, or it could be a separate package purely focused on the data which this SDK could consume.

Jameswlepage:
Interesting; today I also came across [models.dev](https://models.dev), which is an open source project intended to be a reference for model providers. Though the LiteLLM offering is definitely more complete and featured.

OpenRouter, at https://openrouter.ai/api/v1/models also has a pretty impressive list of models that we may be able to use here...

JasonTheAdams:
I want to make sure I'm grasping this, @felixarntz. Is the hope that the client would be able to somehow infer the capabilities and limitations of a model in some way?

In my mind I've thought of this so far as two properties of the models:

1. Meta data
2. Supports

Something like this:
```php
<?php
$limits = [
  'max_tokens' => 16384,
 'max_input_tokens' => 128000,
 // ...
];

$supports = [
  // imagining a ServiceFeatures enum
  ServiceFeatures::PDF_INPUT,
  ServiceFeatures::FUNCTION_CALLING,
  // ...
];

$servicesRegistrar->registerService('gpt-4o', $limits, $supports);
```

[...]

PR #39, opened 2025-08-05
Provider base and implementation for Anthropic, Google, and OpenAI via common abstractions

felixarntz:
This PR provides basic implementations for text generation for Anthropic, Google, and OpenAI, all based on their respective OpenAI API compatible REST API endpoints. While Anthropic and Google provide their alternative API specifications that they generally recommend, focusing on their OpenAI API compatible endpoints first is a more pragmatic approach that will get us up to speed more quickly. We can reconsider using their recommended APIs later, if there is any additional value from doing that.

For documentation on the relevant OpenAI API specifications for the endpoints this implements support for, see:
* https://platform.openai.com/docs/api-reference/models/list
* https://platform.openai.com/docs/api-reference/chat/create

The approach is as follows:
* Implement abstract classes for API based providers.
* Implement abstract classes for OpenAI API compatible providers (extending the respective classes for API based providers).
* Implement Anthropic, Google, and OpenAI providers based on these classes.
    * Due to the abstraction, their implementations are quite simple. The only heavy lifting is about the model metadata, which (at least for now) needs to be hand-maintained for the most part, since their APIs barely provide any details about what each model supports. See #4 for a potential long-term solution to simplify that.
    * It's worth noting that at least this is only important for model discovery. If the developer knows which model to use, the part about what each model supports doesn't matter, as it's not validated inside the model implementations in any way. The API itself would already validate and error as needed, and due to the complexity as well as allowing to always use the latest API capabilities, implementing this in our own code is not worth the tradeoff.
* Hook up central interface implementations for `HttpTransporter` and `RequestAuthentication` to the provider classes, orchestrated via `ProviderRegistry`. This ensures under the hood that every relevant provider class has the necessary implementations set, without the provider class implementation itself having to worry about that.

A subsequent PR will be opened to implement image generation for Google and OpenAI (Anthropic doesn't support that).


felixarntz:
@JasonTheAdams I implemented this for an early reference. Don't try it, it doesn't work - for that reason, also ignore any of the PHPCS / PHPStan errors :)
But please feel free to review and leave early questions / feedback.

src/Results/DTO/Candidate.php
JasonTheAdams, line 1: Should we move these changes to a separate PR so they don't wait for this? Seems like working on this confirmed we don't want `tokenCount`? 😄

src/Providers/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -0,0 +1,105 @@
...
+     * @return bool True if there is metadata for the model.
+     */
+    final public function hasModelMetadata(string $modelId): bool
+    {
+        try {
+            $this->getModelMetadata();
+        } catch (InvalidArgumentException $e) {
+            return false;
+        }
+        return true;
```
JasonTheAdams: This feels like more overhead than necessary for what would otherwise be:
```php
return isset($this->getModelMetadataMap()[$modelId]);
```

src/Providers/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    final public function getModelMetadata(string $modelId): ModelMetadata
+    {
+        $modelsMetadata = $this->getModelMetadataMap();
+        if (!isset($modelsMetadata[$modelId])) {
+            throw new InvalidArgumentException(
+                sprintf('No model with ID %s was found in the provider', $modelId)
+            );
+        }
+        return $modelsMetadata[$modelId];
```
JasonTheAdams: If we simplify the has function, per my last comment, then we can simplify this, too:
```php
if (!$this->has($modelId)) {
	throw new InvalidArgumentException(
		sprintf('No model with ID %s was found in the provider', $modelId)
	);
}

return $this->getModelMetadataMap()[$modelId];
```

src/Results/DTO/Candidate.php
felixarntz: Good idea, see #42

felixarntz:
@JasonTheAdams Note: For now, I put all the abstract and base classes into `AiClient\Providers`. Some of them may be better suited under `AiClient\Providers\Models`, let's discuss that as we iterate here.

Part of me likes them being in one place as they're so closely related, which goes back to my earlier preference to _not_ separate provider and model code into different namespaces. But I'm happy to stay aligned with our current approach, we just have to figure out how that would work best.

src/Providers/AbstractApiBasedModelMetadataDirectory.php
JasonTheAdams, line 1: I think this would be simpler if the list was passed into the constructor and then stored in list and map properties for later use. Since this is immutable I think that would work nicely.

If we don't want the models passed, the constructor could use an abstract method.

src/Providers/Models/Traits/WithHttpTransporterTrait.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Providers\Models\Traits;
+
+use RuntimeException;
+use WordPress\AiClient\Providers\Contracts\HttpTransporterInterface;
```
JasonTheAdams: So does this just knowingly not exist? Or am I missing something?

felixarntz:
@JasonTheAdams Aside: https://github.com/WordPress/php-ai-client/pull/39/commits/d9b32fcc66e2ca662f1efce2ed460dffc43dbbba shows how complex maintaining the list of models and their capabilities is. The current list here for the OpenAI provider is purely based on my own testing that I did for AI Services, so even that is probably slightly inaccurate or out of date. Realistically, there's just no way for us to always stay on top of all those nuances, especially not for all the different providers.

In any case, that's what's needed for now, until we hopefully find a way to source this from some more reliable external database of sorts.

src/Providers/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response to check.
+     * @throws ResponseException If the response is not successful.
+     */
+    protected function throwIfNotSuccessful(Response $response): void
+    {
+        ResponseUtil::throwIfNotSuccessful($response);
+    }
```
JasonTheAdams: Why not just call `ResponseUtil::throwIfNotSuccessful($response)` where this would be used?

felixarntz:
@JasonTheAdams This is now ready for a full review. Obviously it's a lot, so I advise going either commit by commit, or via reviewing individual areas. FWIW, please ignore the tests first, as that's about ~50% of all the code changes in the PR.

I have updated the PR description to provide some high-level context.

If you have any aspects which you'd suggest to pull out into their own PR to simplify review or merge sooner, please let me know. There are definitely parts we could pull out, but at the same time some of them may not be obvious in what they're used for without seeing it in context here.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function createRequest(HttpMethodEnum $method, string $path, array $headers = [], $data = null): Request
+    {
+        return new Request(
+            $method,
+            AnthropicProvider::BASE_URI . '/' . ltrim($path, '/'),
```
JasonTheAdams, line 58: Would be nice to have a static method in `AnthropicProvider` for this so we don't have to do this concatenation and ltrim in so many places.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
```diff
@@ -0,0 +1,134 @@
...
+            $data
+        );
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function parseResponseToModelMetadataList(Response $response): array
+    {
+        $responseData = $response->getData();
```
JasonTheAdams, line 72: Since we know the array shape, let's specify it here. That way we don't have to do the "@var" tags below, and it's better documented.

src/ProviderImplementations/Anthropic/AnthropicProvider.php
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @since n.e.x.t
+ */
+class AnthropicProvider extends AbstractProvider
+{
+    public const BASE_URI = 'https://api.anthropic.com/v1';
+
+    /**
+     * @inheritDoc
+     */
```
JasonTheAdams: We should be doing this so we can still track when these concrete methods were created
```suggestion
    /**
     * {@inheritDoc}
     * 
     * @since n.e.x.t
     */
```
Same everywhere else... 😅

src/Common/AbstractEnum.php
```diff
@@ -393,4 +394,17 @@ final public function __toString(): string
     {
         return $this->value;
     }
+
+    /**
+     * Converts the enum to a JSON-serializable format.
+     *
+     * @since n.e.x.t
+     *
+     * @return mixed The JSON-serializable representation.
```
JasonTheAdams:
```suggestion
     * @return string The internal enum string value.
```

src/ProviderImplementations/Google/GoogleModelMetadataDirectory.php
```diff
@@ -0,0 +1,224 @@
...
+            $data
+        );
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function parseResponseToModelMetadataList(Response $response): array
+    {
+        $responseData = $response->getData();
```
JasonTheAdams, line 88: Same here that we can provide the array shape to clean up the subsequent typing and help document this

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
```diff
@@ -0,0 +1,213 @@
...
+                    $dalleImageOptions,
+                    $gptImageOptions,
+                    $ttsCapabilities,
+                    $ttsOptions,
+                ): ModelMetadata {
+                    /** @var string $modelId */
+                    $modelId = $modelData['id'];
+                    if (
+                        str_starts_with($modelId, 'dall-e-') ||
+                        str_starts_with($modelId, 'gpt-image-')
```
JasonTheAdams, line 177: What do you think about exploding on the hyphen and doing `in_array` checks instead? That would make it so these parts don't assume the full string shape. Maybe it's fine, though.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Providers\Http\Exception;
```
JasonTheAdams, line 5: The directory name should also be plural.
```suggestion
namespace WordPress\AiClient\Providers\Http\Exceptions;
```

JasonTheAdams (changes requested):

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -0,0 +1,67 @@
...
+     * a ResponseException if the status code indicates an error (i.e., not in the
+     * 2xx range). It also attempts to extract a more detailed error message from
+     * the response body if available.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response to check.
+     * @throws ResponseException If the response is not successful.
+     */
+    public static function throwIfNotSuccessful(Response $response): void
```
JasonTheAdams, line 30: Nit: I prefer naming without negatives for less mental effort
```suggestion
    public static function throwIfFailed(Response $response): void
```

src/Providers/Models/DTO/SupportedOption.php
```diff
@@ -85,6 +85,21 @@ public function isSupportedValue($value): bool
             return true;
         }
 
+        // If the value is an array, consider it a set (i.e. order doesn't matter).
+        if (is_array($value)) {
+            foreach ($this->supportedValues as $supportedValue) {
+                if (!is_array($supportedValue)) {
+                    continue;
+                }
+                sort($value);
```
JasonTheAdams: Should we move this out of the `foreach` so it doesn't sort on every iteration?

src/Providers/Models/AbstractApiBasedModel.php
JasonTheAdams, line 1: I wonder if it would be useful to include the base URL in this so we can make methods (such as path building) based on it.

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,722 @@
...
+        }
+        if ($type->isFile()) {
+            $file = $part->getFile();
+            if (!$file) {
+                // This should be impossible due to class internals, but still needs to be checked.
+                throw new RuntimeException(
+                    'The file typed message part must contain a file.'
+                );
+            }
+            if ($file->getFileType()->isRemote()) {
```
JasonTheAdams:
```suggestion
            if ($file->isRemote()) {
```

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,722 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param MessagePart $part The message part to get the data for.
+     * @return ?array<string, mixed> The data for the message content part, or null if not applicable.
+     * @throws InvalidArgumentException If the message part type or data is unsupported.
+     */
+    protected function getMessagePartContentData(MessagePart $part): ?array
+    {
+        $type = $part->getType();
```
JasonTheAdams: This makes me want to have passthrough methods like we did with `File`, so we can do `MessagePart::isText()` without having to grab the `getType()`.

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,722 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response to check.
+     * @throws ResponseException If the response is not successful.
+     */
+    protected function throwIfNotSuccessful(Response $response): void
+    {
+        ResponseUtil::throwIfNotSuccessful($response);
+    }
```
JasonTheAdams: Why the extra abstraction? Why not just call the utility method where this is used?

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,722 @@
...
+     * Parses the response from the API endpoint to a generative AI result.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The response from the API endpoint.
+     * @return GenerativeAiResult The parsed generative AI result.
+     */
+    protected function parseResponseToGenerativeAiResult(Response $response): GenerativeAiResult
+    {
+        $responseData = $response->getData();
```
JasonTheAdams: As elsewhere, let's provide the array shape here.

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
JasonTheAdams, line 1: Let's add a `MessageDataArrayShape` type so we can clarify the data shape throughout.

src/Providers/AbstractApiBasedModelMetadataDirectory.php
felixarntz: Good point!

src/Providers/AbstractApiBasedModelMetadataDirectory.php
felixarntz: I would push back on this: If you argue above that calling `get()` to determine `has()` is overhead (which I agree with!), then calling `has()` to determine `get()` is equally overhead. I think handling it just internally of the method is most pragmatic from that perspective.

src/Providers/AbstractApiBasedModelMetadataDirectory.php
felixarntz: Hmm I'm not sure I follow. Can you clarify?

src/Providers/Models/Traits/WithHttpTransporterTrait.php
felixarntz: Maybe this was an older comment when things were still inflight waiting for another PR to be merged? This does exist now.

src/Providers/AbstractOpenAiCompatibleModelMetadataDirectory.php
felixarntz: I intentionally made it that way so a developer could decide to override the method and handle this differently. What is considered successful and what the exception message could look like may differ.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
felixarntz: I agree with your thinking, but am wondering what would be the best place for that. I'm inclined to do this kind of thing via separate utility classes, so that it's decoupled from other interfaces. We could just have e.g. a `UrlUtil::join` method that handles this.

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
felixarntz: I prefer this, as it's more precise. I think we talked about that the other day, we can't really make any assumptions about those strings appearing in specific segments.

src/Providers/Http/Exception/ResponseException.php
felixarntz: Not sure about this. Having `Exception` namespaces is pretty common in PHP projects, and it's also aligned with our `DTO` namespaces, which are singular as well.

src/Providers/AbstractApiBasedModelMetadataDirectory.php
felixarntz: Updated in https://github.com/WordPress/php-ai-client/pull/39/commits/39edb57aaed82cb20e91c3df49820beefec3bb03

src/ProviderImplementations/Anthropic/AnthropicProvider.php
felixarntz: Updated in https://github.com/WordPress/php-ai-client/pull/39/commits/c6575f5803640b08f6840e547754114f28e70b1f

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
felixarntz: Updated in https://github.com/WordPress/php-ai-client/pull/39/commits/1c53f39bbbb6e5a64c50f681daa7606362bc269a

src/ProviderImplementations/Google/GoogleModelMetadataDirectory.php
felixarntz: Updated in https://github.com/WordPress/php-ai-client/pull/39/commits/1c53f39bbbb6e5a64c50f681daa7606362bc269a

src/Providers/AbstractApiBasedModelMetadataDirectory.php
JasonTheAdams: My issue wasn't with methods referencing one another; I think that's just reducing redundancy. I more view `get()` as a higher order function than `has()`. That is, it makes sense for `get()` to check whether it _has_ the value in order to return it. But having `has()` infer itself from a failed get feels weird.

That said, it's really not a big deal. It's such a simple check. 😄

src/Providers/AbstractApiBasedModelMetadataDirectory.php
JasonTheAdams: This only made sense when I thought that the models were statically defined. But now I understand that it's the result of a request, so this makes sense the way it's setup. 👍

src/Providers/AbstractOpenAiCompatibleModelMetadataDirectory.php
JasonTheAdams: Ah, I see. May be good to include that in the docblock so folks understand its intention?

src/Providers/Http/Exception/ResponseException.php
JasonTheAdams: My _preference_ is to pluralize to indicate type directories, but not a big deal. I only did `DTO` as a directory since `DataTransferObjects` is so long (which, honestly I wonder if that was a bad idea to use the acronym), and `DTOs` is awkward. 😆

JasonTheAdams (changes requested):
Pushing my comments because GH is being weird.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
JasonTheAdams: See https://github.com/WordPress/php-ai-client/pull/39#discussion_r2302383155

It could be useful to have a class constant that gets overridden or else an abstract static method that provides the base URL so methods like this can exist.

src/Providers/ApiBasedImplementation/GenerateTextApiBasedProviderAvailability.php
```diff
@@ -0,0 +1,70 @@
...
+        if (!($model instanceof TextGenerationModelInterface)) {
+            throw new Exception(
+                'The model class to check provider availability must implement TextGenerationModelInterface.'
+            );
+        }
+        $this->model = $model;
+    }
+
+    /**
+     * @inheritdoc
```
JasonTheAdams: This can also use the `@since` tag.

src/Providers/ApiBasedImplementation/ListModelsApiBasedProviderAvailability.php
```diff
@@ -0,0 +1,50 @@
...
+     * @param ModelMetadataDirectoryInterface $modelMetadataDirectory The model metadata directory to use for checking
+     *                                                                availability.
+     */
+    public function __construct(ModelMetadataDirectoryInterface $modelMetadataDirectory)
+    {
+        $this->modelMetadataDirectory = $modelMetadataDirectory;
+    }
+
+    /**
+     * @inheritdoc
```
JasonTheAdams: Also add `@since` here.

src/Providers/ApiBasedImplementation/ListModelsApiBasedProviderAvailability.php
```diff
@@ -0,0 +1,50 @@
...
+use Exception;
+use WordPress\AiClient\Providers\Contracts\ModelMetadataDirectoryInterface;
+use WordPress\AiClient\Providers\Contracts\ProviderAvailabilityInterface;
+
+/**
+ * Class to check availability for an API-based provider via a test request to the endpoint to list models.
+ *
+ * @since n.e.x.t
+ */
+class ListModelsApiBasedProviderAvailability implements ProviderAvailabilityInterface
```
JasonTheAdams: Per our chat, let's clarify that this class is best used when the API key is required for listing the models since it assumes this as part of how it checks for availability.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Throws an exception if the response is not successful.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response to check.
+     * @throws ResponseException If the response is not successful.
+     */
+    protected function throwIfNotSuccessful(Response $response): void
```
JasonTheAdams, line 72: Let's document this is here for the purpose of being overloaded.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,727 @@
...
+                        'content' => json_encode($functionResponse->getResponse()),
+                        'tool_call_id' => $functionResponse->getId(),
+                    ];
+                }
+                return [
+                    'role' => $this->getMessageRoleString($message->getRole()),
+                    'content' => array_values(array_filter(array_map(
+                        function (MessagePart $part): ?array {
+                            return $this->getMessagePartContentData($part);
+                        },
```
JasonTheAdams: We don't need a middleman function
```suggestion
                        [$this, 'getMessagePartContentData],
```

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,727 @@
...
+                        $messageParts
+                    ))),
+                    'tool_calls' => array_values(array_filter(array_map(
+                        function (MessagePart $part): ?array {
+                            return $this->getMessagePartToolCallData($part);
+                        },
```
JasonTheAdams: Same. Extra function isn't needed.
```suggestion
                        [$this, 'getMessagePartToolCallData],
```

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,727 @@
...
+     * Parses the message from a choice in the API response.
+     *
+     * @since n.e.x.t
+     *
+     * @param array<string, mixed> $messageData The message data from the API response.
+     * @return Message The parsed message.
+     */
+    protected function parseResponseChoiceMessage(array $messageData): Message
+    {
+        $role = isset($messageData['role']) && 'user' === $messageData['role']
```
JasonTheAdams, line 673: Since the roles are used in a couple places, may be good to add `ROLE_USER` (etc.) class constants?

src/Providers/AbstractProvider.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private static array $availabilityCache = [];
+
+    /**
+     * @var array<string, ModelMetadataDirectoryInterface> Cache for model metadata directory per class.
+     */
+    private static array $modelMetadataDirectoryCache = [];
+
+    /**
+     * @inheritdoc
```
JasonTheAdams: Let's add the `@since` tags to this class' methods as well.

src/Providers/ProviderRegistry.php
```diff
@@ -66,6 +83,27 @@ public function registerProvider(string $className): void
             );
         }
 
+        // If there is already a HTTP transporter instance set, hook it up to the provider as needed.
+        try {
+            $httpTransporter = $this->getHttpTransporter();
+            $this->setHttpTransporterForProvider($className, $httpTransporter);
+        } catch (RuntimeException $e) {
+            // Ignore.
+        }
```
JasonTheAdams: Per our chat, let's update the comment here to make it clear that we're avoiding a race condition here, and also setting the transporter later updates all the providers.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,727 @@
...
+     * Parses the response from the API endpoint to a generative AI result.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The response from the API endpoint.
+     * @return GenerativeAiResult The parsed generative AI result.
+     */
+    protected function parseResponseToGenerativeAiResult(Response $response): GenerativeAiResult
+    {
+        $responseData = $response->getData();
```
JasonTheAdams, line 557: Let's define a response shape type for the class to be used throughout, as there are a number of explicit `@var` instances that would be cleaned up by that.

Note that we can define multiple types and combine them for internal arrays of arrays like tools.

src/Providers/ProviderRegistry.php
```diff
@@ -204,7 +247,20 @@ public function getProviderModel(
 
         // Use static method from ProviderInterface
         /** @var class-string<ProviderInterface> $className */
```
JasonTheAdams: This shouldn't be necessary since `resolveProviderClassName` specifies it already.

src/Providers/ProviderRegistry.php
```diff
@@ -204,7 +247,20 @@ public function getProviderModel(
...
+        if ($modelInstance instanceof WithHttpTransporterInterface) {
+            $modelInstance->setHttpTransporter($this->getHttpTransporter());
+        }
+
+        if ($modelInstance instanceof WithRequestAuthenticationInterface) {
+            $requestAuthentication = $this->getProviderRequestAuthentication($className);
+            if ($requestAuthentication !== null) {
+                $modelInstance->setRequestAuthentication($requestAuthentication);
+            }
+        }
```
JasonTheAdams, line 284: Per our chat, let's do this:
1. Create a `ProviderRegistry::bindModelDependencies($model)` method that handles this work
2. Make sure that the model instance in the `PromptBuilder` uses this via the `AiClient`
3. Update the `usingModel` method to support the same parameters as this **as well** (😉) as the existing model instance syntax

JasonTheAdams (changes requested):
Finished going through all the code! Tons of great work! Excited to wrap this up!

src/Providers/ProviderRegistry.php
JasonTheAdams: I did step 1 here: a8dfd8a

src/Providers/ProviderRegistry.php
JasonTheAdams: Resolved in a8dfd8a

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
JasonTheAdams: Resolved in e87d4f36

src/Providers/ProviderRegistry.php
felixarntz: Done in https://github.com/WordPress/php-ai-client/pull/39/commits/50850b6f3beef1a6fcd8ab35969716ad5330e89b

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
felixarntz: Done in https://github.com/WordPress/php-ai-client/pull/39/commits/005fed1575aa1bf0d27564d416df2326b4149ac3

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
```diff
@@ -0,0 +1,143 @@
...
+{
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    public function getRequestAuthentication(): RequestAuthenticationInterface
+    {
+        /*
+         * Since we're calling the primary Google API models endpoint here, we need to use the Google specific API key
```
dkotter: Don't think this comment is accurate since this is the Anthropic class, not Google

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
```diff
@@ -0,0 +1,220 @@
...
+            ]),
+            new SupportedOption(OptionEnum::outputMediaAspectRatio(), ['1:1', '3:2', '2:3']),
+        ];
+        $ttsCapabilities = [
+            CapabilityEnum::textToSpeechConversion(),
+        ];
+        $ttsOptions = [
+            new SupportedOption(OptionEnum::inputModalities(), [[ModalityEnum::text()]]),
+            new SupportedOption(OptionEnum::outputModalities(), [[ModalityEnum::audio()]]),
+            new SupportedOption(OptionEnum::outputMimeType(), ['audio/mpeg', 'audio/ogg', 'audio/wav']),
```
dkotter: I know there are other options that [OpenAI supports](https://platform.openai.com/docs/guides/text-to-speech#supported-output-formats), though maybe there's a reason we don't want to support those here? Like AAC, FLAC, and PCM

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -0,0 +1,752 @@
...
+                ];
+            },
+            $messages
+        );
+
+        if ($systemInstruction) {
+            array_unshift(
+                $messagesParam,
+                [
+                    'role' => 'system',
```
dkotter, line 262: I know OpenAI suggests using `developer` now instead of `system` on their newer models, is that something we should look to support?

dkotter (commented):
This all looks really good to me. Left a few comments though feel free to ignore those if they're not valid. Was giving this a look to see how we would implement this in ClassifAI

src/Providers/Models/AbstractOpenAiCompatibleTextGenerationModel.php
JasonTheAdams: Resolved in 11ccd68

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
felixarntz: Great point! Do you know by any chance which exact MIME types make sense to use here? I suppose `audio/aac`, `audio/flac`, audio/pcm`? I think especially for PCM it's not entirely clear, since it's a low-level format unlike the other types?

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
felixarntz: Great catch, fixed in https://github.com/WordPress/php-ai-client/pull/39/commits/61ac6d301627f3a0a5eff9e6caf63f695440e918

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
felixarntz: I wasn't aware of that, do you have any reference docs about this?

Does it mean that passing `system` will fail on some models? Or it's just discouraged? I mostly am curious whether there's any meaningful difference between them or it's just another value to use for the same thing?

src/Providers/ProviderRegistry.php
felixarntz: Failing tests, but otherwise this is done

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
dkotter: Yeah, seems for AAC it's `audio/aac` and for FLAC it's `audio/flac`, though `audio/x-flac` also seems to sometimes be used (but marked as non-standard so can probably ignore). For PCM, not sure either. Seeing `audio/pcm` mentioned but also seeing a few other options. May be fine to just start with `audio/pcm` (or leave it out entirely), unsure how widely used that format is

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
dkotter: From their [API docs](https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages): 

> Developer-provided instructions that the model should follow, regardless of messages sent by the user. With o1 models and newer, `developer` messages replace the previous `system` messages

In my testing, it doesn't fail if you use `system` instead but may be something they stop supporting in the future

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
felixarntz: I'm just going to include `audio/aac` and `audio/flac` for now - they seem most common of those additional ones, should probably be sufficient.

felixarntz: Fixed in https://github.com/WordPress/php-ai-client/pull/39/commits/d6c3ab2eb2a81891d6f2c8b8f6aa292970c38f19

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
felixarntz: Leaving this as is for now since it doesn't seem to do any harm (yet), but added a `TODO` comment to revise in the future. https://github.com/WordPress/php-ai-client/pull/39/commits/ef3b0ef200d427dcf6be06b08aa688181fd4505a

src/Providers/ApiBasedImplementation/ListModelsApiBasedProviderAvailability.php
felixarntz: Fixed in https://github.com/WordPress/php-ai-client/pull/39/commits/296040a7b67f7b407a45017279a174fb0608c30c

src/Providers/Models/DTO/SupportedOption.php
felixarntz: Good point, fixed in https://github.com/WordPress/php-ai-client/pull/39/commits/3186fbcacb057dbfeb1b3cefc44e26d37af1daf4

JasonTheAdams (approved):
I love it! LGTM!! 🎉

PR #40, opened 2025-08-05
Run PHPCS on unit tests as well

felixarntz:
This PR does two things:
* Ensure tests are covered by PHPCS as well. While not everything necessarily applies for tests, our checks so far are very basic and make sense to be consistently applied.
* Remove the odd test restriction to 8 parallel files, which actually led to only 8 files being linted when running `composer phpcs`.


JasonTheAdams (approved):
Nice and clean! ✨

PR #41, opened 2025-08-05
Corrects DTO abstract class name

JasonTheAdams:
I apparently accidentally named the abstract DTO class "AbstractData**Value**Object" instead of `AbstractDataTransferObject`. 🤦‍♂️ 

This corrects that typo.


felixarntz (approved):
@JasonTheAdams Good catch!

PR #42, opened 2025-08-06
Remove candidate token count

felixarntz:
As previously considered and clearly identified in #39, the `tokenCount` per candidate is not relevant. Therefore this PR removes it.

Cherry picked from #39

Additionally, it contains a minor follow up fix to #41.


JasonTheAdams (approved):
Less is more! 😄

PR #44, opened 2025-08-07
Adds interfaces for provider-based operations

JasonTheAdams:
This was _meant_ to be a part of #35... but I made a git oops and didn't add the files. Hahah! So here are the interfaces for making operations within the provider!


felixarntz (approved):
@JasonTheAdams Thanks, looks great! 😆

PR #45, opened 2025-08-09
Implement additional model config params

felixarntz:
This PR adds model config params for:
* `outputFileType` (`inline` or `remote`)
* `outputMediaOrientation` (`square`, `landscape`, or `portrait`)
* `outputMediaAspectRatio` (a string like `3:2` or `16:9` etc.)

The `outputMediaAspectRatio` is a more specific version of `outputMediaOrientation`, but both are valid to have. Some use-cases may require a specific aspect ratio, while for others purely the orientation (without a precise aspect ratio) may be sufficient.

Aside: Other than common sense for why these are relevant for e.g. image generation, I came across a concrete need while working on #39, specifically in https://github.com/WordPress/php-ai-client/pull/39/commits/d9b32fcc66e2ca662f1efce2ed460dffc43dbbba.

### DIFF
--- a/src/Files/Enums/MediaOrientationEnum.php
+++ b/src/Files/Enums/MediaOrientationEnum.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Files\Enums;
+
+use WordPress\AiClient\Common\AbstractEnum;
+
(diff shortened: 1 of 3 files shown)


JasonTheAdams (approved):
Looks good! Nice additions!

PR #47, opened 2025-08-11
Adds support for Provider HTTP Handling

JasonTheAdams:
Waiting on: #46 

This introduces an HTTP system to the Providers framework so models can send requests to the Provider API.

This is built to not depend on a specific HTTP Client (e.g. Guzzle), but instead is happy to work with any PSR-18 compatible Client. It also uses the [HTTPlug](https://github.com/php-http) set of packages to improve auto-discovery of said client.

When we go to implement this into the WordPress system, this will allow us to make our own client that uses (and navigate the quirks of) the WP request functions.


src/Providers/Contracts/AuthenticationInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Providers\Contracts;
```
felixarntz: I think this should live under `Providers\Http\Contracts` as well as it's directly related to the HTTP requests infrastructure.

src/Providers/Contracts/AuthenticationInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Interface for authentication implementations.
+ *
+ * Provides a contract for different authentication mechanisms
+ * to modify HTTP requests with appropriate credentials.
+ *
+ * @since n.e.x.t
+ */
+interface AuthenticationInterface extends WithJsonSchemaInterface
```
felixarntz: As proposed in #46, I think this should be renamed to `RequestAuthenticationInterface`.

`Authentication` can be for many things. But what this interface is for is specifically to authenticate a HTTP request (different from e.g. validating incoming authentication credentials in a platform).

This also further underlines the direct connection to the HTTP request layer.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,238 @@
...
+     * @since n.e.x.t
+     *
+     * @param string $method The HTTP method.
+     * @param string $uri The request URI.
+     * @param array<string, string|list<string>> $headers The request headers.
+     * @param string|null $body The request body.
+     *
+     * @throws InvalidArgumentException If the method is empty.
+     */
+    public function __construct(string $method, string $uri, array $headers = [], ?string $body = null)
```
felixarntz: See my notes in #46, alternative to `?string $body` we need to allow providing `?array $data`. A few reasons:
* For `GET` requests setting a `$body` does not make sense. We need to support setting key value pairs that would be used in query parameters.
    * In other words, this would technically alter the `$uri` under the hood.
* For `POST` requests setting a `$body` works, but passing key value pairs or raw PHP array data would still be more intuitive. The correct formatting for the request could be handled at the transport layer.
    * In other words, this would technically transform the array data into a `$body` string, encoding it accordingly.

Having a `$body` class property _does_ make sense to me. But how you provide data is a different consideration. It definitely should be possible to provide a raw body as string, but other options are relevant too.

felixarntz: I think we should support something like an `?array $options` parameter here.

There is far more to a request than these 4 (or 5) fields. For example, where would you be able to set a timeout (quite critical for some use-cases), or how to handle redirects, or what the user agent should be?

Related to my comment regarding `$body` and `$data`, I think another useful option to include would be an enum in what shape `$data` is provided. Here's an idea:
* We combine `$body` and `$data` into one variable `$data`, which can be either a `string` or `array`. That already makes sense since you should only ever provide one or the other.
* We have a `data_format` option and a related `RequestDataFormatEnum` which can be one of:
    * `BODY` (this would automatically be implied if `$data` is a string)
    * `QUERY_PARAMS` (this would automatically be implied if `$data` is an array and the method is `GET`)
    * `BODY_PARAMS` (if `$data` is an array, this would mean it would be sent as URL-encoded params in the body)
    * `JSON` (if `$data` is an array, this would mean it would be sent as JSON-encoded data in the body)

WDYT?

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,238 @@
...
+    {
+        if (empty($method)) {
+            throw new InvalidArgumentException('HTTP method cannot be empty.');
+        }
+
+        if (empty($uri)) {
+            throw new InvalidArgumentException('URI cannot be empty.');
+        }
+
+        $this->method = strtoupper($method);
```
felixarntz: I think we should introduce a simple `RequestMethodEnum` for this :)

src/Providers/Http/DTO/Response.php
```diff
@@ -0,0 +1,237 @@
...
+     * Checks if the response indicates success.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if status code is 2xx, false otherwise.
+     */
+    public function isSuccessful(): bool
+    {
+        return $this->statusCode >= 200 && $this->statusCode < 300;
+    }
```
felixarntz: See my notes on #46, maybe these are a bit too opinionated to include here?

1. What's `reasonPhrase` supposed to be?
2. Whether a request is considered successful if between 200-299 or only if 200 exactly doesn't always have a clear answer, it depends on the specific environment and use-case.

felixarntz (commented):
@JasonTheAdams Overall this looks like a great start - it gives me really clean software vibes which I love 😆 

A few points of feedback on the interfaces and DTOs mostly. The high level approach looks awesome.

src/Providers/Http/DTO/Response.php
JasonTheAdams: 1. The "reason phrase" is (per the spec) the human readable text associated with a status code (e.g. 200 "OK").
2. The 200 range is the ["successful" range](https://www.rfc-editor.org/rfc/rfc9110.html#name-successful-2xx). So this is semantically correct. If an implementor needs to do something for a specific code they can check the code directly and not use this.

felixarntz: Follow up reply:
1. In that case, I think this shouldn't be passed, but _if_ we truly need it, it should be automatically set up via a utility. There could be a very basic static lookup class somewhere to look up the reason phrase for a given HTTP status code. It feels unnecessarily verbose to me to require that to be passed. What the reason phrase is is not dependent on the `Response` DTO, it's purely dependent on the HTTP status code `int` - so IMO not something this class should have to store.
2. That's fair, happy to keep it.

src/Providers/Http/DTO/Request.php
JasonTheAdams: Good idea! Resolved in e05e36c

src/Providers/Contracts/AuthenticationInterface.php
JasonTheAdams: Resolved in d3dfa92

src/Providers/Contracts/AuthenticationInterface.php
JasonTheAdams: Resolved in d3dfa92

src/Providers/Http/DTO/Response.php
JasonTheAdams: Yeah, I wondered about that. We could make an enum with a `HttpStatus::getReasonPhrase()` method that gets the reason phrase for a given status (the internal enum value being the integer value). The only reason it does make sense to pass it is that, again per the spec, the reason phrase is loosely coupled with the code. For example, 200 _generally_ means "OK", but if a response set it to "AI Message Received" that's perfectly fine.

All that said, I'm tempted to just remove it for now as I really don't think it's used much.

src/Providers/Http/Contracts/RequestAuthenticationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+    public function authenticate(Request $request): void;
+
+    /**
+     * Returns the JSON schema for this authentication.
+     *
+     * @since 1.0.0
+     *
+     * @return array<string, mixed> The JSON schema.
+     */
+    public static function getJsonSchema(): array;
```
felixarntz: Instead of this method here, I think this interface needs to extend `WithJsonSchemaInterface`?

src/Providers/Http/Contracts/RequestAuthenticationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Providers\Http\Contracts;
+
+use WordPress\AiClient\Providers\Http\DTO\Request;
+
+/**
+ * Interface for HTTP request authentication.
+ *
+ * @since 1.0.0
```
felixarntz: Can you replace all the `@since 1.0.0` in this PR with `@since n.e.x.t`? :)

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,280 @@
...
+     * @var string|null The request body.
+     */
+    protected ?string $body;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param HttpMethodEnum|string $method The HTTP method.
```
felixarntz: Why do we here allow either when in other internal code we always mandate passing the enum itself? I think this needs to be changed for consistency.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,280 @@
...
+     * Gets the HTTP method enum.
+     *
+     * @since n.e.x.t
+     *
+     * @return HttpMethodEnum The HTTP method enum.
+     */
+    public function getMethodEnum(): HttpMethodEnum
+    {
+        return $this->method;
+    }
```
felixarntz: See above, this is inconsistent with our other DTOs: There needs to be only a single method `getMethod(): HttpMethodEnum`.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,280 @@
...
+     * @since n.e.x.t
+     *
+     * @param HttpMethodEnum|string $method The HTTP method.
+     * @param string $uri The request URI.
+     * @param array<string, string|list<string>> $headers The request headers.
+     * @param string|null $body The request body.
+     *
+     * @throws InvalidArgumentException If the URI is empty or method is invalid.
+     */
+    public function __construct($method, string $uri, array $headers = [], ?string $body = null)
```
felixarntz: This is still pending the feedback from both https://github.com/WordPress/php-ai-client/pull/47#discussion_r2268341021 and https://github.com/WordPress/php-ai-client/pull/47#discussion_r2268353854.

src/Providers/Http/DTO/Response.php
felixarntz: +1 for removal, until we see an actual reason to have it.

src/Providers/Http/DTO/Response.php
```diff
@@ -0,0 +1,267 @@
...
+     */
+    protected string $reasonPhrase;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param int $statusCode The HTTP status code.
+     * @param array<string, string|list<string>> $headers The response headers.
```
felixarntz, line 54: This is something I had mentioned somewhere in #46, we should go with one consistent format for headers. I think `array<string, list<string>>` makes sense.

src/Providers/Http/DTO/Response.php
```diff
@@ -0,0 +1,267 @@
...
+     * @return string|list<string>|null The header value(s) or null if not found.
+     */
+    public function getHeader(string $name)
+    {
+        // Case-insensitive header lookup
+        foreach ($this->headers as $key => $value) {
+            if (strcasecmp($key, $name) === 0) {
+                return $value;
+            }
+        }
```
felixarntz: Instead of this `foreach` loop, how about we add a private lookup map property where each key is transformed to lower-case? This way here you could just check for `isset($this->lookup_map[strtolower($name)])`, no loop needed.

src/Providers/Http/HttpTransporterFactory.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Factory for creating HTTP transporters.
+ *
+ * Uses HTTPlug's Discovery component to automatically find
+ * available HTTP clients and factories.
+ *
+ * @since n.e.x.t
+ */
+class HttpTransporterFactory
```
felixarntz, line 19: We haven't used the factory pattern anywhere else. I tend to agree it makes sense here, but just flagging it, since it may beg the question where else (if anywhere) we should use it, and what is our criteria for using it vs just manually instantiating a class? What do you think?

src/Providers/Models/Contracts/WithHttpTransporterInterface.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Interface for models that require HTTP transport capabilities.
+ *
+ * Models implementing this interface can send HTTP requests
+ * to provider APIs using the configured HTTP transporter.
+ *
+ * @since n.e.x.t
+ */
+interface WithHttpTransporterInterface
```
felixarntz: This is a duplicate of the same contract in `Providers\Http\Contracts`? I don't think that's needed - maybe here by accident? :)

felixarntz (changes requested):
@JasonTheAdams Overall this looks quite solid. A few more points of feedback.

src/Providers/Http/DTO/Response.php
JasonTheAdams: I _think_ this resolves this: eb7d5e2

src/Providers/Http/DTO/Response.php
JasonTheAdams: Resolved in 75c69c1

src/Providers/Http/HttpTransporterFactory.php
JasonTheAdams: This made more sense back when I momentarily had the ability to get an async Transporter. Now it's a really thin abstraction. I'd be happy to move this somewhere else. The most useful thing it does is handle the discovery calls. Any suggestions as to where you think this could/should go, @felixarntz?

src/Providers/Models/Contracts/WithHttpTransporterInterface.php
JasonTheAdams: Hah! Resolved in 7c65178

src/Providers/Http/DTO/Response.php
JasonTheAdams: Resolved in 2ddfa0e

src/Providers/Http/Contracts/RequestAuthenticationInterface.php
JasonTheAdams: Oh Claude... Hahah! Resolved in 310ff9e along with some other since tag silliness.

src/Providers/Http/Contracts/RequestAuthenticationInterface.php
JasonTheAdams: Resolved in deafd0e

src/Providers/Http/DTO/Request.php
JasonTheAdams: Good call! This cleans things up! Resolved in 53ccb48

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in f2552c5

src/Providers/Http/DTO/Request.php
JasonTheAdams: Per our discussion, I added a data property but instead use the http method and content type to determine how it's utilized. Resolved in 7935a54

JasonTheAdams: Resolved in 7935a54

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,383 @@
...
+    {
+        if (empty($uri)) {
+            throw new InvalidArgumentException('URI cannot be empty.');
+        }
+
+        $this->method = $method;
+        $this->uri = $uri;
+        $this->headers = $this->normalizeHeaders($headers);
+        $this->headersMap = $this->buildHeadersMap($this->headers);
+        $this->data = $data;
```
felixarntz: How about this?
```suggestion
        if (is_string($data) {
            $this->body = $data;
        } else {
            $this->data = $data;
        }
```

That keeps our properties type safe. If `$data` is a string, it's always a body ready to send, no processing needed.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,383 @@
...
+        return $this->headers[$this->headersMap[$lower]];
+    }
+
+    /**
+     * Gets the first value of a specific header.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $name The header name (case-insensitive).
+     * @return string|null The first header value or null if not found.
```
felixarntz: This isn't correct, it returns all header parts concatenated.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,383 @@
...
+     */
+    public function getHeaderLine(string $name): ?string
```
felixarntz: How about `getHeaderAsString` for the method name?

src/Providers/Http/HttpTransporterFactory.php
felixarntz: I think let's leave it for now. We can revise it later if we find a better place.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,383 @@
...
+     * @since n.e.x.t
+     *
+     * @param array<string, string|list<string>> $headers The headers to normalize.
+     * @return array<string, list<string>> The normalized headers.
+     */
+    private function normalizeHeaders(array $headers): array
+    {
+        $normalized = [];
+        foreach ($headers as $name => $value) {
+            $normalized[$name] = is_array($value) ? array_values($value) : [$value];
```
felixarntz: Technically, we should probably also make sure there are no duplicates here. Ideally, if there is a `$name` "Content-Type" and another `$name` "content-type", their values should probably be merged? Or only the values from one should be used, or we should throw an exception.

We're already considering case insensitive lookup for the get methods, but we need to also consider it for setting headers.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,383 @@
...
+     * @since n.e.x.t
+     *
+     * @param string $name The header name.
+     * @param string|list<string> $value The header value(s).
+     * @return self A new instance with the header.
+     */
+    public function withHeader(string $name, $value): self
+    {
+        $headers = $this->headers;
+        $headers[$name] = is_array($value) ? array_values($value) : [$value];
```
felixarntz: See my other comment: If the given `$name` was "content-type", but there was already a `$this->headers['Content-Type']`, we probably shouldn't just set it.

felixarntz (commented):
One additional caveat here regarding header handling.

Either we simply transform all header names always to lowercase, or we define some approach of merging while trying to maintain the original casing, e.g.:
* Keep header name in its original casing.
* For lookup of whether there's a conflict, we always check in a case-insensitive way (we already have the lowercase lookup map for that).
* If there's an attempt to set another header that case-insensitive would have the same name as an existing header, set its values under the new header name and remove the old header name.

It might not be a bad idea to have a single-purpose `Headers` class that handles all this case-insensitive while maintaining original case lookup and setting in a central place. Not a must have, just an idea.

src/Providers/Http/DTO/Request.php
JasonTheAdams: I like your idea of a unified way of handling all this, so I made a `HeadersCollection` class, which handles these fun scenarios. Resolved in 928101e

src/Providers/Http/DTO/Request.php
JasonTheAdams: I don't know. I think it makes sense to simply overwrite the value. It's an odd way of going about it, but it's not technically wrong.

Tentatively resolved in 928101e

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in 928101e

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in a632982

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in 720c294

src/Providers/Http/DTO/Request.php
felixarntz: 👍  Some follow up feedback, but I'll comment separately.

docs/ARCHITECTURE.md
```diff
@@ -982,14 +1142,6 @@ direction LR
...
     }
 
-    namespace AiClientNamespace.Providers.Models.EmbeddingGeneration.Contracts {
-        class EmbeddingGenerationModelInterface {
-            +generateEmbeddingsResult(Message[] $input) EmbeddingResult
-        }
-        class EmbeddingGenerationOperationModelInterface {
-            +generateEmbeddingsOperation(Message[] $input) EmbeddingOperation
-        }
-    }
```
felixarntz: Can you reinstate these pieces? I'm not sure why we should remove the embedding definition in the architecture. At the very least it's unrelated to what this PR does, so let's discuss separately if needed.

docs/ARCHITECTURE.md
```diff
@@ -787,6 +787,166 @@ direction LR
...
+    }
+    
+    <<interface>> WithHttpTransporterInterface
+    
+    SomeProviderTextGenerationModel ..|> TextGenerationModelInterface
+    SomeProviderTextGenerationModel ..|> WithHttpTransporterInterface
+    SomeProviderTextGenerationModel --> HttpTransporterInterface : uses
+    SomeProviderTextGenerationModel --> Request : creates
+    SomeProviderTextGenerationModel --> Response : receives
+```
```
felixarntz: There is a lot of very detailed documentation in here. Maybe that's helpful, but I'm a bit wary of this getting out of date because of how precise and verbose it is. Do we really want to keep all of this in here? If so, have you verified it's up to date with the latest changes you've been making to the actual code in this PR?

JasonTheAdams: Oops! I don't know why it did that. Good catch! Resolved in 577327a

docs/ARCHITECTURE.md
JasonTheAdams: Resolved in 577327a

docs/ARCHITECTURE.md
felixarntz: @JasonTheAdams This comment was not about the embeddings :)

src/Providers/Http/Collections/HeadersCollection.php
```diff
@@ -0,0 +1,144 @@
...
+
+        // If header exists with different casing, use the existing casing
+        if (isset($this->headersMap[$lowerName])) {
+            $actualName = $this->headersMap[$lowerName];
+            $this->headers[$actualName] = $normalizedValues;
+        } else {
+            // New header, use provided casing
+            $this->headers[$name] = $normalizedValues;
+            $this->headersMap[$lowerName] = $name;
+        }
```
felixarntz: I would argue the opposite makes more sense: "If header exists with different casing, use the new casing"

The same way we set the new value here (obviously), I think we should also set the new name.
```suggestion
        // If header exists with different casing, replace it with provided casing.
        if (isset($this->headersMap[$lowerName])) {
            $actualName = $this->headersMap[$lowerName];
            if ($actualName !== $name) {
                unset($this->headers[$actualName]);
            }
        }

        $this->headers[$name] = $normalizedValues;
        $this->headersMap[$lowerName] = $name;
```

src/Providers/Http/Collections/HeadersCollection.php
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Simple collection for managing HTTP headers with case-insensitive access.
+ *
+ * This class stores HTTP headers while preserving their original casing
+ * and provides efficient case-insensitive lookups.
+ *
+ * @since n.e.x.t
+ */
+class HeadersCollection
```
felixarntz, line 15: Not strictly necessary, but this class IMO is a good candidate to implement things like `ArrayAccess`, `IteratorAggregate`, `Countable`. That way you can work with it as if it was an array.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+ * }
+ *
+ * @extends AbstractDataTransferObject<RequestArrayShape>
+ */
+class Request extends AbstractDataTransferObject
+{
+    public const KEY_METHOD = 'method';
+    public const KEY_URI = 'uri';
+    public const KEY_HEADERS = 'headers';
+    public const KEY_DATA = 'data';
```
felixarntz: I'm thinking we may want to still treat `KEY_BODY` as the source of truth maybe for the array/JSON representation? The `$data` thing is just a helper to make setting and getting data easier.

At the end of the day, `$data` either becomes part of `KEY_URI` or `KEY_BODY`.

Alternatively, we also have both in the array/JSON representation: If `$data` is a string, we'll include `KEY_BODY`, otherwise we'll include `KEY_DATA`. But I feel like this confuses the simplifications of this class too much with what's the actual information in it. Method, URI, headers, and body is really all that's foundational to a request, so I'd vote to use those 4 in the array/JSON representation.

WDYT?

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Gets the request body string.
+     *
+     * @since n.e.x.t
+     *
+     * @return string|null The request body string.
+     */
+    public function getBodyString(): ?string
```
felixarntz: Why not simply `getBody()`? It's always going to be a string (unless it's `null`).

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+                    'additionalProperties' => [
+                        'type' => 'array',
+                        'items' => ['type' => 'string'],
+                    ],
+                    'description' => 'The request headers.',
+                ],
+                self::KEY_DATA => [
+                    'type' => ['string', 'array', 'null'],
+                    'description' => 'The request data.',
+                ],
```
felixarntz: In JSON schema, this can also be `object` (associative array in PHP). In fact, many times that'll be the case.

But per my above comment, I'm thinking it would be better to only expose `KEY_BODY` for the array/JSON representation. The `$data` piece is a simplification, but if we expose it like this, it means the outside code would need to know what to do with it. On the other hand, if we expose `KEY_BODY`, that's immediately obvious - it's simply the request body.

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+            self::KEY_URI => $this->uri,
+            self::KEY_HEADERS => $this->headers->getAll(),
+        ];
+
+        // Include whichever data type is set
+        if ($this->body !== null) {
+            $array[self::KEY_DATA] = $this->body;
+        } elseif ($this->data !== null) {
+            $array[self::KEY_DATA] = $this->data;
+        }
```
felixarntz: See my comment above: I think it would be great to expose this with the `$data` part internally handled, i.e. either baked into `KEY_URI` or `KEY_BODY`.

Something like this:
```suggestion
        $array = [
            self::KEY_METHOD => $this->method->value,
            self::KEY_URI => $this->getUri(),
            self::KEY_HEADERS => $this->headers->getAll(),
        ];

        $body = $this->getBody();
        if ($body !== null) {
            $array[self::KEY_BODY] = $body;
        }
```

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+     */
+    public static function fromArray(array $array): self
+    {
+        static::validateFromArrayData($array, [self::KEY_METHOD, self::KEY_URI, self::KEY_HEADERS]);
+
+        return new self(
+            HttpMethodEnum::from($array[self::KEY_METHOD]),
+            $array[self::KEY_URI],
+            $array[self::KEY_HEADERS] ?? [],
+            $array[self::KEY_DATA] ?? null
```
felixarntz: Per the above suggestions, this would become:
```suggestion
            $array[self::KEY_BODY] ?? null
```

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,371 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @return string|null The body.
+     * @throws JsonException If the data cannot be encoded to JSON.
+     */
+    public function getBody(): ?string
+    {
+        // GET requests don't have a body
+        if ($this->method === HttpMethodEnum::GET()) {
```
felixarntz: Why not use the `hasBody()` method this PR implements? Seems to have exactly this purpose.
```suggestion
        if (!$this->method->hasBody()) {
```

src/Providers/Http/Enums/HttpMethodEnum.php
```diff
@@ -0,0 +1,138 @@
...
+    public const TRACE = 'TRACE';
+
+    /**
+     * Checks if this method is safe (doesn't modify server state).
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if the method is safe, false otherwise.
+     */
+    public function isSafe(): bool
```
felixarntz: Why this method? When is it relevant to call? In any case, I find the name a bit weird, as it makes it sound like calling write methods are inherently unsafe.

src/Providers/Http/Enums/HttpMethodEnum.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Checks if this method is idempotent.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if the method is idempotent, false otherwise.
+     */
+    public function isIdempotent(): bool
```
felixarntz, line 106: What does "idempotent" mean? Also, what would this be used for?

src/Providers/Http/HttpTransporter.php
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param ResponseInterface $psr7Response The PSR-7 response.
+     * @return Response The custom response.
+     */
+    private function convertFromPsr7Response(ResponseInterface $psr7Response): Response
+    {
+        /** @var array<string, string|list<string>> $headers */
+        $headers = [];
+        foreach ($psr7Response->getHeaders() as $name => $values) {
+            $headers[(string) $name] = count($values) === 1 ? $values[0] : array_values($values);
```
felixarntz: Why not just always pass the array? 🤔

felixarntz (commented):
@JasonTheAdams This looks really close. A few last points of feedback, most importantly I think we should treat the body (string) as source of truth for the `Request`. The `$data` constructor parameter and class property are fully interoperable with it and purely provide an easy layer to set and get raw data.

src/Providers/Http/Collections/HeadersCollection.php
JasonTheAdams: It originally did work that way and I pruned it back to reduce complexity. After seeing it all in there and how it's actually used I realized it was just a lot of code. But I can add it back if you think it's not too much.

src/Providers/Http/DTO/Request.php
JasonTheAdams: Both certainly work. This is closer to how enum comparison actually works, so I've left it if the AI happened to write it this way.

src/Providers/Http/Enums/HttpMethodEnum.php
JasonTheAdams: It's just a helper function. Idempotency means that calling a function multiple times will not have any side-effects each time. A PUT request, for example can be called multiple times in a row with the same values and all it will do is set the resource values.

src/Providers/Http/DTO/Request.php
felixarntz: How do you mean "closer to how enum comparison actually works"? Why not use the method that exists to check whether the method supports a body? To me that's precisely what this clause is for.

src/Providers/Http/Collections/HeadersCollection.php
felixarntz: I don't think it is. IMO makes it easier to work with. But either way is fine, especially since it's only used internally.

src/Providers/Http/Enums/HttpMethodEnum.php
JasonTheAdams: Yeah, you're right about this one. I don't see much value in it. Resolved in [f46fcde](https://github.com/WordPress/php-ai-client/pull/47/commits/f46fcdef0b72906c131be89d7c2d19d497309f89)

src/Providers/Http/DTO/Request.php
JasonTheAdams: I totally misread this. Hahah! Resolved in 9f3a3ce

src/Providers/Http/DTO/Request.php
JasonTheAdams: There is a `getBody()` method and it has conditions based on the method and such. This was to directly get the body property, but since it's not used anywhere I'm removing it for now.

Resolved in a733263

src/Providers/Http/HttpTransporter.php
JasonTheAdams: Good catch. This was a leftover from before I added the `HeadersCollection`.

Resolved in df49b6f

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in adfcd5f

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in adfcd5f

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in 1f7b7bd

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in adfcd5f

src/Providers/Http/Collections/HeadersCollection.php
JasonTheAdams: Yeah, I agree! Resolved in 06f2da1

docs/ARCHITECTURE.md
JasonTheAdams: My bad. I mixed it up with the other comment.

Honestly, I'm not married to the extra docs at all. I wasn't sure if having it would help Extenders better understand how to use the HttpTransporter and DTOs. But you're right that it's more to maintain, and I'm not convinced in its value. I'm happy to remove this if you think it's best.

felixarntz: I would say the initial sections, including the Mermaid sequence diagram, are definitely worth keeping. IMO mainly the additional very detailed class diagrams could be removed, because:
* they would mean more substantial additional maintenance keeping it up to date
* they have some overlap with the detailed class diagrams above in the file that are already in place

src/Providers/Http/DTO/Request.php
```diff
@@ -0,0 +1,359 @@
...
+                    ],
+                    'description' => 'The request headers.',
+                ],
+                self::KEY_BODY => [
+                    'type' => ['string', 'array', 'object', 'null'],
+                    'description' => 'The request body. Always a string when serialized. ' .
+                        'Can be a string, array, or object when deserializing.',
+                ],
+            ],
+            'required' => [self::KEY_METHOD, self::KEY_URI, self::KEY_HEADERS],
```
felixarntz: Shouldn't this be just `string`?
* `array` and `object` are only relevant when instantiating the class in PHP, but not for the array / JSON representation.
* If this could be `array` or `object`, it would make handling the DTO elsewhere (e.g. in JS) unnecessarily complicated, as the logic would need to figure out what to do with the data, which would be duplicating something that the PHP class already does.
* `KEY_BODY` is not marked as `required`, so including `null` explicitly isn't needed.

Basically, I think `KEY_BODY` should always be a string, and the `$data` thing we can keep PHP-only. You can pass a PHP array to the constructor, and you can even pass a PHP array to the `fromArray` method, but when consuming the array shape, you consistently get the `body` string.

This is already the case based on the `toArray` implementation, but it's not clear from this schema.

docs/ARCHITECTURE.md
felixarntz: Per https://github.com/WordPress/php-ai-client/pull/47#discussion_r2278021057, I'd say let's remove this portion to avoid too much detail for us to maintain:
```suggestion
```

felixarntz (approved):
@JasonTheAdams Two last follow up comments that I think should be addressed. But they should be quick.

I'm going to tentatively approve, and feel free to merge without another review after you've addressed them. If you have any concerns, let me know. Awesome work, I'm really happy to see this final piece of the foundation land. 🍻

src/Providers/Http/DTO/Request.php
JasonTheAdams: Ah, I misunderstood a previous comment, then. Thanks for breaking your thoughts down!

Resolved in 1c74d37

PR #49, opened 2025-08-14
Add the Prompt Builder

JasonTheAdams:
This introduces the Prompt Builder, which is intended to be the primary face for Implementors to construct and execute prompts. See the ["Fluent API" examples](https://github.com/WordPress/php-ai-client/blob/trunk/docs/ARCHITECTURE.md#code-examples) in the code examples in the Architecture doc.


JasonTheAdams:
Ok, @felixarntz! I have a whopping PR for you to look over! Hahah! 2600 lines of this are the unit tests for the builder, so that helps. But it's still quite a bit. Happy to talk through any part of this!

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -913,6 +944,131 @@ static function (FunctionDeclaration $function_declaration): array {
...
+     * @return list<RequiredOption> The required options.
+     */
+    public function toRequiredOptions(): array
+    {
+        $requiredOptions = [];
+
+        // Map properties that have corresponding OptionEnum values
+        if ($this->outputModalities !== null) {
+            $requiredOptions[] = new RequiredOption(
+                OptionEnum::outputModalities()->value,
```
felixarntz: I was wondering about this the other day - do we really need `OptionEnum`? Maybe it's a good idea to decouple `ModelConfig` from these "options", but almost every option directly corresponds to a `ModelConfig` parameter. So maybe we can get rid of `OptionEnum` and instead rely on the `ModelConfig::KEY_*` constants?

Alternatively, if we decide `OptionEnum` continues to be useful, then shouldn't we update `RequiredOption` and `SupportedOption` to accept an `OptionEnum` instead of just a `string` for the name?

Curious what's your perspective on this.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -913,6 +944,131 @@ static function (FunctionDeclaration $function_declaration): array {
...
+            $requiredOptions[] = new RequiredOption(
+                OptionEnum::outputMimeType()->value,
+                $this->outputMimeType
+            );
+        }
+
+        if ($this->outputSchema !== null) {
+            $requiredOptions[] = new RequiredOption(
+                OptionEnum::outputSchema()->value,
+                true // Just indicate that schema is required
```
felixarntz: Why? I think we can just use `$this->outputSchema` here, just like elsewhere.
```suggestion
                $this->outputSchema
```

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -913,6 +944,131 @@ static function (FunctionDeclaration $function_declaration): array {
...
+            );
+        }
+
+        // Handle properties without OptionEnum values as custom options
+        // These would need to be handled specially by providers
```
felixarntz, line 993: Somewhat related to my above point, this is not right.

Either these need to be added to `OptionEnum`, or we get rid of that entirely and use the `ModelConfig::KEY_*` constants.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -202,6 +203,36 @@ public function getOutputModalities(): ?array
...
+     *
+     * Adds the given modality to the output modalities list if it's not
+     * already included. If output modalities is null, initializes it with
+     * the given modality.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModalityEnum $modality The modality to include.
+     */
+    public function includeOutputModality(ModalityEnum $modality): void
```
felixarntz: To the point I raised the other day regarding consistency: If we add this, we'll need to think about other array parameters of `ModelConfig` too. I think consistency is important, especially in API design, so I don't think we should add public methods for new aspects or patterns that we haven't considered holistically.

Now, thinking about this holistically would probably take some thinking and discussion, so I'd prefer to avoid it at this point where we're close to our ideal MVP deadline.

But there's an easy solution that leaves all our options open in the long-term: Instead of this being a public method on `ModelConfig`, why not make it a private method in `PromptBuilder` that operates on the underlying `ModelConfig` instance? This way, we don't need to commit to anything, we can later decide to do exactly what you have here - or not. Point being, we don't code ourselves into a corner we can't get out of, and we don't need to make any decisions right now. Private methods are always fine without any major API design consideration.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+        if ($this->isMessagesList($prompt)) {
+            $this->messages = $prompt;
+            return;
+        }
+
+        // Check if it's a MessageArrayShape - add to messages
+        if (is_array($prompt) && Message::isArrayShape($prompt)) {
+            $this->messages[] = Message::fromArray($prompt);
+            return;
+        }
```
felixarntz: Since `parseMessage` already accepts pretty much anything that can be parsed into a message, why not make this clause part of it too? It doesn't make sense to me that it's separate.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     * Parses various input types into a Message with the given role.
+     *
+     * @since n.e.x.t
+     *
+     * @param mixed $input The input to parse.
+     * @param MessageRoleEnum $role The role for the message.
+     * @return Message The parsed message.
+     * @throws InvalidArgumentException If the input type is not supported or results in empty message.
+     */
+    private function parseMessage($input, MessageRoleEnum $role): Message
```
felixarntz: Nit-pick: Maybe call the second parameter `$defaultRole` or `$fallbackRole`? It's only used if the input doesn't already specify a role.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     * @param string $mimeType The MIME type of the image.
+     * @return self
+     */
+    public function withRemoteImage(string $uri, string $mimeType): self
+    {
+        $file = new File($uri, $mimeType);
+        $part = new MessagePart($file);
+        $this->appendPartToMessages($part);
+        return $this;
+    }
```
felixarntz: I'm trying to think why we wouldn't want the `$mimeType` parameter on these to be similarly handled as in `File`. Why is it not optional here? And also, shouldn't `withInlineImage` also allow _either_ just the base64 blob (and mandatory MIME type) _or_ a data URI (and optional MIME type)? That would be similar to how `File` works.

My other question about this is: If we have these two methods for image, why don't we have them for other things like audio, video, etc? There's also nothing in this method that actually implies it's an image. You could call `withInlineImage` passing an audio file, and there would be no complaint about it. Maybe the focus on image in the naming is just not the right one? Maybe a simple `withInlineFile` and `withRemoteFile` would be more clear.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     *
+     * @param File $file The video file.
+     * @return self
+     */
+    public function withVideoFile(File $file): self
+    {
+        $part = new MessagePart($file);
+        $this->appendPartToMessages($part);
+        return $this;
+    }
```
felixarntz: Related to the above: The code of all these is the same. Shouldn't we consolidate them to a simple `withFile`? There's nothing in the method logic that would actually enforce the type of the file, and it would be better aligned with how `MessagePart` works.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+
+    /**
+     * Adds a function response to the current message.
+     *
+     * @since n.e.x.t
+     *
+     * @param FunctionResponse $functionResponse The function response.
+     * @return self
+     */
+    public function withFunctionResponse(FunctionResponse $functionResponse): self
```
felixarntz, line 145: I was wondering why there is no `withFunctionCall`. I suppose it makes sense because a user message would not typically trigger a function call. But if we want to enforce it, shouldn't we also ensure that in `Message` itself?

Similarly, a model `Message` should never include a function response.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     * @since n.e.x.t
+     *
+     * @param Message ...$messages The messages to add to history.
+     * @return self
+     */
+    public function withHistory(Message ...$messages): self
+    {
+        foreach ($messages as $message) {
+            $this->messages[] = $message;
+        }
```
felixarntz: This doesn't look right to me. I'd argue, since this is a history, it should by definition be inserted before whatever the current message is. If you call `withText(...)->withHistory(...)`, the `withText` should still be for the latest message, not for the first message before all other history. That wouldn't make sense to me.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     * @since n.e.x.t
+     *
+     * @param ProviderRegistry $registry The provider registry to use.
+     * @return self
+     */
+    public function usingRegistry(ProviderRegistry $registry): self
+    {
+        $this->registry = $registry;
+        return $this;
+    }
```
felixarntz: Just in the interest of not allowing something too advanced before we have thought this through, can we remove this for now and reconsider later? It's not an important thing to allow for the MVP anyway.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+                    } elseif ($file->isDocument() || $file->isText()) {
+                        $inputModalities[ModalityEnum::document()->value] = ModalityEnum::document();
+                    }
+                }
+
+                // Check for function calls/responses (these might require special capabilities)
+                if ($part->getFunctionCall() !== null || $part->getFunctionResponse() !== null) {
+                    // Function calling capability would go here if we had it in CapabilityEnum
+                    // For now, we'll just note this requires text generation
+                }
```
felixarntz: I think we should look for the `functionDeclarations` option in that case. I realize this is still a special case, but no need to add more capabilities. A model that can accept `functionDeclarations` can also process function call and function response messages. Basically the `functionDeclarations` option is the same as saying "function calling".

So this would be something to merge with the `$requiredOptions` below.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+        // Build required options from ModelConfig
+        $requiredOptions = $this->modelConfig->toRequiredOptions();
+
+        // Add input modalities if we have non-text inputs
+        if (count($inputModalities) > 0) {
+            $requiredOptions[] = new RequiredOption(
+                OptionEnum::inputModalities()->value,
+                array_values($inputModalities)
+            );
+        }
```
felixarntz: This logic seems wrong based on the comment above? Further above, text inputs are already added to `$inputModalities`. So checking `count($inputModalities) > 0` will basically always be true I think.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+
+    /**
+     * Checks if the current prompt is supported by the selected model.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModalityEnum|null $intendedOutput Optional output modality to check support for.
+     * @return bool True if supported, false otherwise.
+     */
+    public function isSupported(?ModalityEnum $intendedOutput = null): bool
```
felixarntz: Can we make this private or protected? Seems like a lower-level method. Again, I wouldn't say this _definitely_ shouldn't be public, but it's always better to start private/protected if we're not sure. Once it's public, there's no turning back (or at least, it'll be _a lot_ more difficult).

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1275 @@
...
+     * Checks if the prompt is supported for speech generation.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if speech generation is supported.
+     */
+    public function isSupportedForSpeech(): bool
+    {
+        return $this->isSupported(ModalityEnum::audio());
+    }
```
felixarntz: Can we rename these methods to clarify their purpose is about the specific generation capability rather than a modality? The descriptions already say e.g. "text generation", so the method should also be called e.g. `isSupportedForTextGeneration` IMO.

I'm also not sure why this is relying on modalities. While there's quite some overlap, the capabilities are the primary aspect to look for here I think. For things like `text` output modality vs `text generation` capability, this may not be very obvious, but take the example of `audio` output modality vs `speech generation` _or_ `text to speech conversion` capability. The latter is what matters, and it carries more meaning. A model may support audio output, but might only support one of the two capabilities for example.

Trying to reason about this further: "text generation" capability support always implies support for "text" output modality - anything else wouldn't make sense. Since all the necessary requirements other than the main capability are dictated by the `list<Message>` and the `ModelConfig`, the only thing missing (which I think these methods are supposed to cover) is the _capability_, e.g. "text generation" etc.

felixarntz (changes requested):
@JasonTheAdams This looks like a great start, but some things I think need to be revised.

For context, so far I've reviewed the `PromptBuilder` class up from the top up until before `generateResult`, plus all other class changes.

For my `PromptBuilder` feedback, I think mostly the one about modality vs capability is most significant. Without having reviewed the `generate*` methods yet, there's a good chance, that conversation will also influence the code for those to some degree.

Regarding all other class changes outside of `PromptBuilder`, those already almost LGTM.

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: I actually saw that and had a similar thought. I'd prefer it to accept the enums, otherwise we're basically treating the keys _as_ an enum.

Are the `ModelConfig` keys always going to be a 1:1 with the `OptionEnum`?

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Resolved in d976ccf

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Right, good note. I think this all comes down to my question above regarding whether `OptionEnum` is a 1:1 withthe `ModelConfig` keys.

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Resolved in bba08dd

src/Builders/PromptBuilder.php
JasonTheAdams: Good catch! Resolved in 9281c54

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 9281c54

src/Builders/PromptBuilder.php
JasonTheAdams: Yeah, now that you saw that I don't think there's any reason to have separate methods for the types and such, especially since `getModelRequirements()` just checks the type of file to determine what's needed. Cleans up the API to just reduce this to a couple methods for all types of files. I'll also address the  `$mimeType` parameter.

JasonTheAdams: So I ended up with a single `withFile` method. It's simple, and takes both inline and remote files — because `File` simply takes both and handles it well. I also added passthrough `File::isRemote()` and `File::isInline()` method to make life simple.

Resolve in d7320ce and cb54e0b

Ref34t:
@JasonTheAdams Thank you for your great work here. You accomplished the intended goal of this PR. I've some thoughts that may have some improvements in return will share over the files. Well Done 💯

src/Builders/PromptBuilder.php
Ref34t: Exactly right! I completely agree with your comment - history should definitely be inserted before the current message, not after.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1261 @@
...
+        if ($this->model !== null) {
+            $this->model->setConfig($this->modelConfig);
+            return $this->model;
+        }
+
+        // Find a suitable model based on requirements
+        $modelsMetadata = $this->registry->findModelsMetadataForSupport($requirements);
+
+        if (empty($modelsMetadata)) {
+            throw new InvalidArgumentException(
```
Ref34t, line 936: The current error messages could be more actionable:
Instead of generic "No models found..."
` throw new InvalidArgumentException(
    'No models found that support the required capabilities. ' .
    'Consider: 1) Checking provider configuration, ' .
    '2) Using a different model, or 3) Simplifying prompt requirements. ' .
    'Required: ' . $this->formatRequirements($requirements)
);`
Can be more helpful for debugging in real-world usage.

src/Providers/Models/DTO/ModelConfig.php
Ref34t: Great discussion here! I lean toward eliminating `OptionEnum` and using `ModelConfig::KEY_*` constants instead, for several reasons:

The 1:1 mapping creates unnecessary abstraction
Tight coupling between ModelConfig and OptionEnum violates separation of concerns More complex to maintain two parallel hierarchies. The enum doesn't add meaningful type safety since `RequiredOption/SupportedOption` still accept strings

**Suggested approach:**

Instead of:
`new RequiredOption(OptionEnum::outputModalities()->value, $this->outputModalities)`

 Use:
`new RequiredOption(ModelConfig::KEY_OUTPUT_MODALITIES, $this->outputModalities)`
This is simpler, more direct, and follows the existing pattern you've established with other ModelConfig constants.

If keeping `OptionEnum:` Then definitely update `RequiredOption/SupportedOption` constructors to accept `OptionEnum` directly:
`phppublic function __construct(OptionEnum $name, mixed $value)`
But given the 1:1 mapping, the constants approach seems cleaner and more maintainable long-term.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1261 @@
...
+    /**
+     * Generates multiple text candidates from the prompt.
+     *
+     * @since n.e.x.t
+     *
+     * @param int|null $candidateCount The number of candidates to generate.
+     * @return list<string> The generated texts.
+     * @throws InvalidArgumentException If the prompt or model validation fails.
+     */
+    public function generateTexts(?int $candidateCount = null): array
```
Ref34t, line 775: Consider adding more specific return type hints:
**Current**
`public function generateTexts(?int $candidateCount = null): array`

 **Suggested**
`public function generateTexts(?int $candidateCount = null): list<string>`
This provides better IDE support and static analysis.

src/Providers/Models/DTO/ModelConfig.php
felixarntz: Two thoughts here:
* To @Ref34t's point, I agree we wouldn't want to duplicate the actual string identifiers that already exist in `ModelConfig`.
* To @JasonTheAdams's point, there are a few `OptionEnum` values that we need that _won't_ be part of `ModelConfig` (e.g. `inputModalities`). But the majority would be a 1:1 match - basically `ModelConfig` keys are a subset of `OptionEnum` values.

Based on this, I think we should do the following:
* Keep `OptionEnum`, but for all its enum values that are directly referencing keys from `ModelConfig` we should have the enum values directly reference those `KEY_*` constants, instead of duplicating their values (and thus making the connection less obvious).
* Include any additional options that don't relate to `ModelConfig` in `OptionEnum` directly.
* Adjust `RequiredOption` and `SupportedOption` to require an `OptionEnum` instead of `string` for the option identifier.

Let me know if you agree or have any concerns.

src/Providers/Models/DTO/ModelConfig.php
felixarntz: Replied to that thread. 👍

src/Builders/PromptBuilder.php
felixarntz: This doesn't work? `list<string>` is not valid PHP.

Ref34t: @felixarntz You're absolutely right - my mistake! `list<string>` is PHPStan annotation syntax, not valid PHP.

 The correct approach is:
- Keep PHP method signature as: `public function generateTexts(?int $candidateCount = null): array`
- Use `@return list<string>` only in the PHPDoc comment for static analysis

 Thanks for the correction - I mixed up PHPStan annotations with actual PHP type hints!

src/Providers/Models/DTO/ModelConfig.php
Ref34t: @felixarntz That approach sounds solid! The hybrid solution addresses both concerns well:

- Eliminates duplication by having OptionEnum reference ModelConfig::KEY_* constants directly
- Maintains type safety with OptionEnum parameter in constructors
- Keeps clean separation for non-ModelConfig options like `inputModalities`

src/Builders/PromptBuilder.php
JasonTheAdams: Sure! Resolved in f324b96

src/Builders/PromptBuilder.php
JasonTheAdams: I like that! I added parts validation in `Message`. Resolved in a07c242

JasonTheAdams: I like that! I added parts validation in `Message`. Resolved in a07c242

src/Builders/PromptBuilder.php
JasonTheAdams: You mean keep the other `isSupported...` methods, but make this one private? I figured it would be worth keeping this public for multi-modal output. Though I'm realizing now it would need to accept more than one modality.

What do you think, @felixarntz?

src/Builders/PromptBuilder.php
JasonTheAdams: Good catch! Resolved in 0d39e74

src/Builders/PromptBuilder.php
JasonTheAdams: For what it's worth, I wish there was a PHP list type. 😄

src/Builders/PromptBuilder.php
felixarntz: IMO the lower-level method should be private to start with - again, for the same reason I generally advise for private unless we know we need it public.

I think more important than whether this is private or public is my earlier feedback in https://github.com/WordPress/php-ai-client/pull/49#discussion_r2295105853 on this having to consider _capabilities_ rather than just modalities.

We'll need `isSupportedForTextGeneration`, `isSupportedForImageGeneration` etc., and each of these needs to check support based on:
* the `list<Message>`
* the `ModelConfig`, _including_ the necessary output modality for the respective capability (which is basically what you're doing now)
* the capability, e.g. text generation, image generation, i.e. one of the `CapabilityEnum` values (this is what's missing from the current implementation)

src/Common/AbstractEnum.php
```diff
@@ -253,7 +253,7 @@ private static function getInstance(string $value, string $name): self
      * @return array<string, string> Map of constant names to values.
      * @throws RuntimeException If invalid constant found.
      */
-    final protected static function getConstants(): array
+    protected static function getConstants(): array
```
felixarntz: @JasonTheAdams I see you needed to make the `$cache` property protected, and that now is handled in the child class as well. Can we keep this self-contained? That should make this implementation safer and simpler. For example:
* Keep this method final, and the `$cache` property private.
* Move the actual logic to determine the constant names and values into another method that should be protected but not final. Could be called something like `determineConstantNamesAndValues()`, or `determineConstants()`.
* _That_ method you can then override in the child class without needing to worry about the cache.

src/Providers/Models/DTO/ModelConfig.php
felixarntz: Fixed via https://github.com/WordPress/php-ai-client/pull/49/commits/7d25a0389c9634aab1826da23d72f3c25a638ad5

felixarntz (commented):
@JasonTheAdams This is very close. No other feedback on my end than what's left from the previous reviews.

src/Common/AbstractEnum.php
JasonTheAdams: Good idea! Resolved in 3396da6

src/Builders/PromptBuilder.php
JasonTheAdams: Agreed! Resolved in 5d081b4

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 5d081b4

src/Builders/PromptBuilder.php
JasonTheAdams: Upon further inspection this whole method's logic was weird. Resolved in cb7fb19

src/Builders/PromptBuilder.php
JasonTheAdams: We're actually already doing this in `ModelConfig::toRequiredOptions()` — does that mean we're good here? Or should we still check this?

JasonTheAdams: We chatted in person and decided it's best to check here as well. Resolved in 2c7b443

src/Builders/PromptBuilder.php
JasonTheAdams: This likely isn't a simple solution for the implementor, as they should be using the `isSupportedForXGeneration` methods as the way of catching this. As such, I'm reluctant to give instructions when we'd be guessing at this point as to what they should do.

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+
+    /**
+     * Configures the prompt for JSON response output.
+     *
+     * @since n.e.x.t
+     *
+     * @param array<string, mixed>|null $schema Optional JSON schema.
+     * @return self
+     */
+    public function asJsonResponse(?array $schema = null): self
```
felixarntz, line 352: Maybe `usingJsonOutput` for a bit more naming parity with other related names?

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+        // Always need text generation capability
+        $capabilities[] = CapabilityEnum::textGeneration();
+
+        // Check if we have chat history (multiple messages)
+        if (count($this->messages) > 1) {
+            $capabilities[] = CapabilityEnum::chatHistory();
+        }
+
+        // Analyze all messages to determine required input modalities
+        $hasFunctionCallingCapability = false;
```
felixarntz: Nit-pick: I'd say either call this `$requiresFunctionCalling` or `$hasFunctionMessageParts`?

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+
+        // Get the configured output modalities
+        $outputModalities = $this->modelConfig->getOutputModalities();
+
+        // Default to text if no output modality is specified
+        if ($outputModalities === null || empty($outputModalities)) {
+            $outputModalities = [ModalityEnum::text()];
+        }
+
+        // Multi-modal output (multiple modalities) uses TextGenerationModelInterface
```
felixarntz: Maybe clarify that eventually we want to have a more multimodal dedicated interface?
```suggestion
        /*
         * For now, multi-modal output (multiple modalities) uses TextGenerationModelInterface.
         * In the future, there will be an additional multimodal generation model interface that extenders can use.
         */
```

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+            $this->usingCandidateCount($candidateCount);
+        }
+
+        // Generate text result
+        $results = $this->generateTextResult();
+        $candidates = $results->getCandidates();
+
+        // Extract text from each candidate
+        $texts = [];
+        foreach ($candidates as $candidate) {
```
felixarntz: Don't we have `toTexts()` for this purpose?

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+        $message = $candidates[0]->getMessage();
+        $parts = $message->getParts();
+        if (empty($parts)) {
+            throw new RuntimeException('Generated message contains no parts.');
+        }
+
+        $text = $parts[0]->getText();
+        if ($text === null) {
+            throw new RuntimeException('Generated message part contains no text.');
+        }
```
felixarntz: Don't we have `toText()` for this?

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+        $message = $candidates[0]->getMessage();
+        $parts = $message->getParts();
+        if (empty($parts)) {
+            throw new RuntimeException('Generated message contains no parts.');
+        }
+
+        $file = $parts[0]->getFile();
+        if ($file === null) {
+            throw new RuntimeException('Generated message part contains no image file.');
+        }
```
felixarntz: See above, this should use `toFile()`.

Also, similar feedback for the below methods (not going to comment on these individually).

src/Builders/PromptBuilder.php
```diff
@@ -0,0 +1,1327 @@
...
+     * Generates multiple speech outputs from the prompt.
+     *
+     * @since n.e.x.t
+     *
+     * @param int|null $candidateCount The number of speech outputs to generate.
+     * @return list<File> The generated speech audio files.
+     * @throws InvalidArgumentException If the prompt or model validation fails.
+     * @throws RuntimeException If no audio is generated.
+     */
+    public function generateSpeeches(?int $candidateCount = null): array
```
felixarntz, line 875: I don't love this name - to me this sounds like multiple speeches as in "delivering a speech". But maybe it's what we need to do because the term "speech" like it's meant here doesn't really have a plural?

src/Builders/PromptBuilder.php
JasonTheAdams: I think of the "using" prefix as describing _how_ it's going to do the work (e.g. `usingModel`), whereas `as-` influences _what_ it's going to return.

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 4f4779c

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in f366cdd

src/Builders/PromptBuilder.php
JasonTheAdams: Oh my goodness! Such a good catch! Resolved in 5d9a2a9

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 5d9a2a9

src/Builders/PromptBuilder.php
JasonTheAdams: Maybe `generateSpeechFile` and `generateSpeechFiles`?

src/Builders/PromptBuilder.php
JasonTheAdams: Per our discussion I resolved it via e7ffc3c for consistency

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in Resolved in 5d9a2a9

src/Builders/PromptBuilder.php
felixarntz: Let's consistently use `as` for any output related configuration then.

src/Builders/PromptBuilder.php
JasonTheAdams: @felixarntz decided to we'll keep it, but preserve this gripe for posterity. 😄

felixarntz (approved):
@JasonTheAdams LGTM! 🎉

PR #52, opened 2025-08-14
Allow message parts to have channel, to separate regular content from thought / reasoning content

felixarntz:
As the title says, it's crucial to make a distinction (mostly for model generated content), whether it's regular output or thinking / reasoning. This PR adds that distinction to message parts within a single message.

This is in line with how API providers typically handle it (a single message may contain both kinds of parts/content).

### DIFF
--- a/src/Messages/DTO/MessagePart.php
+++ b/src/Messages/DTO/MessagePart.php
@@ -298,11 +298,6 @@ class MessagePart extends AbstractDataTransferObject
     public static function fromArray(array $array): self
     {
         if (isset($array[self::KEY_CHANNEL])) {
-            if (!MessagePartChannelEnum::isValidValue($array[self::KEY_CHANNEL])) {
-                throw new InvalidArgumentException(
-                    sprintf('Invalid channel value: %s', $array[self::KEY_CHANNEL])
-                );
-            }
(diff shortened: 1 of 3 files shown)


src/Messages/DTO/MessagePart.php
```diff
@@ -260,15 +297,26 @@ public function toArray(): array
      */
     public static function fromArray(array $array): self
     {
+        if (isset($array[self::KEY_CHANNEL])) {
+            if (!MessagePartChannelEnum::isValidValue($array[self::KEY_CHANNEL])) {
+                throw new InvalidArgumentException(
+                    sprintf('Invalid channel value: %s', $array[self::KEY_CHANNEL])
+                );
+            }
+            $channel = MessagePartChannelEnum::from($array[self::KEY_CHANNEL]);
```
JasonTheAdams: I don't think you need to throw an exception here, as the `from` method will throw an exception if invalid. Unless you're wanting the custom exception message?

JasonTheAdams (approved):
Interesting! I was actually just wondering how thinking is identified. There we go!

One really minor suggestion that I'm not sure of your intent on, so I'll approve with handling it at your discretion.

src/Messages/DTO/MessagePart.php
felixarntz: Great catch, fixed!

PR #55, opened 2025-08-15
Add the AiClient with testing suite

Ref34t:
This introduces the AiClient, which is intended to be the primary face for Implementors to construct and execute prompts. See the "Traditional API" examples in the code examples in the Architecture doc.

This provides the foundation that the PromptBuilder (PR https://github.com/WordPress/php-ai-client/pull/49 ) depends on to be complete.


Ref34t:
@felixarntz could you please add the [Feature] label and 'Complete Implementor API' milestone to this PR? This continues the AiClient  work from PR #53, which you had previously labeled and milestone-tagged, but I decided to close it and create this separate one as the other one had commits from the other feature

Ref34t:
@felixarntz @JasonTheAdams after going back and forth a couple of times over this feature, I can say that I'm satisfied with it right now. Looking forward for your thoughts and improvements

JasonTheAdams (changes requested):
Thanks, @Ref34t! I'll give this a more thorough look in the next day. In the meantime, can you please remove everything related to embeddings that got introduced in this PR? We're not introducing embeddings until a later time.

Ref34t:
@JasonTheAdams I've created another PR #56 to keep the embeddings for future reference and cleaned this PR from all embeddings related as you requested. I think now it's better because it was a lot before.

src/Contracts/AiClientInterface.php
JasonTheAdams, line 1: Doesn't look like this is used anywhere? I don't think we need an interface for `AiClient`, anyway.

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+     *
+     * @since n.e.x.t
+     *
+     * @param ProviderAvailabilityInterface $availability The provider availability instance to check.
+     * @return bool True if the provider is configured and available, false otherwise.
+     */
+    public static function isConfigured(ProviderAvailabilityInterface $availability): bool
+    {
+        return $availability->isConfigured();
+    }
```
JasonTheAdams, line 122: This is an architectural comment (likely for @felixarntz), but this API feels weird to me:
```php
AiClient::isConfigured(Google::availability());
```

If this literally just returns the method call, why not just have a `ProviderInterface::isConfigured()` that returns a boolean and does it's own thing under the hood? The whole `ProviderAvailabilityInterface` thing feels like an unnecessary abstraction, unless it's simplifying something I'm not seeing?

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+     * @param string|MessagePart|MessagePart[]|Message|Message[] $prompt The prompt content.
+     * @param ModelInterface $model The model to use for generation.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid or model type is unsupported.
+     */
+    public static function generateResult($prompt, ModelInterface $model): GenerativeAiResult
+    {
+        // Use strategy resolver to determine the appropriate method
+        $method = GenerationStrategyResolver::resolve($model);
```
JasonTheAdams: This feels over-engineered to me. We have a separate resolver class returning a method string (which is not great for type safety), used in only a single place. 😅 

I think this could be a simple `switch...case` or `if...elseif` that just calls the `AiClient` method directly, preserving type safety and keeping things clear and in one place.

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+        // Get model - either provided or auto-discovered
+        $resolvedModel = $model ?? ModelDiscovery::findTextModel(self::defaultRegistry());
+
+        // Validate model supports text generation
+        InterfaceValidator::validateTextGeneration($resolvedModel);
+
+        // Stream the results using the model
+        /** @phpstan-ignore-next-line */
+        yield from $resolvedModel->streamGenerateTextResult($messageList);
+    }
```
JasonTheAdams: @felixarntz Since we're not starting with streaming support, should we just remove all streaming methods for now?

src/Utils/InterfaceValidator.php
```diff
@@ -0,0 +1,179 @@
...
+ */
+class InterfaceValidator
+{
+    /**
+     * Validates that a model implements TextGenerationModelInterface.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModelInterface $model The model to validate.
+     * @return void
```
JasonTheAdams: If you add the [phpstan-assert](https://phpstan.org/writing-php-code/narrowing-types#custom-type-checking-functions-and-methods) tag, then it will inform the type system that this method is validating the instance. This will make it so you don't have to ignore PhpStan errors downstream.
```suggestion
     * @phpstan-assert TextGenerationModelInterface $model	
     * @param ModelInterface $model The model to validate.
     * @return void
```

Apply something like this to the other methods as well. 😄

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+        $messages = PromptNormalizer::normalize($prompt);
+        /** @var list<Message> $messageList */
+        $messageList = array_values($messages);
+
+        // Map type to specific methods
+        switch ($type) {
+            case 'text':
+                $resolvedModel = $model ?? ModelDiscovery::findTextModel(self::defaultRegistry());
+                InterfaceValidator::validateTextGeneration($resolvedModel);
+                /** @phpstan-ignore-next-line */
```
JasonTheAdams: See my other comment for how to improve our type-checking so we don't have to ignore. Same for below!

src/Utils/InterfaceValidator.php
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Utility class for validating model interface implementations.
+ *
+ * Centralizes interface validation logic to reduce code duplication
+ * and provide consistent error messages across the AI Client.
+ *
+ * @since n.e.x.t
+ */
+class InterfaceValidator
```
JasonTheAdams: A personal preference is for utility classes to have simpler names, often referring to what they're validating.

```suggestion
class Models
```

That way it reads nicely: `Models::validateTextGeneration($model)`

Curious if @felixarntz agrees. 😁

src/Utils/ModelDiscovery.php
JasonTheAdams, line 1: If we rename `InterfaceValidator` to `Models`, then I'd merge these together into a utility for validating or retrieving models.

src/Utils/PromptNormalizer.php
```diff
@@ -0,0 +1,112 @@
...
+     * Normalizes various prompt formats into a standardized Message array.
+     *
+     * @since n.e.x.t
+     *
+     * @param string|string[]|MessagePart|MessagePart[]|Message|Message[] $prompt The prompt content in various formats.
+     * @return list<Message> Array of Message objects.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     */
+    public static function normalize($prompt): array
```
JasonTheAdams: I think we can actually simplify some of this while making it a bit more flexible by accepting a list of `string|Message|MessagePart`, looping through the list, and simply resolving to a `Message` on a per item basis.

```suggestion
    /**
     * Normalizes various prompt formats into a standardized Message array.
     *
     * @since n.e.x.t
     *
     * @param string|MessagePart||Message|list<string|MessagePart|Message> $prompt The prompt content in various formats.
     * @return list<Message> Array of Message objects.
     *
     * @throws \InvalidArgumentException If the prompt format is invalid.
     */
    public static function normalize($prompt): array
```

You can even do this at the beginning to normalize:
```php
if (!is_array($prompt)) {
    $prompt = [$prompt]
}
```

JasonTheAdams (changes requested):
Great work, @Ref34t! I left some comments and suggestions!

JasonTheAdams:
Requesting a review from @felixarntz, as I think it would be a good idea for both of us to review this one. 😄

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+     * @since n.e.x.t
+     *
+     * @param string|MessagePart|MessagePart[]|Message|Message[] $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function generateTextResult($prompt, ModelInterface $model = null): GenerativeAiResult
```
JasonTheAdams: `$model` type must be nullable if it can default to null.
```suggestion
    public static function generateTextResult($prompt, ?ModelInterface $model = null): GenerativeAiResult
```

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+     * @since n.e.x.t
+     *
+     * @param string|MessagePart|MessagePart[]|Message|Message[] $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return Generator<GenerativeAiResult> Generator yielding partial text generation results.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function streamGenerateTextResult($prompt, ModelInterface $model = null): Generator
```
JasonTheAdams:
```suggestion
    public static function streamGenerateTextResult($prompt, ?ModelInterface $model = null): Generator
```

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+     * @since n.e.x.t
+     *
+     * @param string|MessagePart|MessagePart[]|Message|Message[] $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function generateImageResult($prompt, ModelInterface $model = null): GenerativeAiResult
```
JasonTheAdams:
```suggestion
    public static function generateImageResult($prompt, ?ModelInterface $model = null): GenerativeAiResult
```

JasonTheAdams: Please update the remaining methods as well.

src/Contracts/AiClientInterface.php
felixarntz: +1, I don't think this needs an interface.

src/Operations/OperationFactory.php
felixarntz, line 1: Why do we need an `OperationFactory`? A few thoughts:

* It wasn't part of the spec of what we need.
* I don't see it being used anywhere.
* Operations are created by the respective provider implementations, not by our SDK.

src/Utils/PromptNormalizer.php
```diff
@@ -0,0 +1,112 @@
...
+use WordPress\AiClient\Messages\DTO\Message;
+use WordPress\AiClient\Messages\DTO\MessagePart;
+use WordPress\AiClient\Messages\DTO\UserMessage;
+
+/**
+ * Utility class for normalizing various prompt formats into standardized Message arrays.
+ *
+ * @since n.e.x.t
+ */
+class PromptNormalizer
```
felixarntz: I have a very similar idea implemented in https://github.com/WordPress/php-ai-client/pull/39/files#diff-6593318e4cf91319bc205a2a05d02dfee870c630eba8ddccb249c9c51243a5d7R19. The names are different, but it's the same idea.

I believe my implementation is more flexible than the one here, but I like your naming. Maybe you can take a look at mine and incorporate / reuse logic in here where it makes sense?

src/AiClient.php
felixarntz: We already discussed this a while back, I agree we could tweak it, but this is currently set up for parity with a similar pattern used in e.g. `usingModel(Google::model('gemini-2.5-flash'))`.

Can you log an issue to revise this, and for now we keep it as is? I think it's more efficient for now to work against the API specification we already have to get it done, and then iterate after completion. Nothing here is set in stone :)

src/AiClient.php
felixarntz: +1, can you move the logic in here?

Also, I thought that this traditional method based API would wrap the builder based API. In other words, the code here should call `prompt($prompt)->usingModel($model)` etc., and then generate the result using the `PromptBuilder`.

I know all of this is not there yet, so either we throw an exception that it's not implemented yet, or we wait for #49 to be completed. The heavy lifting needs to happen there, not here - and certainly not in two places :)

src/AiClient.php
```diff
@@ -0,0 +1,413 @@
...
+
+            case 'speech':
+                $resolvedModel = $model ?? ModelDiscovery::findSpeechModel(self::defaultRegistry());
+                InterfaceValidator::validateSpeechGeneration($resolvedModel);
+                /** @phpstan-ignore-next-line */
+                return $resolvedModel->generateSpeechResult($messageList);
+
+            default:
+                throw new \InvalidArgumentException("Unsupported generation type: {$type}");
+        }
```
felixarntz: See my comment above: This method does something that `PromptBuilder` terminate methods should be doing. I'm wary we're duplicating too many efforts here. `AiClient` should merely be the API entrypoint, and an easy to use wrapper around the heavy logic happening in `PromptBuilder`.

felixarntz (changes requested):
@Ref34t Overall this looks promising, but I think this PR does a couple too many things. It should use `PromptBuilder` for most its pieces, not implement the logic itself.

src/AiClient.php
JasonTheAdams: I'm kind of tempted to leave some of the logic in this PR, such as discovering the model to use, and then refactoring it in the subsequent Prompt Builder PR to have the `generateX` functions use the builder.

felixarntz: OK, that works for me - I wasn't clear about this sequence. I was assuming the `PromptBuilder` PR would come first, as `AiClient` is using it, rather than the other way around. But either way works, as long as we're all on the same page.

Ref34t:
@felixarntz @JasonTheAdams Thank you both for those great insights, I'll spend some time today and tomorrow to review all of them, fix and improve so we can keep this going

src/Contracts/AiClientInterface.php
Ref34t: Resolved by getting rid of it https://github.com/WordPress/php-ai-client/pull/55/commits/b0b77d64b02c71b9e76d87371da34595384e2985

src/AiClient.php
Ref34t: @JasonTheAdams  You're absolutely right - this is unnecessary abstraction that adds cognitive overhead for zero benefit.

@felixarntz I'll log an issue to revisit this pattern as you suggested. The pass-through method does feel like over-engineering, and the reviewer's instinct is correct. `ISSUE` https://github.com/WordPress/php-ai-client/issues/57

For now I'll keep it as-is to maintain the current API spec and focus on the remaining

src/AiClient.php
Ref34t: @JasonTheAdams  @felixarntz You're both absolutely right. I've fixed the over-engineering issues in https://github.com/WordPress/php-ai-client/pull/55/commits/3593af14c3fb066c357fa24e87232cd496fea9e0:

  **Changes made:**
  1. **Removed GenerationStrategyResolver entirely** - was unnecessary for a single use case
  2. **Replaced dynamic method calls** with simple, type-safe if/else logic in `generateResult()`
  3. **Preserved functionality** - the method still works, just with clearer implementation
  4. **Updated documentation** to indicate future PromptBuilder refactoring

  The `generateResult()` method now uses straightforward type checking:
  ```php
  if ($model instanceof TextGenerationModelInterface) {
      return self::generateTextResult($prompt, $model);
  }
  ```

 This maintains type safety, removes the over-engineered resolver class, and keeps the functionality working while
 we plan the future PromptBuilder integration. Much cleaner and more maintainable!

src/AiClient.php
Ref34t: @felixarntz should I remove the streaming methods now?

src/AiClient.php
Ref34t: @JasonTheAdams resolved through https://github.com/WordPress/php-ai-client/pull/55/commits/c75f01acbdfd43d206d793abca64c0487b05495a

src/Utils/InterfaceValidator.php
Ref34t: @JasonTheAdams Great suggestion! I've added `@phpstan-assert` annotations to all the validation methods in `InterfaceValidator`resolved through https://github.com/WordPress/php-ai-client/pull/55/commits/c75f01acbdfd43d206d793abca64c0487b05495a

src/AiClient.php
JasonTheAdams: Great! So yeah, just leave this logic for now, @Ref34t, and I'll refactor it in the subsequent PromptBuilder PR.

src/Utils/ModelDiscovery.php
Ref34t: Resolved Here https://github.com/WordPress/php-ai-client/pull/55/commits/79f4eeb051e728e0f3230251820d2f1fcbc5c736

src/Utils/InterfaceValidator.php
Ref34t: @JasonTheAdams I like the idea I've implemented it here https://github.com/WordPress/php-ai-client/pull/55/commits/79f4eeb051e728e0f3230251820d2f1fcbc5c736

src/Utils/PromptNormalizer.php
Ref34t: @JasonTheAdams Handled here https://github.com/WordPress/php-ai-client/pull/55/commits/b9c4aa696f6fd16b66f7f8ca397c7571e9d69c0d

src/AiClient.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/55/commits/ce4c14a8c6ff7dace95cc6976dbd8a334899b659

src/AiClient.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/55/commits/ce4c14a8c6ff7dace95cc6976dbd8a334899b659

src/AiClient.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/55/commits/ce4c14a8c6ff7dace95cc6976dbd8a334899b659

src/Utils/PromptNormalizer.php
Ref34t: @felixarntz I've spent some time to improve `PromptNormalizer` resolved here https://github.com/WordPress/php-ai-client/pull/55/commits/3612e5f9bd6eedd44ca71d5aff82242832decb01

src/AiClient.php
felixarntz: I think they should throw an exception that they're not implemented yet, but they should be there so that it's clear we're working on it - otherwise people will ask for it. That's what we decided elsewhere too (I think it already came up in #39 or somewhere related).

src/Utils/InterfaceValidator.php
felixarntz: I'm not sure about `Models` - that could mean anything related to models. It could be a container class for multiple models, for example.

For a class that provides static utility methods, if we want to keep them constrained to one purpose, the class name should clarify that purpose (e.g. `ModelValidator`), or if we want to have them have _any_ kind of utility methods related to the DTO, we should at least give them a suffix to indicate they're a utility class (e.g. `ModelUtil`).

WDYT?

src/AiClient.php
Ref34t: @JasonTheAdams @felixarntz resolved here https://github.com/WordPress/php-ai-client/pull/55/commits/29097f2eeb7ee61ee2808049ee5e2887b55d29f5

Ref34t:
@felixarntz @JasonTheAdams All Feedback is handled and waiting for review again thanks alot 👍

src/Operations/OperationFactory.php
JasonTheAdams: Was this comment intentionally marked as resolved, @Ref34t?

src/AiClient.php
```diff
@@ -0,0 +1,416 @@
...
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    private static function executeGeneration($prompt, ?ModelInterface $model, string $type): GenerativeAiResult
+    {
+        // TODO: Replace with PromptBuilder delegation once PR #49 is merged
+        // This should become: return self::prompt($prompt)->usingModel($model)->generate();
+        $messages = PromptNormalizer::normalize($prompt);
+        /** @var list<Message> $messageList */
+        $messageList = array_values($messages);
```
JasonTheAdams: Why do we need to make `$messagesList` like this? The normalizer should be returning a list like that. If it's not, we need to make an adjustment in there.

src/Utils/PromptNormalizer.php
JasonTheAdams: Let's iterate on  improving this to include more of the flexibility from @felixarntz work in a subsequent PR.

src/Operations/OperationFactory.php
JasonTheAdams: @felixarntz it is being used in the `AiClient` operation methods. Does that resolve this for you?

felixarntz: I would prefer if we marked all those parts as not implemented yet, via exceptions.

JasonTheAdams: Resolved in fa10649

src/AiClient.php
JasonTheAdams: Resolved in ab67b8a

src/Utils/PromptNormalizer.php
JasonTheAdams: Resolved in bdb86eb

JasonTheAdams (approved):
I made some changes, mostly to the `PromptNormalizer` — so that it takes advantage of `Message::fromArray()` and returns a single `Message` instead of an array of messages. It's actually meant to take all the parts and turn them into a single message. The `AiClient` is now able to receive an array of messages, but it doesn't need the normalizer for that — other than to verify that it is an array of messages.

src/Utils/Models.php
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * Utility class for model discovery and interface validation.
+ *
+ * Combines model auto-discovery capabilities with interface validation
+ * to provide a unified model utility service.
+ *
+ * @since n.e.x.t
+ */
+class Models
```
felixarntz: What are the methods in this class for? 

I don't see them used anywhere, and they don't make sense to me. We need to find models based on _all_ `ModelRequirements`, i.e. based on the `list<Message>` and `ModelConfig` instances, not just on a single capability like these methods do it.

Maybe I'm missing something, but to me these methods seem neither right nor necessary (at this point at least). 🤔

src/Utils/PromptNormalizer.php
```diff
@@ -0,0 +1,122 @@
...
+     * @since n.e.x.t
+     *
+     * @param string|MessagePart|Message|MessageArrayShape|list<string|MessagePart> $prompt The prompt content.
+     * @return Message The normalized message.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     *
+     * @phpstan-param string|MessagePart|Message|MessageArrayShape|list<string|MessagePart> $prompt
+     */
+    public static function normalize($prompt): Message
```
felixarntz: This is another case of me not knowing what the purpose of this is - it's not used anywhere. Since this is neither part of the `AiClient` requirements nor there is any usage of it, I'm not sure why we're implementing it here?

These situations of unused methods have me wary because they make it very possible that a method is not doing the right thing - we have no concrete need for them yet as of this PR, so how do we know the requirements such a method should fulfill?

FWIW I _do_ think we need such a method eventually (probably in the next PR, and also in #39 which already implements something similar), but it will need to return `list<Message>`, not `Message`. All the underlying model classes etc expect `list<Message>`, so even if the developer provides something that looks like a single `Message`, it should be parsed into `list<Message>`. The other benefit of `list<Message>` is that no information can get lost - the other way around it can (if someone provides `list<Message>` but the return value is `Message`).

src/AiClient.php
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    /**
+     * Sets the default provider registry instance.
+     *
+     * @since n.e.x.t
+     *
+     * @param ProviderRegistry $registry The provider registry to set as default.
+     */
+    public static function setDefaultRegistry(ProviderRegistry $registry): void
```
felixarntz: I sort of get why this might make sense, but do we have a concrete need for this? I could see the ability of _replacing_ the default registry creating a bunch of unexpected situations.

I would much rather prefer starting just with a single default registry and allowing to create custom registries (we could even have a `PromptBuilder::usingRegistry(ProviderRegistry $registry)` method). We could consider allowing to replace the default registry later, but I'm not sure it's worth the risk. In any case I'd prefer to discuss this in its own issue.

src/AiClient.php
```diff
@@ -0,0 +1,421 @@
...
+            return self::generateImageResult($prompt, $model);
+        }
+
+        if ($model instanceof TextToSpeechConversionModelInterface) {
+            return self::convertTextToSpeechResult($prompt, $model);
+        }
+
+        if ($model instanceof SpeechGenerationModelInterface) {
+            return self::generateSpeechResult($prompt, $model);
+        }
```
felixarntz: This doesn't look entirely right - a model might implement all 4 of these for example, so we can't rely purely on whether the model class implements an interface.

We need to look at `$model->getConfig()->getOutputModalities()` to determine what needs to be called. And then _if_ e.g. the output modalities only include `TEXT`, we need to check that the class implements `TextGenerationModelInterface` - similar for the other modalities and generation methods.

@JasonTheAdams Based on what we discussed earlier today in our call, the way this is implemented is actually exactly the wrong way around. It's `generateTextResult`, `generateImageResult`, etc., which should call `generateResult`, setting the `outputModalities` as needed based on the method called. `generateResult` should be the single "source of truth" main execution method. Someone could also call it directly if they manually specified the `outputModalities` or if they wanted to have multimodal output.

src/AiClient.php
```diff
@@ -0,0 +1,421 @@
...
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @param string $type The generation type.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    private static function executeGeneration($prompt, ?ModelInterface $model, string $type): GenerativeAiResult
```
felixarntz: Related to my above comment, I don't think this is the right abstraction. No need to pass a `$type` - the model config's `outputModalities` should determine what to generate.

Passing a type duplicates what those are supposed to be for, but more importantly they prevent the implementation of multimodal output (it wouldn't be always just one `$type`).

felixarntz (changes requested):
@Ref34t @JasonTheAdams I'm still seeing several issues with the implementation here. Those concerns still point back to this PR being somewhat premature - most of what happens in here needs to use logic from other classes including `PromptBuilder` which doesn't exist yet. Also see my older note in #20 about this ideally being implemented only after all the other infrastructure PRs.

I think the fact that we don't have `PromptBuilder` yet while implementing `AiClient` is leading to some confusion and duplication. Concrete feedback below.

JasonTheAdams:
Hey @Ref34t! 👋 

I spoke with @felixarntz and we're going to pause for a moment on this PR. I thought the `PromptBuilder` depends on this PR, but it's actually the other way around. I'm going to work on that and then follow up with this PR, refactoring it to depend on the `PromptBuilder` as it's intended to.

Really appreciate all your work! This will get merged! Just need to do things in the proper order. 😄

Ref34t:
@felixarntz @JasonTheAdams I completely agree with you. looking for the `PromptBuilder`

src/Utils/PromptNormalizer.php
JasonTheAdams: We cleared up that this _is_ being used in `AiClient` right now. It was my bad for switching it to return `Message` instead of `list<Message>`. I'll make a version of this in the `PromptBuilder` PR that does what we're describing.

src/Utils/Models.php
JasonTheAdams: This should be unnecessary in light of the `PromptBuilder` changes.

src/Utils/PromptNormalizer.php
JasonTheAdams, line 1: This can also be removed, now.

Ref34t: @JasonTheAdams I've completed refactoring the `AIClient`, kindly review and let me know if you have any improvements

src/AiClient.php
JasonTheAdams: @Ref34t Per Felix's suggestion, let's remove this for now. 😄

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+     * @since n.e.x.t
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the provided model doesn't support any known generation type.
+     * @throws \RuntimeException If no suitable model can be found for the prompt.
+     */
+    public static function generateResult($prompt, ?ModelInterface $model = null): GenerativeAiResult
```
felixarntz: I am just realizing that there is a severe limitation here that we didn't consider before (this is also lacking in the `ARCHITECTURE.md` doc): We need to still allow passing a `ModelConfig`, even if we don't pass a specific `ModelInterface` instance.

I just chatted about this with @JasonTheAdams, we're thinking that we should change the second parameter here to allow `ModelInterface|ModelConfig|null`:
* It needs to remain optional.
* If the dev wants to use a specific model, they should provide that instance and set the `ModelConfig` inside of that instance already.
* Else if they don't want to use a specific model but want to use specific configuration, they should directly provide the `ModelConfig`.

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param CapabilityEnum $capability The desired generation capability.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the model doesn't support the specified capability.
+     * @throws \RuntimeException If no suitable model can be found for the prompt and capability.
+     */
+    public static function generateResultWithCapability(
```
felixarntz: I don't think we need this method? When should it be used?

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+            return self::generateResultWithCapability($prompt, CapabilityEnum::speechGeneration(), $model);
+        }
+
+        throw new \InvalidArgumentException(
+            sprintf(
+                'Model "%s" must implement at least one supported generation interface ' .
+                '(TextGeneration, ImageGeneration, TextToSpeechConversion, SpeechGeneration)',
+                $model->metadata()->getId()
+            )
+        );
```
felixarntz: We should always rely on `PromptBuilder` here, even if a `ModelInterface` is provided:
* If the second parameter is a `ModelConfig`, we need to pass it to `PromptBuilder::usingModelConfig`.
* If a `ModelInterface` is provided, we need to call `PromptBuilder::usingModel` before calling `PromptBuilder::generateResult`.

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+     * @since n.e.x.t
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function generateTextResult($prompt, ?ModelInterface $model = null): GenerativeAiResult
```
felixarntz: The exact feedback that I provided for `generateResult` applies to this method too.

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+     * @since n.e.x.t
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function generateImageResult($prompt, ?ModelInterface $model = null): GenerativeAiResult
```
felixarntz: Same here, and below.

src/AiClient.php
```diff
@@ -0,0 +1,373 @@
...
+     * @since n.e.x.t
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|null $model Optional specific model to use.
+     * @return \WordPress\AiClient\Files\DTO\File The generated image file.
+     */
+    public static function generateImage($prompt, ?ModelInterface $model = null)
+    {
+        return self::generateImageResult($prompt, $model)->toFile();
+    }
```
felixarntz: Let's not include these shortcut methods, at least not for now. We can always rely on the `AiClient::generate*Result` methods for the traditional API, and appending a `toText()` or `toFile()` is very straightforward.

We can separately discuss or look out for feedback to decide whether we should add shortcuts like these in the future.

src/Builders/PromptBuilder.php
```diff
@@ -74,8 +74,7 @@ class PromptBuilder
      * @since n.e.x.t
      *
      * @param ProviderRegistry $registry The provider registry for finding suitable models.
-     * @param Prompt $prompt
-     *     Optional initial prompt content.
+     * @param Prompt $prompt Optional initial prompt content.
      */
     // phpcs:enable Generic.Files.LineLength.TooLong
     public function __construct(ProviderRegistry $registry, $prompt = null)
```
felixarntz, line 80: I realize that the part about passing a `ModelConfig` here as the third parameter is currently only implemented in #39.

Please feel free to duplicate this here so you can unblock the other changes based on my feedback above.

felixarntz (commented):
@Ref34t Looks pretty good! A few points of follow up feedback where we can simplify, and most importantly enable passing a `ModelConfig`.

felixarntz:
@Ref34t I've updated this PR just to include the updates from #65, so that you can use `PromptBuilder::usingModelConfig` in here. Back to you!

src/AiClient.php
Ref34t: yes you're right. It adds a level of complexity without a really good outcome. will remove it

src/Builders/PromptBuilder.php
Ref34t: handled

Ref34t:
@felixarntz  @JasonTheAdams I believe now that the PR is in good shape. Kindly let me know if you have any other thoughts

src/Builders/PromptBuilder.php
JasonTheAdams: I noticed below you used `usingModelConfig`, which is the right way to go about this now. 👍 

Marking this resolved!

src/Utils/PromptNormalizer.php
JasonTheAdams: This has been removed and is no longer relevant. It's now handled in the `PromptBuilder`.

src/AiClient.php
JasonTheAdams: I updated this to work like the other methods where it simply relies on the `PromptBuilder` to figure things out. Happily, it did point out a bug in the builder where it didn't infer the capability from the model if provided, which was resolved in 403e3f3.

JasonTheAdams (approved):
I helped move this along based on the feedback. I think we should be just about there! Awaiting a (hopefully) final review from @felixarntz!

felixarntz (approved):
@Ref34t @JasonTheAdams Thank you for all the iterations. Looks good to go 🎉

PR #56, opened 2025-08-19
Feature/embeddings support

Ref34t:
This is a draft PR split from #55 to enable separate review of the embeddings functionality.


JasonTheAdams:
As a note, I'm not sure when we'll add embeddings. It's not a high priority at this time and requires considerable thought downstream as to how it's used, especially in a WordPress context.

Ref34t:
@JasonTheAdams I do agree with you. I think I can try to leave it to be a separate package dependent on `PHP-AI-SDK` 
I already have a lot of trials I wanna do with RAG in PHP in the future.

PR #58, opened 2025-08-21
Fixes Function Call and Declaration DTOs

JasonTheAdams:
The `$args` in the `FunctionCall` and `FunctionDeclaration` DTOs are swapped. The Function Call args can be anything based on what the AI is told. But the Function Declaration args should be a schema. This PR corrects that.


felixarntz (approved):
@JasonTheAdams Thanks for fixing this, LGTM!

PR #59, opened 2025-08-22
Removes system message concept

JasonTheAdams:
After some discussion with @felixarntz, we decided that having _both_ system instructions in the `ModelConfig` _and_ a `SystemMessage` within `Message` arrays is confusing. This was the case because some model APIs treat system instructions as a configuration, while others expect it as the first message. But trying to do both in a client intended to be a unified API is confusing — both for the Extender and for the Implementor.

This PR removes the concept of the system message in favor of system instructions as part of the `ModelConfig`. This just focuses on how it's removed. Using the instructions in the config will be reflected in subsequent PRs.


docs/ARCHITECTURE.md
```diff
@@ -944,8 +940,8 @@ direction LR
         class ModelConfig {
             +setOutputModalities(ModalityEnum[] $modalities) void
             +getOutputModalities() ModalityEnum[]
-            +setSystemInstruction(string|MessagePart|MessagePart[]|Message $systemInstruction) void
-            +getSystemInstruction() Message?
+            +setSystemInstruction(string $systemInstruction) void
+            +getSystemInstruction() string?
```
felixarntz, line 944: Regarding this change, I was wondering whether your PR was missing that adjustment in `ModelConfig` itself, but then I checked and realized we never implemented it like this in the first place, it's always been a string - so that's perfect 😆

felixarntz (approved):
@JasonTheAdams Thank you for updating this, LGTM!

PR #63, opened 2025-08-27
Updated Inline Document order in polyfills.php File

shail-mehta:
Updated Inline Document Order in IDocumentation in `polyfills.php` File according to `Php Document Standards`

Added `@since` tag before `@param` and `@return` tag


felixarntz (approved):
@shail-mehta Thank you for contributing this!

shail-mehta:
Thank you for the review.

PR #65, opened 2025-08-27
Adds usingModelConfig method to PromptBuilder

JasonTheAdams:
This adds a `PromptBuilder::usingModelConfig` method, which receives and merges a config into the builder. This also fixes an issue that custom options in the model config were also included in `toArray()`, when they should be optional as the other keys.

As a note, I decided _not_ to worry about merging the custom options array at this point. The reason being the `PromptBuilder` will not have custom options, so if the config introduced has custom options then they'll be preserved.


felixarntz (approved):
🎉

PR #69, opened 2025-08-29
refactor: centralize model requirements logic per Issue #66

### LINKED ISSUE #66: Move responsibility for checking model requirements into `ModelRequirements` class
A small API enhancement:

* Currently, `ModelConfig` has a `toRequiredOptions` method, which is used as part of creating a `ModelRequirements` instance. This is not ideal, as it makes the `ModelConfig` class, a key component of both the implementer and extender APIs, dependent on the model discovery piece, which, while crucial, is of secondary nature when making LLM calls.
* Additionally, `ModelMetadata` has a `meetsRequirements` method, which similarly means it ties a key component to that model discovery piece.
* Both of these methods should be moved to another class.
* The original plan from `ARCHITECTURE.md` foresaw creating a `RequirementsUtil` class for this, but after discussion with @JasonTheAdams we don't think this is a solid solution either.

**We think it would be best to put both of these functionalities into `ModelRequirements`.**

For example as follows:

* `public function areMetBy(ModelMetadata $metadata): bool`
* `public static function fromPromptData(CapabilityEnum $capability, list<Message> $messages, ModelConfig $modelConfig): self`

This also means we should remove the private `PromptBuilder::getModelRequirements` method, as its logic will basically be moved to the second method above.

Ref34t:
## Summary
Implements Issue #66 by centralizing model requirements logic into the `ModelRequirements` class, improving separation of concerns between core API components and model discovery functionality.

## Changes Made
- **Move methods to a centralized location:**
- `ModelConfig::toRequiredOptions()` → `ModelRequirements::fromPromptData()`
- `ModelMetadata::meetsRequirements()` → `ModelRequirements::areMetBy()`
- Remove `PromptBuilder::getModelRequirements()` method per issue requirements

- **Update call sites:**
- PromptBuilder now calls `ModelRequirements::fromPromptData()` directly
- ProviderRegistry now calls `ModelRequirements::areMetBy()`

- **Clean up data holder classes:**
- ModelConfig and ModelMetadata are now pure data holders
- Removed unused optimization maps from ModelMetadata

- **Add comprehensive tests:**
- 4 new tests for `areMetBy()` method (matching, missing capability, unsupported values, empty requirements)
- 4 new tests for `fromPromptData()` method (text generation, chat history, multimodal input, config options)

## Architecture Impact
**Improved separation of concerns**: Core API components no longer depend on model discovery
**Single Responsibility**: `ModelRequirements` handles all requirements analysis

Fixes #66


Ref34t:
@felixarntz The refractor is ready for review, in your hands

felixarntz:
@Ref34t Just spotting this has merge conflicts - would you be able to resolve these please?

Ref34t:
@felixarntz will tackle it today sure

Ref34t:
@felixarntz The merge conflicts have been resolved. kindly review now 🥇

src/Providers/Models/DTO/ModelRequirements.php
```diff
@@ -88,6 +91,192 @@ public function getRequiredOptions(): array
...
+        if ($hasFunctionMessageParts) {
+            $requiredOptions[] = new RequiredOption(OptionEnum::functionDeclarations(), true);
+        }
+
+        // Add input modalities if we have any inputs
+        if (!empty($inputModalities)) {
+            // Remove duplicates
+            $inputModalities = array_unique($inputModalities, SORT_REGULAR);
+            $requiredOptions[] = new RequiredOption(OptionEnum::inputModalities(), array_values($inputModalities));
+        }
```
felixarntz: This is missing the logic to ensure the `RequiredOption` is only added once to the array.

I think this is best solved by reinstating the private `includeInRequiredOptions` method from before, but now in this class.

src/Providers/Models/DTO/ModelRequirements.php
```diff
@@ -88,6 +91,192 @@ public function getRequiredOptions(): array
...
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * Creates ModelRequirements from prompt data and model configuration.
+     *
+     * @since 0.2.0
```
felixarntz: While we will _very likely_ add this to 0.2.0, please always use `n.e.x.t` for any new code. This leaves the version up for a future decision, and we can replace all `n.e.x.t` instances with the correct version number right before a new release.

src/Providers/Models/DTO/ModelRequirements.php
```diff
@@ -88,6 +91,192 @@ public function getRequiredOptions(): array
...
+            );
+        }
+
+        if ($modelConfig->getOutputSchema() !== null) {
+            $requiredOptions[] = new RequiredOption(
+                OptionEnum::outputSchema(),
+                $modelConfig->getOutputSchema()
+            );
+        }
+
```
felixarntz, line 294: This method is missing several options from other `ModelConfig` parameters. Please check the original `ModelConfig::toRequiredOptions()` method and ensure everything is covered here too.

Actually, since this method is now very long, I would suggest you re-implement `toRequiredOptions(ModelConfig $modelConfig): array` here as a private method, simply to break it up, and since that part is a distinct piece of logic only dependent on the `ModelConfig` instance provided.

src/Providers/Models/DTO/ModelRequirements.php
```diff
@@ -88,6 +91,192 @@ public function getRequiredOptions(): array
...
+     *
+     * @param ModelMetadata $metadata The model metadata to check against.
+     * @return bool True if the model meets all requirements, false otherwise.
+     */
+    public function areMetBy(ModelMetadata $metadata): bool
+    {
+        // Check if all required capabilities are supported
+        foreach ($this->requiredCapabilities as $requiredCapability) {
+            $supported = false;
+            foreach ($metadata->getSupportedCapabilities() as $supportedCapability) {
```
felixarntz: Instead of a nested `foreach` loop, which can be problematic for performance, why not maintain the original approach of using lookup maps that's present in `ModelMetadata::meetsRequirements`?

You can call `getSupportedCapabilities()` and `getSupportedOptions()` once, in the beginning of this method and then create a `$capabilitiesMap` and an `$optionsMap` from them, just like today in `ModelMetadata`.

This way you can take over the more efficient code as is - which already reduces the risk of introducing new bugs.

felixarntz (changes requested):
@Ref34t This looks like a solid start!

There are a few areas where maintaining the original logic would be preferable from a performance and architecture perspective. Likely because that's not the case, there are currently a few problems that need to be addressed either way.

src/Providers/Models/DTO/ModelRequirements.php
Ref34t: Excellent approach. resolved here https://github.com/WordPress/php-ai-client/pull/69/commits/29c06c8b42564d6e834215a500edbaecac4b39ba

src/Providers/Models/DTO/ModelRequirements.php
Ref34t: Good catch. Resolved here https://github.com/WordPress/php-ai-client/pull/69/commits/29c06c8b42564d6e834215a500edbaecac4b39ba

src/Providers/Models/DTO/ModelRequirements.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/69/commits/29c06c8b42564d6e834215a500edbaecac4b39ba 
will keep this on my mind for the future too :)

src/Providers/Models/DTO/ModelRequirements.php
Ref34t: resolved here https://github.com/WordPress/php-ai-client/pull/69/commits/29c06c8b42564d6e834215a500edbaecac4b39ba

felixarntz (approved):
@Ref34t This looks excellent - thank you for the iterations!

PR #74, opened 2025-08-29
Adds PHP 7.4 compatibility checking and fix violations

JasonTheAdams:
This adds PHP 7.4 compatibility checking. It also fixes an issue it found with a polyfill.


felixarntz (approved):
Great catch!

PR #76, opened 2025-08-31
Add `AGENTS.md`

felixarntz:
[`AGENTS.md`](https://agents.md/) is an emerging standard for a central LLM assistance file supported by many providers and tools.

While previously we were not considering any provider-specific or tool-specific files in this repository for reasons of neutrality, the new `AGENTS.md` format addresses that problem and is already widely supported. For tools without out-of-the-box support, potentially it can be configured manually as an additional context file, or it can be symlinked under another file name mandated by the respective tool.

As for the contents of the file, let's start pragmatically. We don't have to cover everything right away, instead we can iterate with time based on practical usage of AI coding tools. As we notice aspects that should be captured here or aspects that confuse the LLM or tool and should be fixed, we can address them in subsequent PRs.

For reference and credit: Much of the content is inspired by or directly copied from an older discarded `CLAUDE.md` file created by @JasonTheAdams, see https://github.com/WordPress/php-ai-client/pull/25/commits/1ab1a241494bef8bdc25c76df4388725bc7ba758.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ vendor/
 
 .claude/
 CLAUDE.md
+.clinerules/
 .cursor/
 GEMINI.md
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,98 @@
+# PHP AI Client SDK
+
+## Project brief
+
+The project implements a provider agnostic PHP AI client SDK to communicate with any generative AI models of various capabilities using a uniform API.
+
+The project is stewarded by [WordPress AI Team](https://make.wordpress.org/ai/) members and contributors. It is however WordPress agnostic, so that any PHP project can use it.
+
+## High-level architecture
+
+The project architecture has several key design principles:
+
+### API layers
+
+1. **Implementer API**: For developers using the SDK to add AI features to their applications
+    - Fluent API: Chain methods for readable, declarative code (e.g. `AiClient::prompt('...')->generateText()`)
+    - Traditional API: Method-based approach with arrays of arguments (e.g. `AiClient::generateText('...')`)
+
+2. **Extender API**: For developers adding new providers, models, or extending functionality
+    - Provider Registry system for managing available AI providers
+    - Model discovery based on capabilities and requirements
+
+### Core concepts
+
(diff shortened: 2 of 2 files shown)


JasonTheAdams (approved):
Love it!! 🎉

justlevine:
Was this evaled in any way, or compared with 
- no AGENTS.md
- an AGENTS.md that just says "Follow the guidelines in CONTRIBUTING.md"

I don't see a mention on this PR, and the source was for scaffolding an empty project which then still had to undergo significant work _after the file's removal_. 

Would be helpful to keep a record, since we know this is inevitably going to be used as prior art elsewhere.

cc: 
- https://github.com/WordPress/abilities-api/issues/30
- https://github.com/WordPress/mcp-adapter/issues/9

felixarntz:
@justlevine It wasn't, but I think it's fair to assume that having a basic `AGENTS.md` will make AI tools do better than without having it.

Regarding your second point, to just tell it to use `CONTRIBUTING.md`, that's a fair point. We could see if one works better than the other.

As the PR description says, this is just a starting point. How well it works will have to be based on practical usage, and iterations are always welcome. But without having anything in place (like this), each contributor using an AI tool will build their own such files in a vacuum, and having this here gives us a central place to work with and iterate on.

Crowd-sourcing the testing and feedback on this file is going to be more powerful than only one of us doing a deep dive (but inevitably biased) session on evaluating what works well.

justlevine:
> but I think it's fair to assume that having a basic AGENTS.md will make AI tools do better than without having it.

Anecdotally my experience differs, 
- when contributing to public projects with a file made for a specific type/gen of model (e.g. circa GPT3.5 when using 4.1, or 4o when using Gemini or Sonnet)
- when contributing to public and private projects where the sources of truth / human docs differ from the agentic docs.

(More in https://github.com/WordPress/abilities-api/issues/30#issuecomment-3215525735 )

> Crowd-sourcing the testing and feedback on this file is going to be more powerful than only one of us doing a deep dive (but inevitably biased) session on evaluating what works well.

Definitely agree with this this. The important thing here IMO is to leave to document our assumptions before they become Chesterton fences. Now done 🚀

PR #78, opened 2025-09-01
Implement Exception Hierarchy For Better Error Handling

Ref34t:
**Summary**
Implements  exception hierarchy to improve error handling across the WordPress AI Client ecosystem, directly addressing feedback from PR 2 in https://github.com/WordPress/wp-ai-client

**Changes**

**New Exception Classes**

- AiClientExceptionInterface - Base interface for catch-all exception handling
- NetworkException - HTTP transport errors, connection failures, timeouts
- RequestException - API authentication failures, rate limiting, malformed requests
- InvalidArgumentException - Project-scoped extension of \InvalidArgumentException
- RuntimeException - Project-scoped extension of \RuntimeException

  
 **Updated Code**

- Updated existing exception usage in AiClient.php, GenerativeAiResult.php, FunctionResponse.php
- Modified ResponseException to extend RequestException for better categorization
- Added comprehensive test coverage

**Benefits**

Unified Exception Handling:
  ```
try {
      $result = AiClient::prompt('Generate content')->generateText();
  } catch (AiClientExceptionInterface $e) {
      // Catches ALL AI Client exceptions
  }

```
Better Error Categorization:
  - Network issues → NetworkException
  - API/Auth issues → RequestException
  - Invalid arguments → InvalidArgumentException

  
**Related**

Addresses feedback from 2 review comments about improving exception handling for wp-ai-client integration. https://github.com/WordPress/wp-ai-client

### CODE REVIEW

**Comment by Ref34t** [comment]
@felixarntz This implements the exception hierarchy you suggested in PR 2 in the wp-ai-client review. Now developers can catch all AI Client exceptions with catch  (AiClientExceptionInterface $e) for unified error handling, while still having specific exception types for different error categories (network, request, validation). The implementation is minimal and focused - exactly what was requested for better WP-ai-client integration.

File: src/Exceptions/NetworkException.php [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Exception thrown for network-related errors.
+ *
+ * This includes HTTP transport errors, connection failures,
+ * timeouts, and other network-related issues.
+ *
+ * @since 0.2.0
+ */
+class NetworkException extends RuntimeException implements AiClientExceptionInterface
```
felixarntz: What is this exception class used for? We should only introduce exceptions that we have concrete usage for.

I can see how a `NetworkException` might be useful, but then we should also make sure it's used where applicable.

src/Exceptions/RequestException.php
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Exception thrown for AI API request errors.
+ *
+ * This includes authentication failures, rate limiting,
+ * malformed requests, and API-specific errors.
+ *
+ * @since 0.2.0
+ */
+class RequestException extends RuntimeException implements AiClientExceptionInterface
```
felixarntz: What would this class be used for? I'm thinking that `RequestException` is most likely only relevant for bad request data (i.e. as an extension of `InvalidArgumentException`).

Similarly to my other comment, we need to actually use it if we add it here. There are at least several places in the provider implementation code or its abstract base classes where this could be used.

I think this would also be good to throw in scenarios when a `Response` comes back with a `400 Bad Request` status code. In that case it means an invalid argument was set that our code didn't catch, but the API did.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,13 @@
...
-use Exception;
+use WordPress\AiClient\Exceptions\RequestException;
 
 /**
  * Exception class for HTTP response errors.
  *
  * @since 0.1.0
  */
-class ResponseException extends Exception
+class ResponseException extends RequestException
```
felixarntz: This shouldn't extend `RequestException` - a `ResponseException` is definitely something different.

I think the main use-case for `ResponseException` is when the response data is unexpected - which should never happen, unless a provider changed in ways our code is not aware of. That's IMO the primary use-case for this.

felixarntz (commented):
@Ref34t Thank you for getting this started!

A few points of feedback below, and one higher-level point here: I think we should put the exceptions into specific directories. For example:
* `InvalidArgumentException` and `RuntimeException` into `src/Common/Exception`
* `NetworkException` and `RequestException` into `src/Providers/Http/Exception` (where we already have `ResponseException`)

@JasonTheAdams It would be great to get your review as well - including regarding my own feedback.

JasonTheAdams:
I agree with @felixarntz! I like the direction, but we need to see these exceptions being used in a few places to indicate their concrete usage.

Ref34t:
@felixarntz @JasonTheAdams
I've spent some time refactoring and improving the enhancements, you can see below what I did:

  **Architectural Reorganization**

  Moved exceptions to proper directories:
  - InvalidArgumentException & RuntimeException → src/Common/Exception/
  - NetworkException & RequestException → src/Providers/Http/Exception/
  - Updated 50+ import statements across the entire codebase

  **Concrete Usage Examples** 

  - NetworkException: PSR-18 network error handling in HttpTransporter::send()
  - RequestException: 400 Bad Request response handling in HttpTransporter::send()
  - ResponseException: Missing API data handling in OpenAiModelMetadataDirectory
  - InvalidArgumentException: 50+ validation locations throughout codebase
  - RuntimeException: 40+ operational error locations throughout codebase

  **Proper Inheritance**

  Fixed inheritance relationships:
  - RequestException extends InvalidArgumentException (for bad request data validation)
  - ResponseException extends RuntimeException (for unexpected provider responses)
  - NetworkException extends RuntimeException (for network connectivity issues)
  

  **Enhanced Developer Experience (Static factory methods suggestion)**

 "Simply adding static from* methods to the various  *Exception classes"

 RequestException (3 methods):
 ```
 RequestException::fromInvalidParam('OpenAI', 'temperature', 'Must be between 0 and 2');
 RequestException::fromBadRequestResponse('OpenAI', $response);
 RequestException::fromBadRequestToUri($uri, 'Invalid  JSON payload');

```
  ResponseException (4 methods):
 ```
 ResponseException::fromMissingData('OpenAI', 'data');
 ResponseException::fromUnexpectedStructure('OpenAI', 'array', 'string');
 ResponseException::fromMalformedResponse('OpenAI', 'Invalid JSON format');
 ResponseException::fromParsingFailure('OpenAI', 'model list', $jsonException);

```
 NetworkException (5 methods):
 ```
NetworkException::fromPsr18NetworkException($uri, $networkException);
NetworkException::fromConnectionFailure($uri, 'Connection refused');
NetworkException::fromTimeout($uri, 'request', 30);
NetworkException::fromDnsFailure('api.openai.com');
NetworkException::fromSslError($uri, 'Certificate  validation failed');

```

Kindly review and let me know if you have any concerned

**Review by JasonTheAdams — COMMENTED** [review]
Thanks, @Ref34t! I'm not sure I see the point in creating our own `InvalidArgumentException` and `RuntimeException` classes. Do you, @felixarntz?

**Comment by felixarntz** [comment]
@JasonTheAdams I do actually. It's simply about providing classes that can implement a common interface to identify "this is a specific exception from our PHP AI Client SDK".

Consuming code can then decide whether they want to catch e.g. `RuntimeException` and/or `InvalidArgumentException` or whether they want to do a single catch-all for "known" exceptions from our SDK only by catching `AiClientExceptionInterface`.

**Comment by JasonTheAdams** [comment]
I like that, @felixarntz! Admittedly, I hadn't noticed the common Interface. I like that idea of being able to catch exceptions thrown by this package.

As a note, I think we should update the AGENTS.md to include a note that all exceptions should use our own, including primitives, and if a primitive doesn't exist it should be created. Not sure if we should add this elsewhere, too. If we're going down this road it's important that we're consistent.

File: src/Common/Exception/InvalidArgumentException.php [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+
+use WordPress\AiClient\Exceptions\AiClientExceptionInterface;
+
+/**
+ * Exception thrown when an invalid argument is provided.
+ *
+ * This extends PHP's built-in InvalidArgumentException while implementing
+ * the AI Client exception interface for consistent catch handling.
+ *
+ * @since 0.2.0
```
JasonTheAdams: Here and throughout, always use `n.e.x.t` because we don't want to assume we know the version something will be released in.
```suggestion
 * @since n.e.x.t
```

JasonTheAdams (changes requested):
I'm tracking now, @Ref34t! Good work!

I don't think we need all these static methods. As far as I can tell, there all used only one time. If that's the case, the abstraction isn't necessary and we can just let the implementing code determine the code, message, and such.

felixarntz (commented):
@Ref34t @JasonTheAdams I do think some of the static methods are useful - just not all of them.

Per my previous feedback, @Ref34t it would be great if you could review the codebase for where exceptions are thrown and update these exceptions to use the new exception classes and new static methods if applicable.

We should not introduce any exception classes or any static methods that aren't used anywhere - I think that's a good guiding principle. Therefore it's important that we update any existing exceptions to use the new approach, and if there are none for a specific exception or static method, we know it's not useful, at least not at this point.

Ref34t:
very interesting Convo Champs @felixarntz  @JasonTheAdams  🥇 
I'll work it this today and bring the best from it based on my idea and your thoughts

Ref34t:
- Updated @since tags from "0.2.0" to "n.e.x.t" in all exception classes

- Migrated 30+ throw statements across 5 core files to use custom exception classes:
  - ProviderRegistry.php: 5 exceptions converted
  - PromptBuilder.php: 25+ exceptions converted
  - Files/DTO/File.php: 5 exceptions converted
  - Http/DTO/Request.php and Response.php: exceptions converted
  
- Added `ResponseException::fromBadResponse()` static method based on actual usage
- Simplified `ResponseUtil` from 30 lines to 1 line using new static method
- Removed 9 unused static methods across NetworkException, RequestException, and ResponseException
- All exceptions now use project's custom classes instead of PHP built-ins
- Maintained `AiClientExceptionInterface` implementation for unified error handling
- Added exception handling requirements to AGENTS.md coding standards

Ref34t:
The remaining files that were updated 

 1. src/Providers/Http/Traits/WithHttpTransporterTrait.php
 2. src/Providers/Http/Traits/WithRequestAuthenticationTrait.php
 3. src/ProviderImplementations/OpenAi/OpenAiProvider.php
 4. src/ProviderImplementations/Google/GoogleProvider.php
 5. src/ProviderImplementations/Google/GoogleModelMetadataDirectory.php
 6. src/ProviderImplementations/Anthropic/AnthropicProvider.php
 7. src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
 8. src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
 9. src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleImageGenerationModel.php
 10. src/Providers/Models/DTO/ModelRequirements.php
 11. src/Providers/Models/DTO/ModelMetadata.php
 12. src/Providers/Models/DTO/ModelConfig.php
 13. src/Providers/Models/DTO/SupportedOption.php
 14. src/Providers/DTO/ProviderModelsMetadata.php
 15. src/Providers/Contracts/ProviderOperationsHandlerInterface.php
 16. src/Providers/Contracts/ProviderInterface.php
 17. src/Providers/Contracts/ModelMetadataDirectoryInterface.php
 18. src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
 19. src/Tools/DTO/FunctionCall.php
 20. src/Results/DTO/Candidate.php
 21. src/Messages/DTO/MessagePart.php
 22. src/Common/AbstractEnum.php
 23. src/Common/AbstractDataTransferObject.php

src/Common/Exception/InvalidArgumentException.php
Ref34t: resolved

src/Files/DTO/File.php
```diff
@@ -375,12 +375,7 @@ private function determineMimeType(
 
             $extension = pathinfo($cleanPath, PATHINFO_EXTENSION);
             if (!empty($extension)) {
-                try {
-                    return MimeType::fromExtension($extension);
-                } catch (InvalidArgumentException $e) {
-                    // Extension not recognized, continue to error
-                    unset($e);
-                }
+                return MimeType::fromExtension($extension);
```
JasonTheAdams: Did we change this behavior for some reason?

src/Providers/Http/Exception/NetworkException.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Exception thrown for network-related errors.
+ *
+ * This includes HTTP transport errors, connection failures,
+ * timeouts, and other network-related issues.
+ *
+ * @since n.e.x.t
+ */
+class NetworkException extends RuntimeException
```
JasonTheAdams, line 19: I wonder if this should implement `\Psr\Http\Client\NetworkExceptionInterface`?

src/Providers/Http/Exception/RequestException.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Exception thrown for AI API request errors due to bad request data.
+ *
+ * This includes malformed requests, invalid parameters, and scenarios
+ * where the API responds with a 400 Bad Request status code indicating
+ * that our code didn't catch an invalid argument but the API did.
+ *
+ * @since n.e.x.t
+ */
+class RequestException extends InvalidArgumentException
```
JasonTheAdams: Same, should implement `\Psr\Http\Client\RequestExceptionInterface`?

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,84 @@
...
  * Exception class for HTTP response errors.
  *
+ * This is used when response data is unexpected or malformed,
+ * typically indicating that a provider changed in ways our code
+ * is not aware of or when parsing response data fails.
+ *
  * @since 0.1.0
  */
-class ResponseException extends Exception
+class ResponseException extends RuntimeException
```
JasonTheAdams, line 18: Similar, should this implement the PSR-18 `ClientExceptionInterface`?

AGENTS.md
```diff
@@ -83,6 +83,15 @@ The production code in `src` follows a structured namespace hierarchy under the
 
 All parameters, return values, and properties must use explicit type hints, except in cases where providing the correct type hint would be impossible given limitations of backward compatibility with PHP 7.4. In any case, concrete type annotations using PHPStan should be present.
 
+### Exception handling
+
+All exceptions must use the project's custom exception classes rather than PHP built-in exceptions. This includes:
+
+- Use `WordPress\AiClient\Common\Exception\InvalidArgumentException` instead of PHP's `\InvalidArgumentException`
+- Use `WordPress\AiClient\Common\Exception\RuntimeException` instead of PHP's `\RuntimeException`
```
JasonTheAdams:
```suggestion
- Use the custom primitive exceptions in `WordPress\AiClient\Common\Exception\` instead of the PHP primitive exceptions
- If a PHP primitive exception is needed that doesn't have a custom exception counterpart, then create one that extends the primitive and implements AiClientExceptionInterface
```

JasonTheAdams (changes requested):
Awesome! Thanks for updating all those, @Ref34t! I left a couple comments and questions.

src/Files/DTO/File.php
Ref34t: This was my mistake. When I updated the exception imports, I accidentally removed the try-catch block instead of just updating the import. This breaks the graceful handling of unrecognized file extensions.

Thanks for catching this!

resolved here https://github.com/WordPress/php-ai-client/pull/78/commits/469f46e803658735f2050c7191723cae2cbbe165

src/Providers/Http/Exception/NetworkException.php
Ref34t: I'll accept this suggestion! Implemented all three PSR-18 interfaces with getRequest() methods and updated `HttpTransporter` accordingly

resolved in https://github.com/WordPress/php-ai-client/pull/78/commits/04e767fd644b15bb3b9fc43a39af4bfbad74aa0a

src/Providers/Http/Exception/RequestException.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/04e767fd644b15bb3b9fc43a39af4bfbad74aa0a

src/Providers/Http/Exception/ResponseException.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/04e767fd644b15bb3b9fc43a39af4bfbad74aa0a

AGENTS.md
Ref34t: I  agree with you. Resolved here https://github.com/WordPress/php-ai-client/pull/78/commits/d2f5d62cfde4faad4900dd557c15615c9676c908

src/Exceptions/AiClientExceptionInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
felixarntz, line 1: This file should probably live in `Common\Contracts`, or in `Common\Exception\Contracts` - not in its own separate top-level namespace.

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
```diff
@@ -53,9 +53,7 @@ protected function parseResponseToModelMetadataList(Response $response): array
         /** @var ModelsResponseData $responseData */
         $responseData = $response->getData();
         if (!isset($responseData['data']) || !$responseData['data']) {
-            throw new RuntimeException(
-                'Unexpected API response: Missing the data key.'
-            );
+            throw ResponseException::fromMissingData('OpenAI', 'data');
```
felixarntz, line 56: The same change still needs to be made in the `AnthropicModelMetadataDirectory` and in `GoogleModelMetadataDirectory`.

src/Providers/Http/HttpTransporter.php
```diff
@@ -68,7 +71,30 @@ public function __construct(
...
+                $e
+            );
+        }
+
+        // Check for 400 Bad Request responses indicating invalid request data
+        if ($psr7Response->getStatusCode() === 400) {
+            $body = (string) $psr7Response->getBody();
+            $errorDetail = $body ? substr($body, 0, 200) : 'Invalid request parameters';
+            throw RequestException::fromBadRequest($psr7Request, $errorDetail);
+        }
```
felixarntz: This should not be handled here, because we want to make sure the caller of `HttpTransporter::send` is in full control over what to do with the response.

Therefore, this should be moved to `ResponseUtil::throwIfNotSuccessful`, so the caller can decide to use it, or not.

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -33,35 +33,6 @@ public static function throwIfNotSuccessful(Response $response): void
...
-        } elseif (
-            is_array($data) &&
-            isset($data['message']) &&
-            is_string($data['message'])
-        ) {
-            $errorMessage .= ' ' . $data['message'];
-        }
-
-        throw new ResponseException($errorMessage, $response->getStatusCode());
+        throw ResponseException::fromBadResponse($response);
```
felixarntz: I'm thinking that we may be want to move the `fromBadResponse` method from `ResponseException` to `RequestException`? Technically, the lines are blurry, but I think I have a good reason:

The only thing I personally consider a response exception is when something in the response itself is wrong (e.g. missing an expected field in a successful response), which should almost never happen. Everything else IMO is more suitable in `RequestException` - because if the response has an error status code, we can consider the request failed, not the response.

Would love to get your perspective on this too @JasonTheAdams!

felixarntz (commented):
@Ref34t This looks really close now. A few last points of feedback.

felixarntz:
@Ref34t Note the merge conflict with `ModelRequirements`, it would be great if you could please fix that too now that #69 was merged.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,85 @@
...
+     * Creates a ResponseException for missing expected data.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $apiName The name of the API/provider.
+     * @param string $fieldName The field that was expected but missing.
+     * @param string $context Additional context about where the field was expected.
+     * @return self
+     */
+    public static function fromMissingData(string $apiName, string $fieldName, string $context = ''): self
```
felixarntz: There are several other places where this should be used:
* All exceptions thrown in `AbstractOpenAiCompatibleTextGenerationModel::parseResponseToGenerativeAiResult`
* All exceptions thrown in `AbstractOpenAiCompatibleTextGenerationModel::parseResponseChoiceToCandidate`
* All exceptions thrown in `AbstractOpenAiCompatibleTextGenerationModel::parseResponseChoiceMessageParts`
* All exceptions thrown in `AbstractOpenAiCompatibleImageGenerationModel::parseResponseToGenerativeAiResult`
* All exceptions thrown in `AbstractOpenAiCompatibleImageGenerationModel::parseResponseChoiceToCandidate`

felixarntz: For some of these where it's about something more specific than a _missing_ field, we could have another `public static function fromInvalidData(string $apiName, string $message): self` here.

src/Providers/Http/Util/ResponseUtil.php
JasonTheAdams: So in frameworks like Symfony they distinguish this a bit differently, they use:
* `HttpExceptionInterface`, which indicates a 4xx or 5xx response error
* `ClientExceptionInterface`, which extends `HttpExceptionInterface` and is specific to 4xx response errors
* `ServerExceptionInterface`, which extends `HttpExceptionInterface` and is specific to 5xx response errors
* `TransportExceptionInterface`, which means there was a network failure

So they don't actually distinguish it in terms of "response" and "request", I think because of the ambiguity being described here. So if we like that convention better we could go that route.

felixarntz: I'm not opposed to doing it that way, however none of the exceptions you mention actually does what our `ResponseException` here is currently covering: It's the (very likely only hypothetical) scenario where a response does not include the data we expect. This can only happen if e.g. OpenAI made a breaking change to its API response shape tomorrow and we didn't know about it. So very unlikely to ever run into it, but possible - however it's only relevant if the API response itself came back successfully.

JasonTheAdams: I see your point. I'm good with `RequestException`, or if we want something a touch more granular: `UnsupportedRequestException`.

JasonTheAdams: Ok! @felixarntz and I chatted a bit and landed on this suggestion:
* `RequestException` - base class and thrown within Providers
* `RedirectException extends RequestException` - thrown for 3xx responses
* `ClientException extends RequestException` - thrown for 4xx responses
* `ServerException extends RequestException` - thrown for 5xx responses
* `NetworkException extends RequestException` - thrown for network errors

Sound good to you, @Ref34t? 😄

src/Exceptions/AiClientExceptionInterface.php
Ref34t: resolved here https://github.com/WordPress/php-ai-client/pull/78/commits/a6bbaf2cafd54b96ab36579d32a57dbafb155c64

src/Providers/Http/Util/ResponseUtil.php
Ref34t: I like the final outcome you decided upon. Will try to conclude it 👍

src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
Ref34t: resolved here https://github.com/WordPress/php-ai-client/pull/78/commits/7c742aad6d887823ec8e7dc11814609e9eb44cc5

src/Providers/Http/Exception/ResponseException.php
JasonTheAdams: Looks like this has been resolved. 👍

JasonTheAdams (approved):
This is looking good to me! Thanks, @Ref34t!

src/Providers/Http/Exception/RequestException.php
```diff
@@ -0,0 +1,84 @@
...
+ * @since n.e.x.t
+ */
+class RequestException extends InvalidArgumentException implements RequestExceptionInterface
+{
+    /**
+     * The request that failed.
+     *
+     * @var RequestInterface|null
+     */
+    private ?RequestInterface $request = null;
```
felixarntz: We should not store the lower-level PSR `RequestInterface` in `RequestException` class and its child classes - at the very least we shouldn't expose it publicly. We already have our own `Request` DTO for this purpose, so all public APIs should rely on that.

Can we change this accordingly here and in `NetworkException` please?

src/Providers/Http/Exception/RequestException.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Returns the request that failed.
+     *
+     * @since n.e.x.t
+     *
+     * @return RequestInterface
+     * @throws \RuntimeException If no request is available (when using deprecated fromBadRequestToUri)
+     */
+    public function getRequest(): RequestInterface
```
felixarntz: See above, this should use our `Request` DTO instead.

src/Providers/Http/Exception/RequestException.php
```diff
@@ -0,0 +1,84 @@
...
+     * @param string $uri The URI that was requested.
+     * @param string $errorDetail Details about what made the request bad.
+     * @return self
+     *
+     * @deprecated Use fromBadRequest() with RequestInterface for PSR-18 compliance
+     */
+    public static function fromBadRequestToUri(string $uri, string $errorDetail = 'Invalid request parameters'): self
+    {
+        return new self(sprintf('Bad request to %s (400): %s', $uri, $errorDetail));
+    }
```
felixarntz: What's the rationale for having these two methods here rather than on `ClientException`, which is explicitly meant for 4xx errors?

src/Providers/Http/Exception/RequestException.php
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Exception thrown for AI API request errors due to bad request data.
+ *
+ * This includes malformed requests, invalid parameters, and scenarios
+ * where the API responds with a 400 Bad Request status code indicating
+ * that our code didn't catch an invalid argument but the API did.
+ *
+ * @since n.e.x.t
+ */
+class RequestException extends InvalidArgumentException implements RequestExceptionInterface
```
felixarntz: This is something @JasonTheAdams discussed but still didn't fully find a solution for: It doesn't feel right that a generic `RequestException` extends `InvalidArgumentException`. Some request exceptions fit into that (e.g. for 400 errors), others fit better to extend `RuntimeException` (such as 500 errors).

Brainstorming a bit: Do we even need the base `RequestException` class? As far as I can tell, we could have only the more specific `ClientException`, `NetworkException`, `RedirectException`, and `ServerException`, and be fine. Then `ClientException` could extend `InvalidArgumentException` while the other three could extend `RuntimeException`.

IMO this would clean things up nicely. What do you both think?

felixarntz: Why does this implement the PSR `RequestExceptionInterface`? Unless there is a concrete need for that, I don't think we should do this. Same reason we don't implement the PSR `RequestInterface` and `ResponseInterface` interfaces on our `Request` and `Response` DTOs.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,98 @@
...
  * Exception class for HTTP response errors.
  *
+ * This is used when response data is unexpected or malformed,
+ * typically indicating that a provider changed in ways our code
+ * is not aware of or when parsing response data fails.
+ *
  * @since 0.1.0
  */
-class ResponseException extends Exception
+class ResponseException extends RuntimeException implements ClientExceptionInterface
```
felixarntz: See above, can we avoid implementing the PSR exception here? Unless there's a practical reason for it?

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,98 @@
...
+            is_string($data['error'])
+        ) {
+            $errorMessage .= ' ' . $data['error'];
+        } elseif (
+            is_array($data) &&
+            isset($data['message']) &&
+            is_string($data['message'])
+        ) {
+            $errorMessage .= ' ' . $data['message'];
+        }
```
felixarntz: This exact logic is now present in multiple places. It would be great to centralize it somewhere, either in a utility class or in a trait.

felixarntz (commented):
A few more points of follow up feedback. This is getting close though.

One question I have for both of you @Ref34t @JasonTheAdams

src/Providers/Http/Exception/RequestException.php
JasonTheAdams: The only thing we'll want to be sure of, here, is that a PSR-7 `RequestInterface` is always converted to our `Request` DTO before it's passed to the exception. This is likely not an issue, but the only caveat that came to mind.

src/Providers/Http/Exception/RequestException.php
JasonTheAdams: Ah, a couple notes here. First, I agree that this shouldn't extend `InvalidArgumentException`. I think it should either extend our `Exception` primitive or `RuntimeException`.

If you recall, @felixarntz, we made `RequestException` a class instead of an interface so that it could be used inside of Providers/Models when there isn't a status code but we are in "request" territory. Do you not think we need this anymore?

felixarntz: @JasonTheAdams Hmm, right. I'm not sure we need it anymore. I think it would be totally reasonable to use (our) `InvalidArgumentException` in cases of invalid arguments that _we_ catch before an actual request. `ClientException` (for actual 4xx errors from a request) would extend `InvalidArgumentException`, so a developer could still catch `InvalidArgumentException` and be sure to get any relevant scenarios of "invalid arguments" (whether caught by our code or by the actual API).

Therefore I'm leaning towards getting rid of `RequestException`.

src/Providers/Http/Exception/ResponseException.php
Ref34t: resolved

src/Providers/Http/Exception/ResponseException.php
Ref34t: got it. resolved

src/Providers/Http/Exception/RequestException.php
Ref34t: got it. resolved

src/Providers/Http/Exception/RequestException.php
Ref34t: got it. resolved

Ref34t:
@felixarntz Here is the summary of the iteration I did:

1. Build centralized error message extraction utility (4314134)

  - **What**: Created a centralized `ErrorMessageExtractor` utility class
  - **Purpose**: Standardizes parsing of API error responses across different formats
  - **Key Features**:
    - Handles common patterns: {"error":  {"message": "..."}}, {"error": "..."}, {"message": "..."}
    - Added `fromPsrRequest` factory method to Request DTO for PSR-7 conversion
  - Files: src/Providers/Http/DTO/Request.php, src/Providers/Http/Utilities/ErrorMessageExtractor.php
  
  
2. Implement semantic exception hierarchy with proper inheritance patterns (0f18425)

  - **What:** Restructured exception classes with proper object-oriented inheritance
  - **Key Changes:**
    - Enhanced `ClientException` with more sophisticated error handling
    - Refactored `NetworkException` for better network-specific error management
    - Removed `RequestException`
    

   
3. Update exception classes to use centralized error extraction and Request DTOs (2867cbc)

  - **What:** Integrated the new `ErrorMessageExtractor` utility across exception classes
  - **Key Changes:**
    - Simplified `RedirectException`, `ResponseException`, and `ServerException`
    - Reduced code duplication by using centralized error extraction
    - Leveraged Request DTOs for consistent data handling
 
4. Update exception tests for new hierarchy  structure (e16fba4)

  - **What:** Updated test suite to match the new exception hierarchy
  - **Purpose:** Ensured test coverage remains accurate after architectural changes
  - **Files:** tests/unit/Exceptions/ExceptionsTest.php

src/Providers/Http/Exception/RequestException.php
JasonTheAdams: That's fine be me. Do we want to turn `RequestException` into an interface so folks can catch any of the subsequent exceptions easily?

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -15,53 +18,54 @@
...
 
-        throw new ResponseException($errorMessage, $response->getStatusCode());
+        // 4xx Client Errors
+        if ($statusCode >= 400 && $statusCode < 500) {
+            // Special handling for 400 Bad Request
+            if ($statusCode === 400) {
+                $body = (string) $response->getBody();
+                $errorDetail = $body ? substr($body, 0, 200) : 'Invalid request parameters';
+                throw ClientException::fromBadRequestResponse($errorDetail);
+            }
```
felixarntz: Why is this needed? When is the response body a simple string?

I think this should use the same logic as below just like the other 4xx error codes, unless I'm missing something?

src/Providers/Http/Exception/ClientException.php
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * This method extracts error details from common API response formats
+     * and creates an exception with a descriptive message and status code.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response that failed.
+     * @return self
+     */
+    public static function fromClientError(Response $response): self
```
felixarntz: Maybe better call this `fromClientErrorResponse`, since the necessary parameter is a `Response`? This also aligns better with the other methods, making it more intutive.

src/Providers/Http/Exception/ClientException.php
```diff
@@ -0,0 +1,127 @@
...
+     * @param string $uri The URI that was requested.
+     * @param string $errorDetail Details about what made the request bad.
+     * @return self
+     *
+     * @deprecated Use fromBadRequest() with RequestInterface for better type safety
+     */
+    public static function fromBadRequestToUri(string $uri, string $errorDetail = 'Invalid request parameters'): self
+    {
+        return new self(sprintf('Bad request to %s (400): %s', $uri, $errorDetail), 400);
+    }
```
felixarntz: Why do we need all these? Where are they used?

I think it would be sufficient to have the single `fromClientErrorResponse` below.

We can easily centralize any special handling we want to have for 400 errors, e.g. in that method you could add a check that changes the message to `Bad request (400): %s` instead. Even better, you could have a short map of a few important error codes and their reasons, similar to how you have in `fromRedirectResponse`.

On another note: Let's be consistent with the error messages - some messages here include the URI while others don't.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,80 @@
...
+     *
+     * This method extracts error details from common API response formats
+     * and creates an exception with a descriptive message and status code.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response that failed.
+     * @return self
+     */
+    public static function fromBadResponse(Response $response): self
```
felixarntz: Where is this used? I don't think this is relevant, as `ResponseException` should only be used for problems within a successful response (i.e. something that should really never happen, but in principle still might).

src/Providers/Http/Exception/ServerException.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param Response $response The HTTP response that failed.
+     * @return self
+     */
+    public static function fromServerError(Response $response): self
+    {
+        $errorMessage = sprintf(
+            'Server error (%d): Request failed due to server-side issue',
+            $response->getStatusCode()
+        );
```
felixarntz: See my above comment, you could do something similar to `fromRedirectResponse` here and use a different beginning of the message, based on the error reason string for a few common 5xx status codes.

src/Providers/Http/Exception/ServerException.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * This method extracts error details from common API response formats
+     * and creates an exception with a descriptive message and status code.
+     *
+     * @since n.e.x.t
+     *
+     * @param Response $response The HTTP response that failed.
+     * @return self
+     */
+    public static function fromServerError(Response $response): self
```
felixarntz: Similar to my other comment, I think this should be renamed to `fromServerErrorResponse`.

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -15,53 +18,54 @@
...
+            throw ClientException::fromClientError($response);
+        }
+
+        // 5xx Server Errors
+        if ($statusCode >= 500 && $statusCode < 600) {
+            throw ServerException::fromServerError($response);
+        }
+
+        // Other unsuccessful responses - should be extremely rare
+        throw ResponseException::fromBadResponse($response);
```
felixarntz: Per my other comment on `ResponseException`, that exception class shouldn't be used here.

You're right that this should be extremely rare, but the problem really is that it's an invalid HTTP status code. So I think we should instead throw a `RuntimeException` here, with a message like `Response returned invalid status code: %s`.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleImageGenerationModel.php
```diff
@@ -361,8 +361,10 @@ protected function parseResponseChoiceToCandidate(
         } elseif (isset($choiceData['b64_json']) && is_string($choiceData['b64_json'])) {
             $imageFile = new File($choiceData['b64_json'], $expectedMimeType);
         } else {
-            throw new RuntimeException(
-                'Unexpected API response: Each choice must contain either a url or b64_json key with a string value.'
+            throw ResponseException::fromMissingData(
```
felixarntz: Let's use `fromInvalidData` here instead and keep the original message. The `$context` parameter of `fromMissingData` is a bit confusing, we should always use just the `$fieldName` parameter for it.

src/Providers/Http/Exception/ResponseException.php
```diff
@@ -4,13 +4,80 @@
...
+     * Creates a ResponseException for missing expected data.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $apiName The name of the API/provider.
+     * @param string $fieldName The field that was expected but missing.
+     * @param string $context Additional context about where the field was expected.
+     * @return self
+     */
+    public static function fromMissingData(string $apiName, string $fieldName, string $context = ''): self
```
felixarntz: See my other comment, can you get rid of the `$context` parameter. The `$fieldName` parameter alone is sufficient to express everything needed.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -621,14 +621,18 @@ protected function parseResponseChoiceToCandidate(array $choiceData): Candidate
...
         }
 
         if (!isset($choiceData['finish_reason']) || !is_string($choiceData['finish_reason'])) {
-            throw new RuntimeException(
-                'Unexpected API response: Each choice must contain a finish_reason key with a string value.'
+            throw ResponseException::fromMissingData(
+                $this->providerMetadata()->getName(),
+                'finish_reason',
+                'choice data'
             );
```
felixarntz, line 639: Instead of the `$context` parameter here, please specify the _full_ `$fieldName`:
* `choices[{i}].message`
* `choices[{i}].finish_reason`

Replace the `{i}` with the relevant choice index where the inner field is missing. To do so, you can add a new `int $index` argument to this method.

felixarntz (commented):
@Ref34t Thank you for the iterations. A few more follow-up comments below.

src/Providers/Http/Exception/RequestException.php
felixarntz: I think we can leave that out for now, we can always add it later if we feel it's reasonable. I feel most people will anyway just want to catch any of our SDK's exceptions, or maybe split by `InvalidArgumentException` and `RuntimeException`, or something like that.

src/Providers/Http/Util/ResponseUtil.php
Ref34t: That's right. The special handling was inconsistent and made assumptions about the response body format.

I've removed the special 400 case in commit 5e765f8. Now all 4xx errors use the same `ClientException::fromClientError()` logic with the centralized `ErrorMessageExtractor`, which properly handles structured API response formats like `{"error": {"message": "..."}}` instead of assuming string bodies.

This makes the error handling consistent cross all client error codes. Thanks for catching this!

src/Providers/Http/Exception/ClientException.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/01bafd9ef6ba9bb3c8c2e52335803edce322f08c

src/Providers/Http/Exception/ClientException.php
Ref34t: @felixarntz Done! Removed all unnecessary static methods in commit 7bbaadd.

Now ClientException has only `fromClientErrorResponse()` with focused status code mapping (400, 401, 403, 404, 422, 429) similar to RedirectException. All error messages are now consistent without URI discrepancies.

The three removed methods were unused and creating unnecessary complexity. Thanks for the simplification guidance!

src/Providers/Http/Exception/ResponseException.php
Ref34t: @felixarntz resolved in commit 4cc060b. Removed the unused `fromBadResponse()` method completely. The ResponseException should only be for problems within successful responses, not invalid status codes.

src/Providers/Http/Exception/ServerException.php
Ref34t: @felixarntz Done in commit 4cc060b! Added status code mapping to ServerException similar  to RedirectException. Now includes focused mapping for common 5xx codes (500, 502, 503, 504, 507) with proper status text like "Server error (500 Internal Server Error): ...".

src/Providers/Http/Exception/ServerException.php
Ref34t: @felixarntz Renamed in commit 4cc060b. Changed `fromServerError()` to `fromServerErrorResponse()` for consistency with other exception classes. It is much clearer that the parameter is a Response object.

src/Providers/Http/Util/ResponseUtil.php
Ref34t: @felixarntz resolved in commit 4cc060b. ResponseUtil now throws `RuntimeException` instead `ResponseException::fromBadResponse()` for invalid status codes, using your exact suggested message: "Response returned invalid status code: %s". Makes more sense now! thanks

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/436549f28e8ea355c11d17277c6cae781087889d

src/Providers/Http/Exception/ResponseException.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/436549f28e8ea355c11d17277c6cae781087889d

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleImageGenerationModel.php
Ref34t: resolved https://github.com/WordPress/php-ai-client/pull/78/commits/436549f28e8ea355c11d17277c6cae781087889d

src/Providers/Http/Exception/ClientException.php
felixarntz: One follow up feedback below, I think this is still not the message format I suggested above.

src/Providers/Http/Exception/ClientException.php
```diff
@@ -0,0 +1,88 @@
...
+        $statusTexts = [
+            400 => 'Bad Request',
+            401 => 'Unauthorized',
+            403 => 'Forbidden',
+            404 => 'Not Found',
+            422 => 'Unprocessable Entity',
+            429 => 'Too Many Requests',
+        ];
+
+        $statusText = $statusTexts[$statusCode] ?? 'Client Error';
```
felixarntz: With this, it can lead to some weird messages like this: `Client error (405 Client Error): Request was rejected due to client-side issue`

Instead of this, can we change the message completely depending on whether a specific status text is set or not?

Something like:
```php
if (isset($statusTexts[$statusCode])) {
    $errorMessage = sprintf('%s (%d)', $statusTexts[$statusCode], $statusCode);
} else {
    $errorMessage = sprintf('Client error (%d): Request was rejected due to client-side issue', $statusCode);
}
```

src/Providers/Http/Exception/RedirectException.php
```diff
@@ -0,0 +1,62 @@
...
+            300 => 'Multiple Choices',
+            301 => 'Moved Permanently',
+            302 => 'Found',
+            303 => 'See Other',
+            304 => 'Not Modified',
+            307 => 'Temporary Redirect',
+            308 => 'Permanent Redirect',
+        ];
+
+        $statusText = $statusTexts[$statusCode] ?? 'Redirect';
```
felixarntz: Can we do the same approach here that I suggested above in `ClientException`?

src/Providers/Http/Exception/ServerException.php
```diff
@@ -0,0 +1,59 @@
...
+        $statusCode = $response->getStatusCode();
+        $statusTexts = [
+            500 => 'Internal Server Error',
+            502 => 'Bad Gateway',
+            503 => 'Service Unavailable',
+            504 => 'Gateway Timeout',
+            507 => 'Insufficient Storage',
+        ];
+
+        $statusText = $statusTexts[$statusCode] ?? 'Server Error';
```
felixarntz: And same here.

felixarntz (approved):
@Ref34t Thank you for the iterations, this looks pretty much good to go!

Just a few last points, but those are quick to address, so I can do that inline and merge.

PR #79, opened 2025-09-02
Allow retrieving all registered provider IDs from registry

felixarntz:
Our registry so far does not allow this, but it's a crucial requirement IMO. Code needs to be able to cycle through all registered providers, e.g. in our WordPress specific package this need is already relevant.

### DIFF
--- a/src/Providers/ProviderRegistry.php
+++ b/src/Providers/ProviderRegistry.php
@@ -114,6 +114,18 @@ class ProviderRegistry implements WithHttpTransporterInterface
     }
 
     /**
+     * Gets a list of all registered provider IDs.
+     *
+     * @since 0.1.0
+     *
+     * @return list<string> List of registered provider IDs.
+     */
+    public function getRegisteredProviderIds(): array
+    {
+        return array_keys($this->providerClassNames);
+    }
+
+    /**
      * Checks if a provider is registered.
      *
      * @since 0.1.0
--- a/src/Providers/ProviderRegistry.php
+++ b/src/Providers/ProviderRegistry.php
@@ -118,7 +118,7 @@ class ProviderRegistry implements WithHttpTransporterInterface
      *
      * @since 0.1.0
      *
-     * @return list<string> List of registered provider IDs.
+     * @return list<class-string<ProviderInterface> List of registered provider IDs.
      */
     public function getRegisteredProviderIds(): array
     {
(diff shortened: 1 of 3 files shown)


src/Providers/ProviderRegistry.php
```diff
@@ -113,6 +113,18 @@ public function registerProvider(string $className): void
         $this->registeredClassNames[$className] = true;
     }
 
+    /**
+     * Gets a list of all registered provider IDs.
+     *
+     * @since 0.1.0
+     *
+     * @return list<string> List of registered provider IDs.
```
JasonTheAdams, line 121: We can make this even clearer, which will help the IDE as well:
```suggestion
     * @return list<class-string<ProviderInterface> List of registered provider IDs.
```

JasonTheAdams (approved):
Good add! One really minor suggestion, so I'll go ahead and approve. 👍

src/Providers/ProviderRegistry.php
```diff
@@ -113,6 +113,18 @@ public function registerProvider(string $className): void
         $this->registeredClassNames[$className] = true;
     }
 
+    /**
+     * Gets a list of all registered provider IDs.
+     *
+     * @since 0.1.0
+     *
+     * @return list<class-string<ProviderInterface> List of registered provider IDs.
```
JasonTheAdams:
```suggestion
     * @return list<class-string<ProviderInterface>> List of registered provider IDs.
```

felixarntz:
@JasonTheAdams The change you made assumes this is supposed to return provider class names, but I was intending for it to return provider IDs, per the method name. So something is not right here now, one way or another.

Should we maybe add methods for both? `getRegisteredProviderIds` and `getRegisteredProviderClassNames`?

JasonTheAdams:
Oh goodness, my bad. I saw the `$array_keys($this->providerClassnames)` and made the assumption. 🤦‍♂️ 

Why not have it return the `id => class` structure we have internally?

felixarntz:
@JasonTheAdams 
> Why not have it return the `id => class` structure we have internally?

IMO too low-level. I think methods for exactly what you want (one or the other) are more intuitive.

JasonTheAdams:
@felixarntz Works for me! In that case, lets stick with what you have. The dev can use `getProviderClassName` to get the class for a given id.

Sorry for messing it up. 😆

JasonTheAdams:
I put it back to how you had it when I first approved. So let's just pretend I wasn't "helpful" and merge this bad boy. 😂

PR #80, opened 2025-09-02
Fix empty tool calls and arguments in OpenAI compatible models

Jameswlepage:
**Summary**

Fixes critical API compatibility issues when AI models make function/tool calls with no parameters. The SDK was generating invalid payloads rejected by OpenAI and Anthropic APIs.

**Changes**

- **Fix empty tool_calls handling**: Only include `tool_calls` field when array contains items (OpenAI rejects empty arrays)
- **Fix empty arguments encoding**: Convert empty PHP arrays to objects before JSON encoding (APIs expect `{}` not `[]` for empty parameters)
- **Add test coverage**: New tests for both edge cases to prevent regression

**Technical Details**

The issue manifests in two ways:

1. **OpenAI API Error**: `Invalid 'messages[X].tool_calls': empty array. Expected at least one item`
   - Occurs when `tool_calls: []` is included in messages
   - Fixed by conditionally including the field only when non-empty

2. **Anthropic API Error**: `messages.X.content.0.tool_use.input: Input should be a valid dictionary`
   - Occurs when empty arguments encode as `"[]"` instead of `"{}"`
   - Fixed by converting empty arrays to objects before JSON encoding

**Testing**

```bash
# Run new tests
./vendor/bin/phpunit --filter="testGetMessagePartToolCallDataEmptyArguments|testPrepareMessagesParamNoToolCalls"

# Verify all tests pass
composer run phpunit  # 878 tests, 3547 assertions ✓
composer run phpstan  # No errors ✓
```

**Impact**

- Enables WordPress Abilities API integration (many parameterless functions)
- Fixes production failures with function calling
- Maintains full backward compatibility

**Related**

- Community reports: [OpenAI empty tool_calls error](https://community.openai.com/t/openai-badrequesterror-400-error-message-invalid-messages-1-tool-calls-empty-array/1112016)
- Similar issues reported with Anthropic's Claude API in August 2024


src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -394,12 +401,17 @@ protected function getMessagePartToolCallData(MessagePart $part): ?array
                     'The function call typed message part must contain a function call.'
                 );
             }
+            $args = $functionCall->getArgs();
+            // Ensure empty arrays become empty objects for JSON encoding.
+            if (is_array($args) && empty($args)) {
+                $args = (object) array();
+            }
```
felixarntz: This one is tricky. Technically speaking, for JSON-encoding this should still only be turned into `stdClass` when the expected input shape is `object`. While that's often the case, it only depends on the schema - it could also be `array`, in which case leaving it a PHP array would be correct.

What happens if the `parameters` JSON schema specified for the function expects an actual `array` (indexed/numeric)? Does the OpenAI API always fail if it's an empty array? If so, that would seem like a flaw in the API (although, if that's the case, of course we still need to fix / work around it in our SDK).

tests/unit/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModelTest.php
```diff
@@ -742,6 +762,32 @@ public function testGetMessagePartToolCallDataFunctionCallPart(): void
...
+     * Tests getMessagePartToolCallData() with empty arguments (should encode as empty object).
+     *
+     * @return void
+     */
+    public function testGetMessagePartToolCallDataEmptyArguments(): void
+    {
+        $functionCall = new FunctionCall(
+            'call_1',
+            'list_capabilities',
+            [] // Empty arguments array
```
felixarntz, line 777: See my above comment, the problem is in PHP it's undefined whether this is an empty associative array (`object` in JSON schema) or an empty indexed/numeric array (`array` in JSON schema). Just to clarify what I meant above.

tests/unit/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModelTest.php
```diff
@@ -518,6 +518,26 @@ public function testPrepareMessagesParamModelMessageWithFunctionCall(): void
...
+    public function testPrepareMessagesParamNoToolCalls(): void
+    {
+        $message = new Message(
+            MessageRoleEnum::model(),
+            [new MessagePart('Hello, I am a simple text response.')]
+        );
+        
+        $model = $this->createModel();
+        $prepared = $model->exposePrepareMessagesParam([$message], null);
+
```
felixarntz: PHPCS errors here:
```suggestion

        $model = $this->createModel();
        $prepared = $model->exposePrepareMessagesParam([$message], null);

```

felixarntz (commented):
@Jameswlepage Thanks, great catch!

For the empty `tool_calls`, this makes total sense. About the other change, I have a follow up question - technically always making it an empty object doesn't seem right (unless this is truly the only way to make it work against the OpenAI API).

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
felixarntz: I thought about this a bit more, and while my concern is still valid in theory, in practice the OpenAI API specification does not support arrays as the root value. So it's fair to assume the data should be an object in this scenario.

felixarntz (approved):
@Jameswlepage It's been a while, so I hope it's fine that I finished this up for you! It was only a few minor details anyway.

Hoping to get 0.2.0 out by end of this week, it's long overdue!

PR #83, opened 2025-09-02
Adds MessageBuilder class

### LINKED ISSUE #81: Implement the `MessageBuilder` class
This was originally part of the MVP (0.1.0), but was moved out as it wasn't a crucial blocker (see #20). It's still important to include though to simplify working with multiple messages in a single prompt, so let's implement it for the next release.

cc @JasonTheAdams

JasonTheAdams:
Resolves #81

This introduces the `MessageBuilder` class — Fluent API for constructing messages, very similar to the `PromptBuilder`. 

Note that I've intentionally moved structure validation to the moment of termination because I don't think we should be checking things against the role and such until the implementor considers the builder complete.


docs/ARCHITECTURE.md
```diff
@@ -351,14 +351,14 @@ direction LR
...
         class MessageBuilder {
-            +usingRole(MessageRole $role) self
+            +usingRole(MessageRoleEnum $role) self
+            +usingUserRole() self
+            +usingModelRole() self
             +withText(string $text) self
-            +withImageFile(File $file) self
-            +withAudioFile(File $file) self
-            +withVideoFile(File $file) self
+            +withFile($file, ?string $mimeType) self
```
felixarntz, line 357: Great catch updating these! Since you're already on it, could you revise the `PromptBuilder` definition here as well, at least replacing similar outdated instances?

src/Builders/MessageBuilder.php
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param string|null $text Optional initial text content.
+     * @param MessageRoleEnum|null $role Optional role.
+     */
+    public function __construct(?string $text = null, ?MessageRoleEnum $role = null)
```
felixarntz: This should allow more than just `$text` - let's make it `mixed $input`, similar to `PromptBuilder`.

We should add parity here for all possibilities that identify something _within_ a `Message`, at least anything that can become a single `MessagePart`.

At least this should support passing a `string` (text) or a `MessagePart`, or a `MessagePartArrayShape`. Maybe even the lower-level pieces to turn into a `MessagePart` (which `string` is already one of), like `File`, `FunctionCall`, or `FunctionResponse`.

felixarntz (changes requested):
@JasonTheAdams Thanks for putting this together! Looks great, except one piece that needs to be revised to make it more flexible, especially as we already do it more flexibly in `PromptBuilder`.

src/Builders/MessageBuilder.php
JasonTheAdams: Good call! Done in 13a3529

docs/ARCHITECTURE.md
JasonTheAdams: Resolved in b51ca57

felixarntz (approved):
@JasonTheAdams Love it!

Issue #8, opened 2025-07-07
Development roadmap


JasonTheAdams:
I think it would be helpful to take the requirements doc and derive a development roadmap. I think what exactly is the MVP can be defined as this is thought through, but isn't my initial intent to define. The phases are prioritized based on the greater WP AI initiative, with the goal being that we can move to other projects within that effort faster, without having to wait for this to be done in its entirety. Lastly, this is intended to be iterative with no backwards compatibility issues. 

I'm not sure if we want to update the requirements doc with any/all of this, or if it's best to consider them separate. I'm good either way.

## 0.1.0: Basic Universal Client
**Goal:** Providers can register models with the Client, and the client is capable of the most fundamental and common tasks  and model configurations.

- [x] Client provides a unified API for interacting with various models
- [x] Basic model parameters (e.g. temperature and top P) are standardized
- [x] Providers and models can be registered
- [x] Basic AI modalities
  - [x] Text
  - [x] Image
  - [x] Audio
  - [x] Video
- [x] Client can receive a chat history
- [x] Tools can be defined and passed to the model
- [x] Resources can be defined and passed to the model
- [x] Registered models can specify whether they support tools and/or resources
- [x] Model prompts support tool calls and results
- [x] Structured output from AI

## 0.2.0: Battle testing improvements
**Goal:** With 0.1.0 out, others can begin testing. This release will provide basic API improvements and address any issues discovered during testing.

- [x] All registered providers can be retrieved (see #79)
- [ ] Individual Messages can be fluently constructed (see #83)
- [ ] Exceptions are distinct and catchable by implementing systems (see #78)
- [x] Bugs discovered in testing are addressed

## 0.3.0: Streaming support
**Goal:** Client is capable of providing a streamed chat UX.

- [ ] Response streaming is supported

## 0.4.0: Embedding
**Goal:** The client is capable of supporting embeds for use with vector databases

- [ ] Embeddings can be generated by the client

## No Gos
The following are ideas and directions we are intentionally not taking at this time. Some indefinitely and others in the meantime

* The client is not responsible for persisting chat history in any way. It can receive chat history for context, but it's up to the implementing code to persist it.

Jameswlepage:
For both phase 2 and 3, I'd drop the "agent" terminology. 

Phase 2 seems like it should be Phase 1, in that it's not much of a larger scope, and passing message back scroll is an essential part of any chat-completion/instruct model nowadays (ie; system prompts). I can see streaming adding some complexity though.

[...]

Issue #6, opened 2025-06-26
Supporting specific, common providers


JasonTheAdams:
During the AI Chat in Slack, we discussed two questions surrounding support for specific providers (e.g. Google, Anthropic):

1. Which providers should be supported as part of the project and how do we determine that?
2. Where should the modules for said providers live?

It was proposed that we consider Google, Anthropic, and OpenAI. The criteria was mostly popularity, though there was admitted not to be much hard data around that.

As to where, @felixarntz laid out three possible scenarios (quoting from Slack):
* No providers are bundled, and all providers are separate packages.
* All providers are bundled, and there are clear requirements that a provider needs to meet to be considered.
* Only providers that meet a set of clear requirements are bundled, while other providers may live in separate packages.

Creating this issue so we can continue the discussion!

Jameswlepage:
> The criteria was mostly popularity, though there was admitted not to be much hard data around that.

I don't think we'll really get hard data for this outside of OpenRouter popularity. I think it's okay to start off with "the big three" and then grow from there.

> Only providers that meet a set of clear requirements are bundled, while other providers may live in separate packages.

This to me sounds like a happy medium. We have the core providers starting with the big three bundled and then the remainder easily extensible so folks can use whatever makes sense to them. 

I'll also note that OpenAI is a relatively well used pattern from other providers (ie DeepSeek). Even Google and Anthropic support the OpenAI API spec.

felixarntz:
Let's continue this discussion to make a long-term decision, but just chiming in here in regards to the MVP (also see #8 and #20):

* I think for the MVP we can go with implementing "the big three" providers and putting them right into the SDK. This will keep things simple for the initial implementation, where it's most important to have the provider code and the API it's based on close together.
* I envision the MVP getting a version like `0.1.0` (i.e. not `1.0.0`), so it is okay if we later decided to remove them from the SDK and have them in separate packages instead.

JasonTheAdams:
I'm down with "the big three". I think we may find ourselves pulling those out into separate packages/plugins sooner than later because their SDKs require a version of PHP higher than we want.

justlevine:
Been rethinking the "big three" approach after a talk with some junior engineers I work with and a few freelance "hobbyists" i know, and the implications if the models we're starting with / offering "officially" are all pay-to-play.

I think it's important we launch with support for a self-hosted model. I'm happy to own this task if that's what it takes. 
[...]

PR #91, opened 2025-09-09
Add xAI as an OpenAI compatible provider

dkotter:
This PR adds in a new AI provider, [xAI](https://docs.x.ai/docs/overview), using the existing OpenAI compatible classes. [xAI claims](https://docs.x.ai/docs/api-reference) to have full compatibility with the OpenAI REST API, making it fairly straightforward to add.

I've added support for both text generation and image generation, as well as support for both text and image input modalities.

This can be quickly tested using our `cli.php` command:

```bash
XAI_API_KEY=xai-KEY php cli.php 'Hello' --providerId=xai
XAI_API_KEY=xai-KEY php cli.php 'Hello' --providerId=xai --modelId=grok-3-mini
```

> [!NOTE]  
> I know there's an [existing roadmap](https://github.com/WordPress/php-ai-client/issues/8) and adding new providers is currently not on that roadmap. I also know there was a [discussion](https://github.com/WordPress/php-ai-client/issues/6) on which providers to initially support, with xAI not being on that list.
>
> Happy to close this out if we don't want to consider this right now or we can also look at supporting this as a stand-alone, 3rd party thing under the 10up/Fueled repo if we don't want to support xAI directly in this package

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -69,6 +69,28 @@ $texts = AiClient::generateTextResult(
 )->toTexts();
 ```
 
+#### Generate multiple text candidates using an xAI model
+
+##### Fluent API
+
+```php
+$texts = AiClient::prompt('Write a 2-verse poem about PHP.')
+    ->usingModel(XAi::model('grok-3-mini'))
+    ->generateTexts(4);
+```
+
+##### Traditional API
+
+```php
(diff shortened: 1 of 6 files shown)


dkotter:
Closing this out now as 1 - this is outdated and would need some updates and 2 - these providers are being moved to separate repos so likely to do the same here

Issue #89, opened 2025-09-05
Provide a way to get `content` instead of `reasoning_content`


dkotter:
If you use a model that supports reasoning, the response you'll get back contains both a `reasoning_content` part and `content` part. The former contains the full reasoning response while the latter contains the final response, which I think is what you normally are wanting.

Right now if I'm using a reasoning model and I call the [`generateText` method](https://github.com/WordPress/php-ai-client/blob/6d722dec419acaf7deaf351abe7d73ba117b0070/src/Builders/PromptBuilder.php#L952), this returns a `GenerativeAiResult` object and then calls the [`toText` method](https://github.com/WordPress/php-ai-client/blob/6d722dec419acaf7deaf351abe7d73ba117b0070/src/Results/DTO/GenerativeAiResult.php#L209) of that.

This method will return the first message part that contains text and when we first parse the API response, we set the `reasoning_content` piece as the [first message part](https://github.com/WordPress/php-ai-client/blob/6d722dec419acaf7deaf351abe7d73ba117b0070/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php#L694) and so this is what gets returned. Same thing happens if you initially use `generateTexts`.

As far as I can see, there isn't a way, other than parsing the response myself, to force it to use the `content` piece instead of `reasoning_content`. I see in #52 we added a way to distinguish between content and thought, seems like it may be useful to use that within these generate methods to determine which message part we return. Curious for thoughts on that though or if there's a better way to achieve this.

felixarntz:
@dkotter Great point. I'd consider this simply a bug. `toText()`, `toTexts()` as well as any other `to*()` methods in `GenerativeAiResult` should only return the data from a part that is the `content` channel, i.e. not a `thought` channel.

I think this would be the most reasonable approach to address it. Since it only is meant to return the first text from the first part, these methods are meant for simple-structure responses that typically consist of only a single part.

So returning only the `content` (vs the `thought` _and_ `content`) makes most sense to me. If you want to get both, then you're properly dealing with multi-part content, i.e. custom processing is probably required.

dkotter:
Opened #92 to make this change, though open to discussion on the best approach there

felixarntz:
This should have been marked as closed a while ago, it was completed via #92.

PR #92, opened 2025-09-11
Modify the `toText` and `toTexts` methods of the `GenerativeAiResult` class to only return the content channel

dkotter:
As described in #89, if you use a reasoning model, the content returned contains both `reasoning_content` and `content` items. The `reasoning_content` is always first in the array and so when `toText` is returned, `reasoning_content` will always be returned instead of `content`.

This PR adjusts the logic of those methods to only return items that are in the `content` channel.

May be worth a discussion on if we:

1. Want methods that allow you to pass in what channel type you want instead of hardcoding that to `content`. Right now if you want the `reasoning_content`, you'd have to parse that out yourself from the response instead of using the built-in helper methods
2. Is there any risk to only returning the `content` channel? For instance, if an LLM for some reason only returns the reasoning result, this new approach will return nothing (or will throw an exception). I debated modifying these methods to fallback to returning the first result if we don't find any result that matches the `content` channel but wasn't sure if we actually wanted it to function that way

### DIFF
--- a/src/Results/DTO/GenerativeAiResult.php
+++ b/src/Results/DTO/GenerativeAiResult.php
@@ -320,7 +320,7 @@ class GenerativeAiResult extends AbstractDataTransferObject implements ResultInt
     }
 
     /**
-     * Converts all candidates to text array.
+     * Converts all content candidates to a text array.
      *
      * @since 0.1.0
      *
@@ -332,8 +332,9 @@ class GenerativeAiResult extends AbstractDataTransferObject implements ResultInt
         foreach ($this->candidates as $candidate) {
             $message = $candidate->getMessage();
             foreach ($message->getParts() as $part) {
+                $channel = $part->getChannel();
                 $text = $part->getText();
-                if ($text !== null) {
+                if (MessagePartChannelEnum::content() === $channel && $text !== null) {
                     $texts[] = $text;
                     break;
                 }
--- a/src/Results/DTO/GenerativeAiResult.php
+++ b/src/Results/DTO/GenerativeAiResult.php
@@ -9,6 +9,7 @@ use RuntimeException;
 use WordPress\AiClient\Common\AbstractDataTransferObject;
 use WordPress\AiClient\Files\DTO\File;
 use WordPress\AiClient\Messages\DTO\Message;
+use WordPress\AiClient\Messages\Enums\MessagePartChannelEnum;
 use WordPress\AiClient\Providers\DTO\ProviderMetadata;
 use WordPress\AiClient\Providers\Models\DTO\ModelMetadata;
 use WordPress\AiClient\Results\Contracts\ResultInterface;
@@ -199,7 +200,7 @@ class GenerativeAiResult extends AbstractDataTransferObject implements ResultInt
     }
 
     /**
-     * Converts the first candidate to text.
+     * Converts the first content candidate to text.
      *
      * @since 0.1.0
      *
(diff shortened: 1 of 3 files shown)


src/Results/DTO/GenerativeAiResult.php
```diff
@@ -208,15 +209,18 @@ public function hasMultipleCandidates(): bool
...
+                $channel = $part->getChannel();
+                $text = $part->getText();
+                if (MessagePartChannelEnum::content() === $channel && $text !== null) {
+                    return $text;
+                }
             }
         }
 
-        throw new RuntimeException('No text content found in first candidate');
+        throw new RuntimeException('No text content found in the candidates');
```
felixarntz: This is changing more than what the issue/PR is discussing. The method is still intended to return the content from the first candidate, i.e. it should be used when you generate a single candidate. If you generate multiple candidates, `toTexts()` should be used instead.

So I think you can keep this code as it was except for adding the `MessagePartChannelEnum::content() === $channel` check. (Nit-pick: You can shorten the check and avoid the extra `use` statement by checking `$channel->isContent()` instead.)

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -328,8 +332,9 @@ public function toTexts(): array
         foreach ($this->candidates as $candidate) {
             $message = $candidate->getMessage();
             foreach ($message->getParts() as $part) {
+                $channel = $part->getChannel();
                 $text = $part->getText();
-                if ($text !== null) {
+                if (MessagePartChannelEnum::content() === $channel && $text !== null) {
```
felixarntz: Same nit-pick as above, just a suggestion:
```suggestion
                if ($channel->isContent() && $text !== null) {
```

felixarntz (changes requested):
@dkotter Thank you for fixing this! The PR changes a bit more than fixing the bug, so I think that needs to be simplified.

The other part is that this exact same bug can also occur for all the other `to*()` methods in `GenerativeAiResult`, so I think we should fix it there too to only return content bits. Granted, it's likely impossible as of today for any provider to include things like images etc in "thought" or "reasoning", but as per the DTO it's in principle possible. So ideally we adjust all these similar methods to only consider "content" parts.

src/Results/DTO/GenerativeAiResult.php
dkotter: 🤦🏻 I read this wrong, was thinking each candidate was an individual message but each candidate contains a message that has multiple parts, which it's the parts we care about here. I'll update this

dkotter:
> The other part is that this exact same bug can also occur for all the other to*() methods in GenerativeAiResult, so I think we should fix it there too to only return content bits. Granted, it's likely impossible as of today for any provider to include things like images etc in "thought" or "reasoning", but as per the DTO it's in principle possible. So ideally we adjust all these similar methods to only consider "content" parts

I've updated the `toFile` and `toFiles` methods now to have the same check, which the following methods all end up using one of those two:

- `toImageFile`
- `toAudioFile`
- `toVideoFile`
- `toImageFiles`
- `toAudioFiles`
- `toVideoFiles`

There's also `toMessage` and `toMessages` but seems like we'd want those to return the entire Message object and not remove extra details (like `reasoning_content`) but let me know if I should update those as well

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -199,7 +199,7 @@ public function hasMultipleCandidates(): bool
     }
 
     /**
-     * Converts the first candidate to text.
+     * Converts the first content candidate to text.
```
felixarntz:
```suggestion
     * Converts the first candidate to text.
     *
     * Only text from the content channel is considered. Text within model thought or reasoning is ignored.
```

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -220,7 +221,7 @@ public function toText(): string
     }
 
     /**
-     * Converts the first candidate to a file.
+     * Converts the first content candidate to a file.
```
felixarntz:
```suggestion
     * Converts the first candidate to a file.
     *
     * Only files from the content channel are considered. Files within model thought or reasoning are ignored.
```

src/Results/DTO/GenerativeAiResult.php
```diff
@@ -316,7 +318,7 @@ public function toMessage(): Message
     }
 
     /**
-     * Converts all candidates to text array.
+     * Converts all content candidates to a text array.
```
felixarntz:
```suggestion
     * Converts all candidates to text.
```

felixarntz (approved):
@dkotter Looks great! Just making some minor doc nit-pick updates.

You're completely right regarding `toMessage()` and `toMessages()`, these should maintain the messages as is.

PR #93, opened 2025-09-13
Explore alternative `AGENTS.md` file

felixarntz:
I researched different `AGENTS.md` structures in other projects and explored a few iterations created by an LLM (I used Gemini CLI) with some guidance on the particular structure).

I'm opening this PR with one that I particularly liked, simply for consideration. 

**I'm not saying that this is holistically particularly better than the current one, but I think it might be helpful for us to compare the different sections.** Maybe there are things we'll want to combine from both variants.

On this note, I would also highly encourage other folks that use a coding agent to explore and share their drafts.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,98 +1,90 @@
-# PHP AI Client SDK
+# PHP AI Client - Coding Agent Guide
 
-## Project brief
+## Project Overview
 
-The project implements a provider agnostic PHP AI client SDK to communicate with any generative AI models of various capabilities using a uniform API.
+The PHP AI Client is a provider-agnostic PHP SDK designed to communicate with any generative AI model across various capabilities through a uniform API. It is a WordPress-agnostic PHP package that can be used in any PHP project, providing a flexible and extensible way to integrate AI features.
 
-The project is stewarded by [WordPress AI Team](https://make.wordpress.org/ai/) members and contributors. It is however WordPress agnostic, so that any PHP project can use it.
+## Commands & Scripts
 
-## High-level architecture
+The following commands are available for development and can be run using Composer:
 
-The project architecture has several key design principles:
+*   `composer lint`: Runs all linting checks.
+*   `composer phpcs`: Runs PHP_CodeSniffer to check for coding standards violations.
+*   `composer phpcbf`: Automatically fixes coding standards violations that can be fixed automatically.
+*   `composer phpstan`: Runs PHPStan for static analysis.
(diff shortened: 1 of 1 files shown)


justlevine:
@felixarntz do you have an eval that can be used to compare the differences?

felixarntz:
@justlevine I don't.

I agree that would probably be useful to a certain extent, but at the same time it wouldn't capture the diversity of the different coding assistant tools that different contributors use anyway.

JasonTheAdams:
I think this new structure is simple enough. The structure I had was from `/claude init` and was also AI-generated. I don't think there's likely to be a significant difference in model performance based on structure, so long as it's sensible. I do understand that models tend to do better with more straightforward instructions, which I think this structure does a good job of. 👍

jeffpaul:
The thing I still have a hard time with this sort of a file is that the content seems wholly fit for something like a CONTRIBUTING.md or DEVELOPERS.md file.  There's not much if anything really in this AGENTS.md file that otherwise wouldn't be helpful context for a human onboarding to the project.

felixarntz:
@jeffpaul Agreed. In theory, I wish there wouldn't need to be this duplication. That being said, based on my anecdotal experience, agents do better with being the relevant content passed directly instead of e.g. putting into `AGENTS.md` something like "You must read the `CONTRIBUTING.md` file." 😐 

An alternative idea would be to fully combine the two and either require agents to be configured to rely on `CONTRIBUTING.md` (either by actually configuring it or by symlinking `AGENTS.md`), or the other way around by centralizing some shared information in `AGENTS.md` and in `CONTRIBUTING.md` refer contributors to that file for those respective pieces of information.

But that would also have problems. Having to look between the two files would be unnecessary mental overhead for new contributors. And maybe more importantly, there are strategies that work well with LLMs that would not go well with contributors (for example **SCREAMING LANGUAGE LIKE THIS** 🙃).

My 2 cents is, while I don't like duplication, I think we'll need to optimize for contributors and for LLMs separately, which will mean some duplication.

JasonTheAdams:
I kind of like the idea of simply referring folks in the `CONTRIBUTING.md` doc to check out this doc for a crash course on project structure and architectural details. That reduces redundancy a bit. There isn't much of anything in the `AGENTS.md` doc that wouldn't be useful to a person, even if the writing style is a bit odd.

JasonTheAdams (approved):
I feel like we can merge this and iterate more. The discussion doesn't seem to be moving away from the importance of having this.

PR #98, opened 2025-09-22
Add further popular AI assistant config folders to gitignore.

felixarntz:
This simply adds more popular AI assistant config directories to `.gitignore`, allowing contributors to use them without having to worry about pushing anything by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,9 +25,12 @@ vendor/
 ############
 
 .claude/
-CLAUDE.md
 .clinerules/
+.codex/
 .cursor/
+.gemini/
+.windsurf/
+CLAUDE.md
 GEMINI.md
 
 ############
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ vendor/
 ############
 
 .claude/
+.clinerules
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -25,9 +25,13 @@ vendor/
 ############
 
 .claude/
-CLAUDE.md
+.clinerules
 .clinerules/
+.codex/
 .cursor/
+.gemini/
+.windsurf/
```
mukeshpanchal27, line 33: Should be in `IDEs` group

mukeshpanchal27 (approved):

.gitignore
felixarntz: Not necessarily. Admittedly there's overlap, Windsurf and Cursor can be considered both an IDE and an AI assistant tool for instance. I don't think it matters too much :)

JasonTheAdams (approved):

.gitignore
JasonTheAdams: I agree that it's best where it is on the grounds that the configurations going in these directories aren't related to the IDE, but are for the agents. 👍

PR #105, opened 2025-10-08
feature/support-request-options

### LINKED ISSUE #99: Support defining request options
There is currently no way to specify request options in this SDK, which is a severe limitation when it comes to certain use-cases. To clarify with what I mean with options: It's request transport behaviors that cannot be expressed by the request URL, method, headers, or body, such as `timeout` or how many `redirections` to follow (if any).

We need to allow passing such options (the two above are crucial examples, though there may be more). In any case, we can start simple, with those two, and add more as a need comes up. The key part is figuring out _how_ to architecturally enable this.

I believe we have two reasonable options:
* Either allow specifying options in our `Request` DTO: This would nicely centralize responsibility in the DTO we already use for everything.
* Or we allow passing options to the `HttpTransporter::send` method, via an optional second parameter: This would mean there are two touch points for how to make a request (the `Request` DTO itself and this parameter), which is not as neat. But on the other hand, one may argue that these request options should be centrally set, or we could even allow providing simple callbacks that can dynamically determine the value to use based on the current `Request` that will be made.

In any case, I believe we should introduce a new DTO `RequestOptions` (which would either be used as a `Request` property or a `HttpTransporter::send` parameter) - let's avoid the WordPress `array $args` pattern and keep things type-safe :)

For reference, this is the shape of our current `Request` DTO:
```
RequestArrayShape array{
    method: string,
    uri: string,
    headers: array<string, list<string>>,
    body?: string|null
}
 ```

The last thing to check that is crucial and may influence our decision is how this is handled by the relevant PSR implementations, including Guzzle. We'll want to make sure our implementation is interoperable with that.

### PR DESCRIPTION
**Pull request by Ref34t**
## Expand HTTP Request Controls

Closes #99 – Introduces transport-level request option support and richer HTTP error reporting across the PHP AI Client SDK.

## Overview

Adds a first-class RequestOptions DTO that flows from fluent builder usage down to PSR-7 conversions, so callers can specify timeouts and redirect policies per request. While wiring that through the stack, the transporter now wraps PSR-18 network failures in our exception hierarchy, and ResponseUtil translates non-2xx status codes into semantic exceptions with contextual messages. Documentation highlights the new configuration entry point to keep implementers aligned.

### Request Configuration

- RequestOptions DTO with validation, immutability helpers, JSON schema definition, and defaults factory.
- Request DTO gains options support, serialization, deserialization, and fromPsrRequest for transporter error wrapping.
- Example usage in README.md so developers discover timeout and redirect controls quickly.

### Tests

 - Comprehensive PHPUnit coverage for RequestOptions (constructors, validation paths, mutators, serialization, schema).
- Updated ResponseUtilTest to assert new exception types, status-code mappings, and extracted
    error details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,22 @@ $imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in t
 
 See the [`PromptBuilder` class](https://github.com/WordPress/php-ai-client/blob/trunk/src/Builders/PromptBuilder.php) and its public methods for all the ways you can configure the prompt.
 
+### Configuring request options
+
+You can configure HTTP transport options like timeout and maximum redirects using the `RequestOptions` DTO:
+
+```php
+use WordPress\AiClient\Providers\Http\DTO\RequestOptions;
+
+// Set custom timeout for long-running requests
+$options = new RequestOptions(120, 10);
+
+// Or use defaults and modify
+$options = RequestOptions::defaults()->withTimeout(60);
+```
+
+For implementation ideas in different environments (WordPress, Guzzle, cURL), check the transporter-specific examples in the SDK source and tests.
+
 **More documentation is coming soon.**
 
 ## Further reading
--- a/src/Providers/Http/DTO/Request.php
+++ b/src/Providers/Http/DTO/Request.php
@@ -23,7 +23,8 @@ use WordPress\AiClient\Providers\Http\Enums\HttpMethodEnum;
  *     method: string,
  *     uri: string,
  *     headers: array<string, list<string>>,
- *     body?: string|null
+ *     body?: string|null,
+ *     options?: array<string, mixed>
  * }
  *
  * @extends AbstractDataTransferObject<RequestArrayShape>
@@ -34,6 +35,7 @@ class Request extends AbstractDataTransferObject
     public const KEY_URI = 'uri';
     public const KEY_HEADERS = 'headers';
     public const KEY_BODY = 'body';
+    public const KEY_OPTIONS = 'options';
 
     /**
      * @var HttpMethodEnum The HTTP method.
@@ -61,6 +63,11 @@ class Request extends AbstractDataTransferObject
     protected ?string $body = null;
 
     /**
+     * @var RequestOptions|null The request options.
+     */
+    protected ?RequestOptions $options = null;
+
+    /**
      * Constructor.
      *
      * @since 0.1.0
@@ -69,11 +76,17 @@ class Request extends AbstractDataTransferObject
      * @param string $uri The request URI.
      * @param array<string, string|list<string>> $headers The request headers.
      * @param string|array<string, mixed>|null $data The request data.
+     * @param RequestOptions|null $options The request options.
      *
      * @throws InvalidArgumentException If the URI is empty.
      */
-    public function __construct(HttpMethodEnum $method, string $uri, array $headers = [], $data = null)
-    {
+    public function __construct(
+        HttpMethodEnum $method,
+        string $uri,
+        array $headers = [],
+        $data = null,
+        ?RequestOptions $options = null
+    ) {
         if (empty($uri)) {
             throw new InvalidArgumentException('URI cannot be empty.');
         }
@@ -81,6 +94,7 @@ class Request extends AbstractDataTransferObject
         $this->method = $method;
         $this->uri = $uri;
         $this->headers = new HeadersCollection($headers);
+        $this->options = $options;
 
         // Separate data and body based on type
         if (is_string($data)) {
@@ -282,6 +296,33 @@ class Request extends AbstractDataTransferObject
     }
 
     /**
+     * Gets the request options.
+     *
+     * @since n.e.x.t
+     *
+     * @return RequestOptions|null The request options or null if not set.
+     */
+    public function getOptions(): ?RequestOptions
+    {
+        return $this->options;
+    }
+
+    /**
+     * Returns a new instance with the specified options.
+     *
+     * @since n.e.x.t
+     *
+     * @param RequestOptions|null $options The request options.
+     * @return self A new instance with the options.
+     */
+    public function withOptions(?RequestOptions $options): self
+    {
+        $new = clone $this;
+        $new->options = $options;
+        return $new;
+    }
+
+    /**
      * {@inheritDoc}
      *
      * @since 0.1.0
@@ -311,6 +352,10 @@ class Request extends AbstractDataTransferObject
                     'type' => ['string'],
                     'description' => 'The request body.',
                 ],
+                self::KEY_OPTIONS => [
+                    'type' => 'object',
+                    'description' => 'The request options.',
+                ],
             ],
             'required' => [self::KEY_METHOD, self::KEY_URI, self::KEY_HEADERS],
         ];
@@ -337,6 +382,11 @@ class Request extends AbstractDataTransferObject
             $array[self::KEY_BODY] = $body;
         }
 
+        // Include options if present
+        if ($this->options !== null) {
+            $array[self::KEY_OPTIONS] = $this->options->toArray();
+        }
+
         return $array;
     }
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -78,6 +78,22 @@ $imageFile = AiClient::prompt('Generate an illustration of the PHP elephant in t
...
+
+// Set custom timeout for long-running requests
+$options = new RequestOptions(120, 10);
+
+// Or use defaults and modify
+$options = RequestOptions::defaults()->withTimeout(60);
+```
+
+For implementation ideas in different environments (WordPress, Guzzle, cURL), check the transporter-specific examples in the SDK source and tests.
+
```
felixarntz, line 96: This seems far too specific to document in the overall `README.md`. Let's not add any documentation for this yet, because we're still missing an overall structure for documentation.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Maximum allowed timeout in seconds (1 hour).
+     */
+    public const MAX_TIMEOUT = 3600;
+
+    /**
+     * Maximum allowed redirects.
+     */
+    public const MAX_REDIRECTS = 100;
```
felixarntz, line 48: What is the rationale for these values? They seem a bit arbitrary to me. I'd say let's not validate this, it's on the developer to choose reasonable values. Validating that every value is a positive integer should be sufficient.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Default timeout in seconds.
+     */
+    public const DEFAULT_TIMEOUT = 30;
+
+    /**
+     * Default maximum number of redirects.
+     */
+    public const DEFAULT_MAX_REDIRECTS = 5;
```
felixarntz, line 38: I don't think we should set any defaults. Every request options should either have a concrete value, or be `null` (the default), in which case it should be ignored. All of these options are optional by definition, and if not specified, whatever the default HTTP client uses will be applied.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param int|null $timeout The request timeout in seconds.
+     * @param int|null $max_redirects The maximum number of redirects to follow.
+     *
+     * @throws InvalidArgumentException If timeout or max_redirects is invalid.
+     */
+    public function __construct(?int $timeout = null, ?int $max_redirects = null)
```
felixarntz, line 70: Instead of a constructor with named arguments, can you adjust this class to be more in line with how https://github.com/WordPress/php-ai-client/blob/trunk/src/Providers/Models/DTO/ModelConfig.php works? We have to anticipate more and more request options in the future (think e.g. `user_agent`, `streaming`, ...), and an ordered list of parameters isn't suitable for that.

Let's for now keep only supporting `timeout` and `max_redirects` like you have it, but we'll need to architect this in a way that it could easily support 20 different options too, e.g. with setter and getter methods.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+                    'Max redirects must be between 0 and %d.',
+                    self::MAX_REDIRECTS
+                )
+            );
+        }
+
+        $new = clone $this;
+        $new->max_redirects = $maxRedirects;
+        return $new;
+    }
```
felixarntz, line 168: See above, let's use setter and getter methods, in line with `ModelConfig`.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+     * Checks if any options are set.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if any options are set, false otherwise.
+     */
+    public function hasOptions(): bool
+    {
+        return $this->timeout !== null || $this->max_redirects !== null;
+    }
```
felixarntz, line 180: I don't think this method is necessary. We have the `toArray` method already, and it's trivial to do a `!$requestOptions->toArray()` check.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @return self A new RequestOptions instance with default values.
+     */
+    public static function defaults(): self
+    {
+        return new self(
+            self::DEFAULT_TIMEOUT,
+            self::DEFAULT_MAX_REDIRECTS
+        );
+    }
```
felixarntz, line 256: See above, this shouldn't be needed.

src/Providers/Http/Exception/ClientException.php
```diff
@@ -55,9 +56,10 @@ public function getRequest(): Request
      * @since n.e.x.t
      *
      * @param Response $response The HTTP response that failed.
+     * @param Request|null $request The originating HTTP request, if available.
      * @return self
      */
-    public static function fromClientErrorResponse(Response $response): self
+    public static function fromClientErrorResponse(Response $response, ?Request $request = null): self
```
felixarntz, line 62: The changes in this file seem unrelated to the purpose of this PR? Can you clarify why you implemented them? Otherwise I think they should be reverted.

The `RuntimeException` fix above is a great catch and simple enough to include, but the other changes here haven't been discussed and aren't related to the problem.

src/Providers/Http/HttpTransporter.php
```diff
@@ -71,6 +71,8 @@ public function send(Request $request): Response
     {
         $psr7Request = $this->convertToPsr7Request($request);
 
+        // PSR-18 clients ignore per-request RequestOptions; custom transporters must apply them.
```
felixarntz, line 74: See https://github.com/WordPress/php-ai-client/issues/99#issuecomment-3383594184 for more context on how we should support this.

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -29,12 +31,13 @@ class ResponseUtil
...
      * @param Response $response The HTTP response to check.
+     * @param Request|null $request The originating HTTP request, if available.
      * @throws RedirectException If the response indicates a redirect (3xx).
      * @throws ClientException If the response indicates a client error (4xx).
      * @throws ServerException If the response indicates a server error (5xx).
-     * @throws \RuntimeException If the response has an invalid status code.
+     * @throws RuntimeException If the response has an invalid status code.
      */
-    public static function throwIfNotSuccessful(Response $response): void
+    public static function throwIfNotSuccessful(Response $response, ?Request $request = null): void
```
felixarntz, line 40: See my above comment, this seems unrelated? Let's try to focus individual PRs on individual problems, otherwise code reviewing becomes difficult and it can hold up work that otherwise would already be mergeable :)

src/Providers/Http/Util/ResponseUtil.php
```diff
@@ -49,15 +52,15 @@ public static function throwIfNotSuccessful(Response $response): void
...
+            throw ClientException::fromClientErrorResponse($response, $request);
         }
 
         // 5xx Server Errors
         if ($statusCode >= 500 && $statusCode < 600) {
             throw ServerException::fromServerErrorResponse($response);
         }
 
-        throw new \RuntimeException(
+        throw new RuntimeException(
```
felixarntz, line 63: Great catch! This one's definitely worth keeping, a simple enough fix :)

tests/unit/Providers/Http/Util/ResponseUtilTest.php
```diff
@@ -61,7 +61,9 @@ public function testThrowIfNotSuccessfulThrowsClientExceptionFor400BadRequest():
 
         $this->expectException(ClientException::class);
         $this->expectExceptionCode(400);
-        $this->expectExceptionMessage('Bad Request (400)');
+        $this->expectExceptionMessageMatches(
```
felixarntz, line 64: See above, the changes in this test file should be reverted, the `ResponseUtil` class doesn't need any updates in regards to supporting request options.

src/Providers/Http/DTO/Request.php
```diff
@@ -69,18 +76,25 @@ class Request extends AbstractDataTransferObject
...
      * @throws InvalidArgumentException If the URI is empty.
      */
-    public function __construct(HttpMethodEnum $method, string $uri, array $headers = [], $data = null)
-    {
+    public function __construct(
+        HttpMethodEnum $method,
+        string $uri,
+        array $headers = [],
+        $data = null,
+        ?RequestOptions $options = null
```
felixarntz, line 88: There's some open discussion in https://github.com/WordPress/php-ai-client/issues/99#issuecomment-3387611431 on whether this is the preferred approach, so please feel free to share your thoughts there. I'm only flagging this here - maybe what you have here is what we should be doing, but it's still undecided.

felixarntz (commented):
@Ref34t Thank you for working on this! It's a bit premature given the ongoing discussion, but a large portion of your PR is on the right track, which is great.

JasonTheAdams:
I think we can close this in favor of #109. I apologize, I didn't realize this PR exists until I was mostly through the other one. It addresses the feedback given here and also adds the Guzzle support we talked about later.

JasonTheAdams:
Closing in favor of #109

PR #109, opened 2025-10-15
Adds support for HTTP Transporter Options

### LINKED ISSUE #99: Support defining request options

JasonTheAdams:
Resolves #99 

This updates the HTTP stack so we can pass request-specific transport options end to end. We now have a nullable RequestOptions DTO that threads through `Request`, `ModelConfig`, and the transporter contracts, so the `HttpTransporter` understands both our option-aware clients and Guzzle-style clients that expose a send($request, $options) signature. The change also touches supporting enums/mocks and adds unit coverage to lock in the new behaviour.

Alongside that, the new `ClientWithOptionsInterface` offers a lightweight extension point for PSR-18 clients that already know how to accept option bags. Libraries that wrap Guzzle or provide their own richer transport (like our WP AI Client repo) can implement this interface to let the transporter hand them a `RequestOptions` instance directly, avoiding reflection while keeping the core client agnostic.


JasonTheAdams:
This is ready for you, @felixarntz! 

One thing I went back and forth on: Are DTOs supposed to be immutable? At first I thought yes, but only because I thought I remembered it, but then I realized that `ModelConfig` has setters. Thinking they're immutable is why I used `withX` immutable methods in `RequestOptions`. If that's not the case I'll switch it to setters to keep things simple.

felixarntz:
> One thing I went back and forth on: Are DTOs supposed to be immutable? At first I thought yes, but only because I thought I remembered it, but then I realized that `ModelConfig` has setters. Thinking they're immutable is why I used `withX` immutable methods in `RequestOptions`. If that's not the case I'll switch it to setters to keep things simple.

Yes, I think they generally should be. The `ModelConfig` only has setters because we didn't want to use a giant `$args` array for the constructor, but that's still a less than ideal workaround - ideally it should be immutable too. I think for now we should model `RequestOptions` like `ModelConfig`, just to be able to proceed quickly. But we should assume it's supposed to be immutable, still. We can discuss in a separate issue how we want to make these two classes technically immutable (which are only different from the others because they support _so many_ potential keys).

src/Providers/Http/DTO/Request.php
```diff
@@ -281,6 +297,104 @@ public function getData(): ?array
...
+     * @return RequestOptions The ensured request options instance.
+     */
+    private function ensureOptions(): RequestOptions
+    {
+        if ($this->options === null) {
+            $this->options = new RequestOptions();
+        }
+
+        return $this->options;
+    }
```
felixarntz: I think we should get rid of these, for the following reasons:
* The DTOs should be immutable.
* If we wanted setters, they should still be only on `RequestOptions` itself.
* `withOptions` is a sufficient way to achieve this goal.

Setting request options is a lower-level API. So I would say we shouldn't add a bunch of convenience wrappers around it. If we start doing this here, there are other way more first-class APIs where we would need to consider this, but I think it's better to keep the API surface centered to its root.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,351 @@
...
+        $clone = clone $this;
+
+        if ($this->allowRedirects === true) {
+            $clone->maxRedirects = $maxRedirects;
+        } else {
+            $clone->maxRedirects = null;
+        }
+
+        return $clone;
+    }
```
felixarntz: See my reply to your comment, I think for now we would better align this class to follow the current architecture of `ModelConfig`. It's a "bag" / "dump" object for all kinds of request options, so it's similar to `ModelConfig`, which is for all kinds of model options.

While the `ModelConfig` setup probably isn't ideal because it's mutable, I think that's fine for now. Maybe it's even a deliberate decision to have these "bag" classes mutable, unlike all other DTOs. Either way, I think we should follow that path here and reconsider separately in another issue.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,351 @@
...
+     * @param int|null $maxRedirects Maximum redirects to follow when enabled.
+     *
+     * @throws InvalidArgumentException When timeout values or redirect limits are invalid.
+     */
+    public function __construct(
+        ?float $timeout = null,
+        ?float $connectTimeout = null,
+        ?bool $allowRedirects = null,
+        ?int $maxRedirects = null
+    ) {
```
felixarntz: See my other comment, let's not do this, but instead follow the `ModelConfig` approach.

Having named parameters here is not very future proof because there could easily be 10-15 keys in here at some point. These 4 are a good starting point, and we don't need to go further, but there are many more potential request options to support.

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,351 @@
...
+     * Validates redirect configuration.
+     *
+     * @since n.e.x.t
+     *
+     * @param bool|null $allowRedirects Whether redirects are enabled.
+     * @param int|null $maxRedirects Maximum redirects.
+     *
+     * @throws InvalidArgumentException When redirect count is invalid.
+     */
+    private function validateRedirectLimit(?bool $allowRedirects, ?int $maxRedirects): void
```
felixarntz: Do we really want to have `allowRedirects` _and_ `maxRedirects`? I never understood why WordPress has both. IMO setting `maxRedirects` to `0` is a perfectly understandable way of saying "do not allow redirects".

src/Providers/Http/HttpTransporter.php
```diff
@@ -66,13 +68,21 @@ public function __construct(
      * {@inheritDoc}
      *
      * @since 0.1.0
+     * @since n.e.x.t Added optional RequestOptions parameter and ClientWithOptions support.
      */
-    public function send(Request $request): Response
+    public function send(Request $request, ?RequestOptions $options = null): Response
     {
         $psr7Request = $this->convertToPsr7Request($request);
+        $options = $options ?? $request->getOptions();
```
felixarntz: We shouldn't do a general either or here IMO, but if both are set, we should merge them.

A key decision here is the precedence, and I'm not sure about that:
* Do we prioritize the given parameter because it's more explicit at the time of this call (could be seen as a way to override the regular request)?
* Or do we prioritize the instance within the `Request` DTO because that's defined more closely to the source of truth (the request defines what kind of content is requested, so things like timeout may directly depend on those considerations)?

Curious what you think.

Either way, I also think we should document the decision above this line here with a brief inline comment.

src/Providers/Http/HttpTransporter.php
```diff
@@ -91,6 +101,122 @@ public function send(Request $request): Response
...
 
+    /**
+     * Determines if the underlying client matches the Guzzle client shape.
+     *
+     * @since n.e.x.t
+     *
+     * @param ClientInterface $client The HTTP client instance.
+     * @return bool True when the client exposes Guzzle's send signature.
+     */
+    private function isGuzzleClient(ClientInterface $client): bool
```
felixarntz, line 172: Love this! It's a really creative and pragmatic solution which, while hacky, is exactly what we need to make for a good DX for most, without forcing Guzzle on everyone. 🙌 

And with these methods being private, there's no risk. If we think this is bad later, we could simply remove it.

src/Providers/Models/DTO/ModelConfig.php
```diff
@@ -73,6 +76,7 @@ class ModelConfig extends AbstractDataTransferObject
     public const KEY_OUTPUT_MEDIA_ASPECT_RATIO = 'outputMediaAspectRatio';
     public const KEY_OUTPUT_SPEECH_VOICE = 'outputSpeechVoice';
     public const KEY_CUSTOM_OPTIONS = 'customOptions';
+    public const KEY_REQUEST_OPTIONS = 'requestOptions';
```
felixarntz: I don't think we should store `RequestOptions` in `ModelConfig`. The two are containers for distinct kinds of options, and `RequestOptions` should only be associated with the request itself.

They are also fundamentally different - one is directly API related, while the other one is generic and would be useful even if were requesting random HTML documents.

felixarntz (changes requested):
@JasonTheAdams Overall this looks great!

A few points of feedback and a few questions that I think need to be addressed, but it's near-perfect, pending a few simplifications.

Note that there was already a PR for this in #105, so we'll need to see how we want to proceed.

src/Providers/Http/DTO/Request.php
JasonTheAdams: Resolved in 88b0075

src/Providers/Http/DTO/RequestOptions.php
JasonTheAdams: Yeah, you're right. It feels like it's philosophically right and pragmatically wrong. Resolved in ba6cf36

src/Providers/Http/DTO/RequestOptions.php
JasonTheAdams: Agreed! Resolved in ba6cf36

src/Providers/Http/DTO/RequestOptions.php
JasonTheAdams: Yeah, I agree. Resolved in 6551352

src/Providers/Http/HttpTransporter.php
JasonTheAdams: I went with the parameter taking precedence since it's optional and last-minute. I documented it.

Resolved in 39a5d82

src/Providers/Models/DTO/ModelConfig.php
JasonTheAdams: Oops, that was a mistake. I'll blame Codex. 😁  Good catch! Resolved in 90e0155

src/Providers/Http/DTO/RequestOptions.php
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * @var float|null Maximum duration in seconds to wait for the full response.
+     */
+    protected ?float $timeout = null;
+
+    /**
+     * @var float|null Maximum duration in seconds to wait for the initial connection.
+     */
+    protected ?float $connectTimeout = null;
```
felixarntz, line 39: Are these values usually handled as `float`? I totally see why that could be reasonable, but at the same I don't recall ever seeing a non-integer timeout value anywhere. Given we're not talking about very small durations for either of these, maybe it seems a bit excessive to support "I want to wait 5.5 seconds".

Curious what you think.

src/Providers/Http/HttpTransporter.php
```diff
@@ -66,13 +68,23 @@ public function __construct(
...
         $psr7Request = $this->convertToPsr7Request($request);
 
+        // Merge request options with parameter options, with parameter options taking precedence
+        $mergedOptions = $this->mergeOptions($request->getOptions(), $options);
+
         try {
-            $psr7Response = $this->client->sendRequest($psr7Request);
+            if ($this->client instanceof ClientWithOptionsInterface) {
+                $psr7Response = $this->client->sendRequestWithOptions($psr7Request, $mergedOptions);
+            } elseif ($this->isGuzzleClient($this->client)) {
```
felixarntz: Nit-pick: We could change these two clauses to only be relevant if `$mergedOptions !== null`. Not a must have by any means though.

src/Providers/Http/HttpTransporter.php
```diff
@@ -91,6 +103,179 @@ public function send(Request $request): Response
...
+
+        if ($parameterOptions->getConnectTimeout() !== null) {
+            $merged->setConnectTimeout($parameterOptions->getConnectTimeout());
+        }
+
+        if ($parameterOptions->getMaxRedirects() !== null) {
+            $merged->setMaxRedirects($parameterOptions->getMaxRedirects());
+        }
+
+        return $merged;
```
felixarntz, line 161: Wouldn't it be a lot more straightforward to call `toArray` on both instances, merge them in the right precedence order, and then pass the result back into `fromArray`?

felixarntz (approved):
@JasonTheAdams Looks great - only minor notes now, nothing blocking.

src/Providers/Http/DTO/RequestOptions.php
JasonTheAdams: It's a float in `WP_Http::request()`, so I see no reason to limit it. If someone needs that level of precision, then power to them. Hahah! If the subsequent system only supports ints, then it can truncate it without much concern.

src/Providers/Http/HttpTransporter.php
JasonTheAdams: Yes, yes it would be. Hahah! Resolved in 4143b11

src/Providers/Http/HttpTransporter.php
JasonTheAdams: Sure! Makes sense. Resolved in 0da66e6

JasonTheAdams: I also removed nullable from the interfaces since we're not guaranteeing it will only be called if the options are present in b976abd

PR #110, opened 2025-10-15
Adds PromptBuilder preferred model support

### LINKED ISSUE #103: Allow specifying preferred / suggested models for a specific prompt / task
This has been raised quite a bit in the past few weeks/months, but we never actually got to implementing it: We need to allow developers to (optionally!) specify a model preference list. I believe this is a critical feature for the following reasons:

* We want to encourage developers to use the SDK in a model agnostic way, i.e. not force usage of a specific model.
* The reality though is that many experienced developers will have preferences for which model(s) to use for a given process in their code, either simply as a personal preference or, ideally, due to having run evals and identified the strongest model(s) for that process.
* Today, these developers can only accomplish their goal by force-choosing their one preferred model, which is an understandable choice from their end, but actively goes against what we want to encourage.

**Allowing to express model preferences/suggestions without _requiring_ that one of these models to be used addresses these problems.**

Users still benefit from the best model(s) for the specific task when they're available, but if on a site that model's provider is not configured, they're not locked out from using the feature.

## Technical design

High-level suggestion:
* We need to have a method on `PromptBuilder` that allows to provide a list of model IDs, or tuples of provider ID plus model ID.

More concretely:
* Naming is a hard question, as always:
    * `usingPreferredModels` seems wrong, as that sounds too _strong_ in that always one of these models will be used, which is not the case.
    * `usingModelPreferences` sounds better, but "preferences" is somehow ambiguous as it's often used as a synonym for "options" or "config", which is not what this is.
    * `usingModelSuggestions` might be a good compromise. Maybe there's another name that's even more suitable though :)
* I think we should allow passing an array/list, where each entry is:
    * either a model ID string (e.g. `'gemini-flash-2.5'`)
    * or a tuple of provider ID string plus model ID string (e.g. `['google', 'gemini-flash-2.5']`)
    * or a `ModelInterface` instance, used simply to extract provider ID and model ID (e.g. `Google::model('gemini-flash-2.5')`)
* Under the hood we could introduce an SDK internal class like `ModelPreferences` or `ModelSuggestions` (pending naming decisions above) that receives the array and then has methods to filter a given `list<ModelMetadata>` or `list<ProviderModelsMetadata>`, or extract the specific "most preferred/suggested" model out of such a list.

JasonTheAdams:
Resolved #103 

This adds the `PromptBuilder::usingModelPreference()` method which allows the user to provide a list of models _it would like_ to use. If no model is present, then it falls back on the discovery system.

### DIFF
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -226,7 +226,8 @@ class PromptBuilder
      *
      * @since n.e.x.t
      *
-     * @param string|ModelInterface|array{0:string,1:string} ...$preferredModels The preferred models as model IDs, model instances, or [provider ID, model ID] tuples.
+     * @param string|ModelInterface|array{0:string,1:string} ...$preferredModels The preferred models as model IDs,
+     * model instances, or [provider ID, model ID] tuples.
      * @return self
      *
      * @throws InvalidArgumentException When a preferred model has an invalid type or identifier.
@@ -1127,9 +1128,37 @@ class PromptBuilder
             return $this->model;
         }
 
-        /** @var list<array{0:string,1:ModelMetadata}> $candidateModels */
-        $candidateModels = [];
+        // Retrieve the models which satisfy the requirements.
+        $candidateModels = $this->getCandidateModels($requirements);
 
+        // Find the model which matches the preference, if any.
+        $selectedModel = $this->findModelByPreference($candidateModels);
+
+        if ($selectedModel === null) {
+            // No preference matched; fall back to the first candidate discovered.
+            [$providerId, $modelMetadata] = $candidateModels[0];
+            $selectedModel = $this->registry->getProviderModel(
+                $providerId,
+                $modelMetadata->getId(),
+                $this->modelConfig
+            );
+        }
+
+        return $selectedModel;
+    }
+
+    /**
+     * Builds the list of candidate provider/model combinations that satisfy the requirements.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModelRequirements $requirements The requirements derived from the prompt.
+     * @return list<array{0:string,1:ModelMetadata}> The candidate provider/model tuples.
+     *
+     * @throws InvalidArgumentException If no suitable models are found.
+     */
+    private function getCandidateModels(ModelRequirements $requirements): array
+    {
         if ($this->providerIdOrClassName === null) {
             // No provider locked in, gather all models across providers that meet requirements.
             $providerModelsMetadata = $this->registry->findModelsMetadataForSupport($requirements);
@@ -1149,51 +1178,60 @@ class PromptBuilder
                 );
             }
 
+            $candidateModels = [];
             foreach ($providerModelsMetadata as $providerModels) {
                 $providerId = $providerModels->getProvider()->getId();
                 foreach ($providerModels->getModels() as $modelMetadata) {
                     $candidateModels[] = [$providerId, $modelMetadata];
                 }
             }
-        } else {
-            // Provider forced, only consider models from that provider.
-            $modelsMetadata = $this->registry->findProviderModelsMetadataForSupport(
-                $this->providerIdOrClassName,
-                $requirements
-            );
 
-            if (empty($modelsMetadata)) {
-                throw new InvalidArgumentException(
-                    sprintf(
-                        'No models found for %s that support the required capabilities and options for this prompt. ' .
-                        'Required capabilities: %s. Required options: %s',
-                        $this->providerIdOrClassName,
-                        implode(', ', array_map(function ($cap) {
-                            return $cap->value;
-                        }, $requirements->getRequiredCapabilities())),
-                        implode(', ', array_map(function ($opt) {
-                            return $opt->getName()->value . '=' . json_encode($opt->getValue());
-                        }, $requirements->getRequiredOptions()))
-                    )
-                );
-            }
-
-            foreach ($modelsMetadata as $modelMetadata) {
-                $candidateModels[] = [$this->providerIdOrClassName, $modelMetadata];
-            }
+            return $candidateModels;
         }
 
-        $selectedModel = null;
-        $selectedPreferenceIndex = null;
+        // Provider set, only consider models from that provider.
+        $modelsMetadata = $this->registry->findProviderModelsMetadataForSupport(
+            $this->providerIdOrClassName,
+            $requirements
+        );
 
+        if (empty($modelsMetadata)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'No models found for %s that support the required capabilities and options for this prompt. ' .
+                    'Required capabilities: %s. Required options: %s',
+                    $this->providerIdOrClassName,
+                    implode(', ', array_map(function ($cap) {
+                        return $cap->value;
+                    }, $requirements->getRequiredCapabilities())),
+                    implode(', ', array_map(function ($opt) {
+                        return $opt->getName()->value . '=' . json_encode($opt->getValue());
+                    }, $requirements->getRequiredOptions()))
+                )
+            );
+        }
+
+        $candidateModels = [];
+        foreach ($modelsMetadata as $modelMetadata) {
+            $candidateModels[] = [$this->providerIdOrClassName, $modelMetadata];
+        }
+
+        return $candidateModels;
+    }
+
+    /**
+     * Finds a preferred model among the given candidates.
+     *
+     * @since n.e.x.t
+     *
+     * @param list<array{0:string,1:ModelMetadata}> $candidateModels Candidate provider/model tuples.
+     * @return ModelInterface|null The model matching the highest-priority preference, or null if none.
+     */
+    private function findModelByPreference(array $candidateModels): ?ModelInterface
+    {
         foreach ($candidateModels as [$providerId, $modelMetadata]) {
             foreach ($this->modelPreference as $preferenceIndex => $preferredModel) {
-                if ($selectedPreferenceIndex !== null && $preferenceIndex >= $selectedPreferenceIndex) {
-                    continue;
-                }
-
                 if (is_array($preferredModel)) {
-                    // [provider ID, model ID] tuple
                     [$preferredProviderId, $preferredModelId] = $preferredModel;
                     if (
                         $providerId !== $preferredProviderId ||
@@ -1202,14 +1240,12 @@ class PromptBuilder
                         continue;
                     }
 
-                    $selectedModel = $this->registry->getProviderModel(
+                    return $this->registry->getProviderModel(
                         $providerId,
                         $modelMetadata->getId(),
                         $this->modelConfig
                     );
-                    $selectedPreferenceIndex = $preferenceIndex;
                 } elseif ($preferredModel instanceof ModelInterface) {
-                    // Model instance
                     $preferredProviderId = $preferredModel->providerMetadata()->getId();
                     $preferredModelId = $preferredModel->metadata()->getId();
 
@@ -1220,44 +1256,25 @@ class PromptBuilder
                         continue;
                     }
 
-                    $selectedModel = $preferredModel;
-                    $selectedPreferenceIndex = $preferenceIndex;
+                    $preferredModel->setConfig($this->modelConfig);
+                    $this->registry->bindModelDependencies($preferredModel);
+
+                    return $preferredModel;
                 } else {
-                    // Model ID
                     if ($modelMetadata->getId() !== $preferredModel) {
                         continue;
                     }
 
-                    $selectedModel = $this->registry->getProviderModel(
+                    return $this->registry->getProviderModel(
                         $providerId,
                         $modelMetadata->getId(),
                         $this->modelConfig
                     );
-                    $selectedPreferenceIndex = $preferenceIndex;
                 }
-
-                if ($selectedPreferenceIndex === 0) {
-                    break 2;
-                }
-
-                break;
             }
         }
 
-        if ($selectedModel === null) {
-            // No preference matched; fall back to the first candidate discovered.
-            [$providerId, $modelMetadata] = $candidateModels[0];
-            $selectedModel = $this->registry->getProviderModel(
-                $providerId,
-                $modelMetadata->getId(),
-                $this->modelConfig
-            );
-        }
-
-        $selectedModel->setConfig($this->modelConfig);
-        $this->registry->bindModelDependencies($selectedModel);
-
-        return $selectedModel;
+        return null;
     }
 
     /**
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -61,7 +61,7 @@ class PromptBuilder
     protected ?ModelInterface $model = null;
 
     /**
-     * @var list<ModelInterface|string> Preferred models to evaluate in order.
+     * @var list<ModelInterface|string|array{0:string,1:string}> Preferred models to evaluate in order.
      */
     protected array $modelPreference = [];
 
@@ -226,7 +226,7 @@ class PromptBuilder
      *
      * @since n.e.x.t
      *
-     * @param mixed ...$preferredModels The preferred models as model IDs or instances.
+     * @param string|ModelInterface|array{0:string,1:string} ...$preferredModels The preferred models as model IDs, model instances, or [provider ID, model ID] tuples.
      * @return self
      *
      * @throws InvalidArgumentException When a preferred model has an invalid type or identifier.
@@ -240,6 +240,37 @@ class PromptBuilder
         $normalizedModels = [];
 
         foreach ($preferredModels as $preferredModel) {
+            if (is_array($preferredModel)) {
+                if (!array_is_list($preferredModel) || count($preferredModel) !== 2) {
+                    throw new InvalidArgumentException(
+                        'Model preference tuple must contain provider ID and model identifier.'
+                    );
+                }
+
+                [$providerId, $modelIdentifier] = $preferredModel;
+
+                if (!is_string($providerId)) {
+                    throw new InvalidArgumentException('Model preference provider identifiers cannot be empty.');
+                }
+                $providerId = trim($providerId);
+                if ($providerId === '') {
+                    throw new InvalidArgumentException('Model preference provider identifiers cannot be empty.');
+                }
+
+                if (!is_string($modelIdentifier)) {
+                    throw new InvalidArgumentException(
+                        'Model preference tuple must contain provider ID and model identifier.'
+                    );
+                }
+                $modelIdentifier = trim($modelIdentifier);
+                if ($modelIdentifier === '') {
+                    throw new InvalidArgumentException('Model preference identifiers cannot be empty.');
+                }
+
+                $normalizedModels[] = [$providerId, $modelIdentifier];
+                continue;
+            }
+
             if ($preferredModel instanceof ModelInterface) {
                 $normalizedModels[] = $preferredModel;
                 continue;
@@ -255,7 +286,7 @@ class PromptBuilder
             }
 
             throw new InvalidArgumentException(
-                'Model preferences must be model identifiers or instances of ModelInterface.'
+                'Model preferences must be model identifiers, instances of ModelInterface, or provider/model tuples.'
             );
         }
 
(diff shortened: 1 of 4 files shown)


src/Builders/PromptBuilder.php
```diff
@@ -59,6 +59,11 @@ class PromptBuilder
      */
     protected ?ModelInterface $model = null;
 
+    /**
+     * @var list<ModelInterface|string> Preferred models to evaluate in order.
+     */
+    protected array $modelPreferences = [];
```
felixarntz: Can we use the singular name here too (`modelPreference`)? It's more in line with the method, and using the plural here re-raises my previous concern from the issue about that name - `modelPreferences` could easily be confused with `modelOptions` or `modelConfig` or `modelSettings` as a concept.

src/Builders/PromptBuilder.php
```diff
@@ -1040,11 +1122,13 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
-            $this->registry->bindModelDependencies($this->model);
-            return $this->model;
+        $explicitModel = $this->model ?? $this->getAvailablePreferredModel();
+
+        if ($explicitModel !== null) {
+            $explicitModel->setConfig($this->modelConfig);
+            $this->registry->bindModelDependencies($explicitModel);
+            $this->model = $explicitModel;
+            return $explicitModel;
         }
```
felixarntz: I think this is not the right sequential place for this logic:
* The model preference is meant to express: Which of all the suitable models do I want to use?
* This requires running the regular logic to find all suitable models (see below your code change here), and then, if a preference list is given, go over the preference list and as soon as one of the preferred models is found in the list of suitable models, return it.
* The way you're currently doing it, it doesn't check whether the models are suitable, instead it's an explicit "you must use one of these models" list, regardless of whether the model technically supports what's requested. This is how we handle it for the "explicit" `usingModel` method, but the model preference approach is intentionally _not_ explicit, it's just a _preference_.

src/Builders/PromptBuilder.php
```diff
@@ -1024,6 +1072,40 @@ protected function appendPartToMessages(MessagePart $part): void
...
     }
 
+    /**
+     * Gets the first preferred model that can be instantiated.
+     *
+     * @since n.e.x.t
+     *
+     * @return ModelInterface|null The instantiated preferred model, or null if none found.
+     */
+    private function getAvailablePreferredModel(): ?ModelInterface
```
felixarntz: See my other comment: Instead of having this method, we need a way to go through the model preference list and, as soon as one of the models is present in the suitable models list for the prompt, return it.

That logic needs to conditionally replace the current hard-coded assignment to use the first provider/model from the suitable providers/models list.

src/Builders/PromptBuilder.php
```diff
@@ -215,6 +220,49 @@ public function usingModel(ModelInterface $model): self
...
+
+        $normalizedPreferences = [];
+
+        foreach ($preferredModels as $preferredModel) {
+            if ($preferredModel instanceof ModelInterface) {
+                $normalizedPreferences[] = $preferredModel;
+                continue;
+            }
+
+            if (is_string($preferredModel)) {
```
felixarntz: In the issue we were also mentioning a third way we should support, a tuple of provider ID and model ID (i.e. an array with exactly 2 elements, both of which are strings). So that still needs to be added here. In terms of precision it's similar to providing a `ModelInterface`.

Since this method is supposed to normalize, I'd say we could easily normalize so that every entry is _either_ the model ID string, or that above tuple of provider ID and model ID: For any `ModelInterface` provided, we can simply call `providerMetadata()->getId()` and `metadata->getId()`, to get the tuple.

src/Builders/PromptBuilder.php
```diff
@@ -1024,6 +1072,40 @@ protected function appendPartToMessages(MessagePart $part): void
...
+    private function getAvailablePreferredModel(): ?ModelInterface
+    {
+        if ($this->modelPreferences === []) {
+            return null;
+        }
+
+        foreach ($this->modelPreferences as $preferredModel) {
+            if ($preferredModel instanceof ModelInterface) {
+                return $preferredModel;
+            }
```
felixarntz: I mostly covered this in the previous comment, but just an aside here: Providing a concrete model class is purely to `usingModelPreference` is purely for DX, this instance is not meant to actually be used. It's purely so that you can write `Google::model('gemini-2.5-pro')` if that's your preference:
* Totally optional.
* But familiar DX if you're using the Vercel AI SDK.
* And a tiny bit safer than hard-coding the provider ID like `['google', 'gemini-2.5-pro']`.

Still, the purpose is purely to specify provider ID and model ID, not to actually use that class instance.

felixarntz (changes requested):
@JasonTheAdams I think there's a bit of a misunderstanding on what a model preference should mean here. See my below feedback.

Probably something that can be quickly addressed though once clarified.

src/Builders/PromptBuilder.php
JasonTheAdams: You're absolutely right! I attempted a solution in ba25084

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in ba25084

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in ba25084

src/Builders/PromptBuilder.php
JasonTheAdams: Got it! Resolved in ba25084

src/Builders/PromptBuilder.php
```diff
@@ -1040,15 +1090,18 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
         if ($this->model !== null) {
+            // Explicit model was provided via usingModel(); just update config and bind dependencies.
             $this->model->setConfig($this->modelConfig);
             $this->registry->bindModelDependencies($this->model);
             return $this->model;
         }
 
-        // Find a suitable model based on requirements
+        /** @var list<array{0:string,1:ModelMetadata}> $candidateModels */
+        $candidateModels = [];
```
felixarntz: This is not very efficient. The model lists can be massive, so iterating over them for every model preference is slow. I'd suggest to use a map with keys as follows:
* The provider ID and model ID concatenated.
* The model ID on its own.

We can also include some sort of identifier (e.g. `providerModel::{provider ID}::{model ID}`) and `model::{model ID}`) to make sure there cannot be any conflicts.

This way it's super quick to look up the correct provider ID and `ModelMetadata`.

src/Builders/PromptBuilder.php
```diff
@@ -1091,16 +1148,77 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+                    continue;
+                }
+
+                $selectedModel = $this->registry->getProviderModel(
+                    $providerId,
+                    $modelMetadata->getId(),
+                    $this->modelConfig
+                );
+                break 2;
+            }
```
felixarntz: See above, we can avoid this costly loop by using a map to look up the model ID or the provider ID/model ID combination.

src/Builders/PromptBuilder.php
felixarntz: @JasonTheAdams I see you refactored the code since I left this comment, but just want to make sure you see it. Still relevant :)

JasonTheAdams: Good idea! Implemented in 722e4aa

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 722e4aa

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in [1e66a08](https://github.com/WordPress/php-ai-client/pull/110/commits/1e66a08e557158bba72a6f8b27b56eafae24eb66)

src/Builders/PromptBuilder.php
```diff
@@ -216,9 +223,74 @@ public function usingModel(ModelInterface $model): self
...
+        foreach ($preferredModels as $preferredModel) {
+            if (is_array($preferredModel)) {
+                // [model identifier, provider ID] tuple
+                if (!array_is_list($preferredModel) || count($preferredModel) !== 2) {
+                    throw new InvalidArgumentException(
+                        'Model preference tuple must contain model identifier and provider ID.'
+                    );
+                }
+
+                [$modelIdentifier, $providerId] = $preferredModel;
```
felixarntz, line 252: The issue proposed this in the opposite order, but I'm certainly open to this alternative order. Let's take a step back and think about it: What makes more sense?

* Model ID is the mandatory part, passing the provider ID is optional, so from that perspective this order makes sense.
* On the other hand, a provider is the overarching thing in the hierarchy, above the model, so from that perspective provider ID first makes more sense.

WDYT?

src/Builders/PromptBuilder.php
```diff
@@ -1066,41 +1169,130 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
-                            return $cap->value;
-                        }, $requirements->getRequiredCapabilities())),
-                        implode(', ', array_map(function ($opt) {
-                            return $opt->getName()->value . '=' . json_encode($opt->getValue());
-                        }, $requirements->getRequiredOptions()))
-                    )
-                );
+        $candidateModels = [];
+        foreach ($modelsMetadata as $modelMetadata) {
+            $candidateModels[] = [$this->providerIdOrClassName, $modelMetadata];
```
felixarntz: Since this can be an ID or a class name, we need to normalize this to always be the ID, in order for the lookup to work.

We already `ProviderRegistry::getProviderClassName($id)`. Maybe we should also have `ProviderRegistry::getProviderId($className)`, and that we could then use here if the value is a class name. That should be fairly straightforward to add.

src/Builders/PromptBuilder.php
```diff
@@ -1040,15 +1112,46 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+        return $selectedModel;
+    }
+
+    /**
+     * Builds the list of candidate provider/model combinations that satisfy the requirements.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModelRequirements $requirements The requirements derived from the prompt.
+     * @return list<array{0:string,1:ModelMetadata}> The candidate provider/model tuples.
```
felixarntz: Worth clarifying:
```suggestion
     * @return list<array{0:string,1:ModelMetadata}> The candidate provider ID/model metadata tuples.
```

src/Builders/PromptBuilder.php
```diff
@@ -216,9 +223,74 @@ public function usingModel(ModelInterface $model): self
...
+                $modelId = $preferredModel->metadata()->getId();
+                $providerId = $preferredModel->providerMetadata()->getId();
+                $providerModelKey = $this->createProviderModelPreferenceKey($providerId, $modelId);
+                $preferenceMap[$providerModelKey] = [$modelId, $providerId];
+
+            } elseif (is_string($preferredModel)) {
+                // Model ID
+                $modelId = $this->normalizePreferenceIdentifier($preferredModel);
+                $modelKey = $this->createModelPreferenceKey($modelId);
+                $preferenceMap[$modelKey] = $modelId;
```
felixarntz: It looks like you're doing the reverse of what I proposed here: You're creating a map of the preferences, but I was proposing to create a map of all the candidates - these are often very long lists.

And then further below, it's not right to iterate through the candidates and return the first preference found - we need to iterate through the preferences and return the first candidate found. Otherwise, the preference order is not respected.

src/Builders/PromptBuilder.php
```diff
@@ -1066,41 +1169,130 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+                return $this->registry->getProviderModel($providerId, $modelId, $this->modelConfig);
             }
 
-            $provider = $this->providerIdOrClassName;
-            $modelMetadata = $modelsMetadata[0];
+            $modelKey = $this->createModelPreferenceKey($modelId);
+            if (isset($this->modelPreferenceMap[$modelKey])) {
+                return $this->registry->getProviderModel($providerId, $modelId, $this->modelConfig);
+            }
         }
```
felixarntz: See above, this doesn't look right. Instead, we need to iterate through the model preference list (which shouldn't be a map), and return the model as soon as a candidate is found that matches the model from the preference list.

It's the candidates that need to be put in a map, the lookup needs to be the other way around. Otherwise it doesn't respect the model preference order.

src/Builders/PromptBuilder.php
```diff
@@ -1066,41 +1169,130 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+     * @since n.e.x.t
+     *
+     * @param list<array{0:string,1:ModelMetadata}> $candidateModels Candidate provider/model tuples.
+     * @return ModelInterface|null The model matching the highest-priority preference, or null if none.
+     */
+    private function findModelByPreference(array $candidateModels): ?ModelInterface
+    {
+        if ($this->modelPreferenceMap === []) {
+            return null;
+        }
```
felixarntz: I think we can avoid some code duplication and make this more intuitive, by always going through this method:
* If no model preferences map is set, we can simply get the model instance based on `$candidateModels[0]` in here (or rather, the first entry of the candidates map, per my other feedback).
* No need to have another conditional after calling this method.

This will also allow to change the method return type from `?ModelInterface` to `ModelInterface`.

felixarntz (commented):
I think there's a fundamental flaw in the current logic, it's doing the reverse lookup of what needs to happen for the model preference list to work properly.

src/Builders/PromptBuilder.php
JasonTheAdams: Removed this method in 79a7a56

src/Builders/PromptBuilder.php
JasonTheAdams: You're right! Refactored in 79a7a56

src/Builders/PromptBuilder.php
JasonTheAdams: Significantly refactored in 79a7a56

src/Builders/PromptBuilder.php
JasonTheAdams: Refactored in 79a7a56

src/Builders/PromptBuilder.php
JasonTheAdams: Refactored in 79a7a56

JasonTheAdams:
Ok, @felixarntz I significantly changed how the mapping works. I'm a bit cross-eyed, but I think this is better. Hahah! Here's how it works now:

1. Instead of storing the model preferences as they were, it's now normalized to a list of the map keys, as that's really all we need for later
2. The candidate models are mapped, so iteration is no longer necessary
3. I used array intersection instead of iteration to find the first preferred model

I'm not really sure how else we'd boost performance, now. I think we're iterating as little as possible. 😄

src/Builders/PromptBuilder.php
```diff
@@ -216,9 +222,72 @@ public function usingModel(ModelInterface $model): self
...
+                $modelId = $this->normalizePreferenceIdentifier($preferredModel);
+                $modelKey = $this->createModelPreferenceKey($modelId);
+                $preferenceKeys[] = $modelKey;
+            } else {
+                // Invalid type
+                throw new InvalidArgumentException(
+                    'Model preferences must be model identifiers, instances of ModelInterface, ' .
+                    'or provider/model tuples.'
+                );
+            }
```
felixarntz: Tiny nit-pick:
```suggestion
                $preferenceKey = $this->createProviderModelPreferenceKey($providerId, $modelId);
            } elseif ($preferredModel instanceof ModelInterface) {
                // Model instance
                $modelId = $preferredModel->metadata()->getId();
                $providerId = $preferredModel->providerMetadata()->getId();

                $preferenceKey = $this->createProviderModelPreferenceKey($providerId, $modelId);
            } elseif (is_string($preferredModel)) {
                // Model ID
                $modelId = $this->normalizePreferenceIdentifier($preferredModel);

                $preferenceKey = $this->createModelPreferenceKey($modelId);
            } else {
                // Invalid type
                throw new InvalidArgumentException(
                    'Model preferences must be model identifiers, instances of ModelInterface, ' .
                    'or provider/model tuples.'
                );
            }

            $preferenceKeys[] = $preferenceKey;
```

src/Builders/PromptBuilder.php
```diff
@@ -1040,67 +1109,179 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
-                            return $opt->getName()->value . '=' . json_encode($opt->getValue());
-                        }, $requirements->getRequiredOptions()))
-                    )
-                );
+            if (!empty($matchingPreferences)) {
+                // Get the first matching preference key
+                $firstMatchKey = key($matchingPreferences);
+                [$providerId, $modelId] = $candidateMap[$firstMatchKey];
+
+                $selectedModel = $this->registry->getProviderModel($providerId, $modelId, $this->modelConfig);
```
felixarntz: I don't think we need the `$selectedModel` variable? We can just do an early return here.
```suggestion
                return $this->registry->getProviderModel($providerId, $modelId, $this->modelConfig);
```

src/Builders/PromptBuilder.php
```diff
@@ -1040,67 +1109,179 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+        if ($selectedModel === null) {
+            // No preference matched; fall back to the first candidate discovered.
+            [$providerId, $modelId] = reset($candidateMap);
+
+            $selectedModel = $this->registry->getProviderModel(
+                $providerId,
+                $modelId,
+                $this->modelConfig
+            );
         }
```
felixarntz: And then here:
```suggestion
        // No preference matched; fall back to the first candidate discovered.
        [$providerId, $modelId] = reset($candidateMap);

        return $this->registry->getProviderModel($providerId, $modelId, $this->modelConfig);
```

src/Builders/PromptBuilder.php
felixarntz: I don't think you actually addressed the concern I mentioned above?

src/Builders/PromptBuilder.php
```diff
@@ -1040,67 +1109,179 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+            return $candidateMap;
+        }
+
+        // Provider set, only consider models from that provider.
+        $modelsMetadata = $this->registry->findProviderModelsMetadataForSupport(
+            $this->providerIdOrClassName,
+            $requirements
         );
+
+        return $this->generateMapFromCandidates($this->providerIdOrClassName, $modelsMetadata);
```
felixarntz: See my comment in https://github.com/WordPress/php-ai-client/pull/110#discussion_r2433846822, I think this is still not right here.

src/Builders/PromptBuilder.php
```diff
@@ -1040,67 +1109,179 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+        foreach ($modelsMetadata as $modelMetadata) {
+            $modelId = $modelMetadata->getId();
+
+            // Add provider-specific key
+            $providerModelKey = $this->createProviderModelPreferenceKey($providerId, $modelId);
+            $map[$providerModelKey] = [$providerId, $modelId];
+
+            // Add model-only key (when merging multiple maps, the + operator preserves first occurrence)
+            $modelKey = $this->createModelPreferenceKey($modelId);
+            $map[$modelKey] = [$providerId, $modelId];
```
felixarntz: What is the + in the comment referring to? I don't see any + in the code :)

felixarntz (commented):
@JasonTheAdams High-level logic looks solid now, there are just a few minor notes. I think the only blocker to merging is the point on properly normalizing the provider class name or ID, so that it's always an ID when using it in a key.

src/Builders/PromptBuilder.php
JasonTheAdams: Good catch! Leftover from refactoring. Resolved in 2101ede

src/Builders/PromptBuilder.php
JasonTheAdams: Resolved in 2101ede

src/Builders/PromptBuilder.php
JasonTheAdams: Oops! A leftover from a previous state. Resolved in 1590816

src/Builders/PromptBuilder.php
JasonTheAdams: Thank you for bringing this back up! You were right! I also found there wasn't a test covering a provider class name being used, which hid this bug. A test has been added and resolved in 8771816

felixarntz (approved):
@JasonTheAdams Awesome, looks great! 🚀

PR #111, opened 2025-10-17
Make provider ID lookup by class more efficient and align provider class name lookup

felixarntz:
Minor enhancement / DX fix, as a follow up to the new `getProviderId` method added in #110. Nothing critical, just makes things a little bit better :)

Also adds unit test coverage.


JasonTheAdams (approved):
Great follow up! 💪

PR #112, opened 2025-10-17
Allow specifying model preference in CLI tool for testing

felixarntz:
After #110, it makes sense to allow specifying a model preference in our little CLI tool for testing.

Example usage:
```sh
php cli.php 'When was WordPress first released?' --modelPreference='gemini-2.5-flash-lite, gpt-5-mini'
```

### DIFF
--- a/cli.php
+++ b/cli.php
@@ -100,6 +100,7 @@ if (str_starts_with($promptInput, '{') || str_starts_with($promptInput, '[')) {
 // Provider ID, model ID, and output format.
 $providerId = $named_args['providerId'] ?? null;
 $modelId = $named_args['modelId'] ?? null;
+$modelPreference = $named_args['modelPreference'] ?? null;
 $outputFormat = $named_args['outputFormat'] ?? 'message-text';
 
 // Any model configuration options.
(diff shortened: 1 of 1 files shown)


felixarntz:
@JasonTheAdams I know we discussed this as part of #110 and for now went with `[modelId, providerId]` when specifying both model and its provider, but now working with it, I want to challenge it again.

It makes sense from the perspective that the model is the centerpiece, but it makes DX a bit weird IMO. It's just not intuitive, at least I think so, to have a list like:
```php
[
    'gemini-2.5-flash',
    ['gemini-2.5-pro', 'google'],
    ['gpt-5-high', 'openai'],
]
```

I think it's more intuitive to have:
```php
[
    'gemini-2.5-flash',
    ['google', 'gemini-2.5-pro'],
    ['openai', 'gpt-5-high'],
]
```

Especially when you get to a syntax where you want to express both (relevant for CLI tools, like ours, but any other CLI tool too, as they don't support array input), having to write `gemini-2.5-flash::google` or `gemini-2.5-flash//google` or so feels even weirder.

What do you think? Would you be aligned to change this? Let's think and talk this through a bit more, since we need to ideally decide and then commit to it before the 0.2.0 release.

JasonTheAdams:
This was not on purpose. Hahah! I meant it to be `[provider, model]`, but apparently missed that detail in the flurry of things. Let's swap that order! 👍

cli.php
```diff
@@ -149,6 +150,19 @@ function logError(string $message, int $exit_code = 1): void
     } elseif ($providerId) {
         $promptBuilder = $promptBuilder->usingProvider($providerId);
     }
+    if ($modelPreference) {
+        $modelPreference = array_map(
+            static function ($item) {
+                $item = trim($item);
+                if (str_contains($item, '::')) {
+                    return explode('::', $item, 2);
```
JasonTheAdams, line 158: TIL about the `explode()` limit parameter! 🤩

JasonTheAdams (approved):
Looks good! I even learned a thing!

PR #113, opened 2025-10-17
Fix provider model preference tuple order to have provider ID be first

felixarntz:
See https://github.com/WordPress/php-ai-client/pull/112#issuecomment-3416697304 for what this fixes.

It also fixes a problem in the unit test which was actually passing regardless of whether model ID or provider ID came first. That's because the expected preferred model was also the first one registered, so it would be returned regardless. Including another model that's registered before that makes the test actually meaningful (i.e. pass/fail as expected).

### DIFF
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -249,9 +249,9 @@ class PromptBuilder
                     );
                 }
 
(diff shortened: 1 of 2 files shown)


JasonTheAdams (approved):
Thanks for fixing this! LGTM! 🚀

PR #114, opened 2025-10-17
Updates preferred model tests to avoid passing by default

JasonTheAdams:
@felixarntz caught in #113 that a preferred model test was passing incorrectly. The reason was that even the preferred model system failed, the default model would be the first model... which was also the model expected by preference.

This corrects a few other tests that were also running into the same false sense of security.

Now, I'm hoping, we can all sleep better at night, only checking under our beds once.


felixarntz (approved):
@JasonTheAdams Thanks, this is great!

PR #115, opened 2025-10-17
Adds AbstractApiProvider class

### LINKED ISSUE #68: Provide an AbstractApiProvider to clean up paths
There are many places where a provider/model sets the various API endpoint paths. Each places concatenates the path and uses things like `ltrim()`. It's redundant and opens up for simple mistakes. It would be great to have an `AbstractApiProvider` class (and likely a corresponding provider interface) that is aware of a base URL and has an `AbstractApiProvider::url($path)` counterpart, which cleans this up.

JasonTheAdams:
Resolves #68 

This adds an `AbstractApiProvider` class, which provides a convenient way for API urls to be constructed, cleaning up a bunch of duplicate path constructing code.

I went with an abstract `baseUrl` method instead of using a constant that gets overridden so that it's clearer to extending classes and doesn't fail at runtime.

### DIFF
--- a/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
+++ b/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
@@ -22,7 +22,7 @@ class GoogleImageGenerationModel extends AbstractOpenAiCompatibleImageGeneration
     {
         return new Request(
             $method,
-            GoogleProvider::url("openai/{$path}"),
+            GoogleProvider::url('openai' . ltrim($path, '/')),
             $headers,
             $data
         );
--- a/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
+++ b/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
@@ -22,7 +22,7 @@ class GoogleImageGenerationModel extends AbstractOpenAiCompatibleImageGeneration
     {
         return new Request(
             $method,
-            GoogleProvider::BASE_URI . '/openai/' . ltrim($path, '/'),
+            GoogleProvider::url("openai/{$path}"),
             $headers,
             $data
         );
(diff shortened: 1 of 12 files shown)


src/ProviderImplementations/Google/GoogleTextGenerationModel.php
```diff
@@ -24,7 +24,7 @@ protected function createRequest(HttpMethodEnum $method, string $path, array $he
     {
         return new Request(
             $method,
-            GoogleProvider::BASE_URI . '/openai/' . ltrim($path, '/'),
+            GoogleProvider::url("openai/{$path}"),
```
felixarntz: The `ltrim($path)` needs to be maintained here :)

src/ProviderImplementations/Google/GoogleImageGenerationModel.php
```diff
@@ -22,7 +22,7 @@ protected function createRequest(HttpMethodEnum $method, string $path, array $he
     {
         return new Request(
             $method,
-            GoogleProvider::BASE_URI . '/openai/' . ltrim($path, '/'),
+            GoogleProvider::url("openai/{$path}"),
```
felixarntz: Same here.

felixarntz (approved):
@JasonTheAdams Two tiny but important problems here that need to be addressed before merge, but otherwise this looks good to go, so I'll preemptively approve.

Please feel free to merge as soon as you have addressed these. 🚢

src/ProviderImplementations/Google/GoogleTextGenerationModel.php
JasonTheAdams: Despite my best efforts to over-engineer this, it's resolved in 71ee912

src/ProviderImplementations/Google/GoogleImageGenerationModel.php
JasonTheAdams: Resolved in 71ee912

PR #120, opened 2025-11-12
Add missing `@throws` annotation

mujuonly:
Missing @throws tag in function comment


felixarntz (approved):
@mujuonly Thanks, great catch!

PR #121, opened 2025-11-17
Sort models per provider based on reasonable heuristics to avoid random default model

### LINKED ISSUE #117: Account for OpenAI's search models
### Issue

While testing something out that utilizes this client to make requests, things were working great yesterday but started to fail today with the following error: `Invalid arguments while trying to generate text result: Bad Request (400) - Model incompatible request argument supplied: temperature`.

In debugging further, found that if I only have OpenAI configured and I don't specify a model, seems today it started picking `gpt-4o-mini-search-preview` as the model. Not sure if that's a new model that OpenAI just released recently or if they changed the model list their API returns but somehow that now ends up on the top of the list when we try and determine which model works best.

But as the error shows, that model doesn't support standard parameters like `temperature` or `n` and so the request doesn't work.

This does highlight a downside of the approach taken in this SDK, where we try and intelligently choose the provider and model. Anytime a new model is released or the model APIs we use change the order of their results, we could end up breaking integrations. I was just testing things locally so while this issue was annoying it wasn't critical but for any production integrations, would be ideal if there was a bit more certainty things won't just break one day.

There's potentially a conversation we could have around that (for instance, should we be setting preferred models within the client to avoid breaking changes like this?) but for now, seems like we need to update the `OpenAiModelMetadataDirectory` class to account for this search model.

### Steps to Reproduce

Things may change but as of right now, I can easily reproduce this by calling our `cli` command:

```bash
OPENAI_API_KEY=sk-proj-XXXX php cli.php 'Hello' --temperature=0.9
```

If the model the SDK chooses changes in the future, can reproduce by running this:

```bash
OPENAI_API_KEY=sk-proj-XXXX php cli.php 'Hello' --modelId=gpt-4o-mini-search-preview --temperature=0.9
```

felixarntz:
Fixes #117

This finally resolves a long-standing issue where the default model chosen for a provider would be sort of "random", or rather the first model returned by the provider's API listings.

The solution is to implement model sorting _per provider_ via reasonable heuristics, without getting too opinionated:

* prefer flagship models (and then the common "lower/cheaper tier" variant, if applicable)
* prefer later versions
* prefer non-preview/non-experimental

Keep in mind that the sorting is _per provider_. There is no issue with this in terms of favoring any specific provider, the sorting is just for their models internally.

For easy testing of the sorted model IDs per provider, do this:

* Temporarily add a `var_dump($modelId);` for each `$modelId` iterated over in `PromptBuilder::generateMapFromCandidates` (e.g. line 1216).
* Use the built-in CLI tool, without specifying a model or model preference, but specifying `--providerId` (with each "anthropic", "google", "openai").

### DIFF
--- a/src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
+++ b/src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
@@ -180,7 +180,7 @@ class AnthropicModelMetadataDirectory extends AbstractOpenAiCompatibleModelMetad
             return 1;
         }
 
-        // Prefer Claude models with type, version number, and date (e.g. 'claude-sonnet-4') over those without.
+        // Prefer Claude models with type, version number, and date (e.g. 'claude-sonnet-4-5-20250929') over those without.
         $aMatch = preg_match('/^claude-([a-z]+)-(\d(-\d)?)-([0-9]+)$/', $aId, $aMatches);
         $bMatch = preg_match('/^claude-([a-z]+)-(\d(-\d)?)-([0-9]+)$/', $bId, $bMatches);
         if ($aMatch && !$bMatch) {
--- a/src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
+++ b/src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
@@ -109,7 +109,7 @@ class AnthropicModelMetadataDirectory extends AbstractOpenAiCompatibleModelMetad
 
         $modelsData = (array) $responseData['data'];
 
-        return array_values(
+        $models = array_values(
             array_map(
                 static function (array $modelData) use (
                     $anthropicCapabilities,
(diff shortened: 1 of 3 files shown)


src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
```diff
@@ -137,5 +137,89 @@ static function (array $modelData) use (
...
+        /*
+         * Prefer Claude models where the version number isn't the second segment (e.g. 'claude-sonnet-4')
+         * over those where it is (e.g. 'claude-2', 'claude-3-5-sonnet'). The latter is only used for older models.
+         */
+        if (!preg_match('/^claude-\d/', $aId) && preg_match('/^claude-\d/', $bId)) {
+            return -1;
+        }
+        if (!preg_match('/^claude-\d/', $bId) && preg_match('/^claude-\d/', $aId)) {
+            return 1;
+        }
```
JasonTheAdams, line 181: I wonder if we can simplify this and below by parsing out the version number(s) and dates, and sorting it in descending order? I'm mildly concerned they may do the OpenAI thing and decide at some point to drop back to `claude-5` or whatever.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
```diff
@@ -137,5 +137,89 @@ static function (array $modelData) use (
...
+
+        // Prefer Claude models with type, version number, and date (e.g. 'claude-sonnet-4') over those without.
+        $aMatch = preg_match('/^claude-([a-z]+)-(\d(-\d)?)-([0-9]+)$/', $aId, $aMatches);
```
JasonTheAdams: Mind adding an example with a date suffix?

JasonTheAdams (commented):
Oh wow! I was literally thinking through doing something like this as I was getting my day going this morning! Thanks for busting this out!

Just a couple comments/questions for the moment.

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
felixarntz: I think that would be more brittle from another perspective, we don't know for sure what shapes those model names could possibly have in the future. I think it makes most sense to optimize for the patterns there currently are. When a new one emerges in the future, we'll inevitably have to update the code here anyway. Anything else is hypothetical IMO :)

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
felixarntz: Done in https://github.com/WordPress/php-ai-client/pull/121/commits/25d9d2d10e0b2fdf1a7ae98483dce0373e880b1f

src/ProviderImplementations/Anthropic/AnthropicModelMetadataDirectory.php
JasonTheAdams: That's true! 👍

JasonTheAdams (approved):
This is great! I think it will really improve the default behavior for these Providers!

PR #122, opened 2025-11-17
Fix incorrect model options being set for GPT search models

### LINKED ISSUE #117: Account for OpenAI's search models

felixarntz:
Fixes #117

This simply fixes the incorrect options configured for the recently launched OpenAI `-search` models, which don't support most configuration options due to their special purpose.

### DIFF
--- a/src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
+++ b/src/ProviderImplementations/OpenAi/OpenAiModelMetadataDirectory.php
@@ -109,6 +109,14 @@ class OpenAiModelMetadataDirectory extends AbstractOpenAiCompatibleModelMetadata
                 ]
             ),
         ]);
+        $gptSearchOptions = [
+            new SupportedOption(OptionEnum::systemInstruction()),
+            new SupportedOption(OptionEnum::outputMimeType(), ['text/plain', 'application/json']),
+            new SupportedOption(OptionEnum::outputSchema()),
+            new SupportedOption(OptionEnum::customOptions()),
+            new SupportedOption(OptionEnum::inputModalities(), [[ModalityEnum::text()]]),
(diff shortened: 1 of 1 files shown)


JasonTheAdams (approved):
Thanks for this!

PR #123, opened 2025-11-17
Fix broken Google API URLs

felixarntz:
Yikes! Requests to the Google API are currently not working at all.

Looks like this broke in #115 - more specifically, see https://github.com/WordPress/php-ai-client/pull/115/files#diff-34150b8509051ad1c9d1e475df24d5454c3a994273831cb477bd01fa172b186aR25 🤦 

We need more comprehensive testing of the concrete provider implementations to prevent such errors to go unnoticed in the future. I think ideally we can actually fire a cheap request to each provider, to really ensure things work end to end, as this would further help to recognize breakage in case a provider makes a breaking change.


JasonTheAdams (approved):
Oops! Good catch! LGTM! 🚀

PR #133, opened 2025-11-26
Prompt Builder can set model request options

### LINKED ISSUE #127: Allow specifying `RequestOptions` as part of `PromptBuilder`
While in #99 we implemented support for `RequestOptions`, we never enabled callers with a reasonable way to provide them. I don't think it's currently even possible to pass request options alongside a prompt that would be used for the eventual API request to the cloud provider.

Here's what I'm proposing:

- expand `PromptBuilder` with a `usingRequestOptions(RequestOptions $requestOptions): self` method
- expand `ModelInterface` with `setRequestOptions(RequestOptions $requestOptions): void` and `getRequestOptions(): ?RequestOptions` methods
- update the provider implementations to call `getRequestOptions()` when constructing their API requests

By implementing this in `ModelInterface` and `AbstractAiProvider`, this is _almost_ not even a breaking change. It technically is, if you're not using `AbstractAiProvider`. But we're so early in the lifecycle of this SDK, I doubt anyone is building custom providers yet.

JasonTheAdams:
Resolves #127

This introduces the request options as something settable through the `PromptBuilder`. If the underlying model is an instance of `AbstractApiBasedModel` then it will bind the request options to the model. Then, when model requests are made, the request options are included.

This allows for things like changing the timeout for models that like to think and take a longer amount of time.

### DIFF
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -13,7 +13,7 @@ use WordPress\AiClient\Messages\DTO\MessagePart;
 use WordPress\AiClient\Messages\DTO\UserMessage;
 use WordPress\AiClient\Messages\Enums\MessageRoleEnum;
 use WordPress\AiClient\Messages\Enums\ModalityEnum;
-use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModel;
+use WordPress\AiClient\Providers\ApiBasedImplementation\Contracts\ApiBasedModelInterface;
 use WordPress\AiClient\Providers\Http\DTO\RequestOptions;
 use WordPress\AiClient\Providers\Models\Contracts\ModelInterface;
 use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
@@ -1199,7 +1199,7 @@ class PromptBuilder
      */
     private function bindModelRequestOptions(ModelInterface $model): void
     {
-        if ($this->requestOptions !== null && $model instanceof AbstractApiBasedModel) {
+        if ($this->requestOptions !== null && $model instanceof ApiBasedModelInterface) {
             $model->setRequestOptions($this->requestOptions);
         }
     }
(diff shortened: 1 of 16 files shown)


src/Builders/PromptBuilder.php
```diff
@@ -1150,14 +1173,35 @@ private function getConfiguredModel(CapabilityEnum $capability): ModelInterface
...
+     * Request options are only applicable to API-based models that make HTTP requests.
+     *
+     * @since n.e.x.t
+     *
+     * @param ModelInterface $model The model to bind request options to.
+     * @return void
+     */
+    private function bindModelRequestOptions(ModelInterface $model): void
+    {
+        if ($this->requestOptions !== null && $model instanceof AbstractApiBasedModel) {
```
felixarntz: I think it would be better to introduce an `ApiBasedModelInterface` contract that extends `ModelInterface` and has the setter and getter for request options. That way we can follow the best practice of checking against contracts instead of concrete implementations.

src/Providers/ApiBasedImplementation/AbstractApiBasedModel.php
```diff
@@ -44,6 +45,11 @@ abstract class AbstractApiBasedModel implements
      */
     private ModelConfig $config;
 
+    /**
+     * @var RequestOptions|null The request options for HTTP transport.
+     */
+    private ?RequestOptions $requestOptions = null;
```
felixarntz, line 51: Based on my other comment, this class would then implement `ApiBasedModelInterface`.

felixarntz (commented):
@JasonTheAdams Looks almost good to go, just one improvement suggestion.

src/Builders/PromptBuilder.php
JasonTheAdams: Good idea! Resolved in f4cf2a7

src/Providers/ApiBasedImplementation/AbstractApiBasedModel.php
JasonTheAdams: Resolved in f4cf2a7

JasonTheAdams:
Thanks, @felixarntz! Back to you!

felixarntz (approved):

PR #134, opened 2025-11-26
Make `PromptBuilder::isSupported` public

### LINKED ISSUE #128: Make `PromptBuilder::isSupported` public
There's no good reason to keep this method private. `generateResult`, which is the equivalent for actual generation is already public. That method is useful when approaching things from a truly multimodal perspective, and we should allow checking for support to work similarly.

Last but not least: Making this public will avoid ugly conditionals like this one: https://github.com/WordPress/wp-ai-client/pull/27/files#diff-6ffc16c3dfbbfc9f75adef72902d8024ef48815359fceae974734c847112e409R160

felixarntz:
Fixes #128

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -340,6 +340,7 @@ direction LR
             +convertTextToSpeeches(?int $candidateCount) File[]
             +generateSpeech() File
             +generateSpeeches(?int $candidateCount) File[]
+            +isSupported(?CapabilityEnum $capability) bool
             +isSupportedForTextGeneration() bool
             +isSupportedForImageGeneration() bool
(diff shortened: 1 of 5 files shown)


JasonTheAdams (changes requested):
@felixarntz Looks good! Just one suggestion: Now that this method is public, let's add unit tests for it.

felixarntz:
@JasonTheAdams Tests added!

I've also refreshed this with latest `trunk`, and bumped the `@since n.e.x.t` annotations to say `@since 0.3.0`. So we should be able to ship shortly after this is merged, pending some smoke testing using `cli.php`.

JasonTheAdams (approved):
Looks great! 🚀

PR #135, opened 2025-11-26
Allow specifying optional provider credentials URL in provider metadata

### LINKED ISSUE #129: Allow specifying optional provider credentials URL in `ProviderMetadata`
A very trivial enhancement. Providing a URL where to get credentials is crucial for users to know where to go if they have to manually configure the provider.

Case in point: https://github.com/WordPress/wp-ai-client/blob/trunk/includes/API_Credentials/API_Credentials_Settings_Screen.php#L121

This is optional, since not every provider will require this (although today's cloud providers all do).

felixarntz:
Fixes #129

### DIFF
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -18,7 +18,8 @@ use WordPress\AiClient\Providers\Enums\ProviderTypeEnum;
  * @phpstan-type ProviderMetadataArrayShape array{
  *     id: string,
  *     name: string,
- *     type: string
+ *     type: string,
+ *     credentialsUrl?: ?string
  * }
  *
  * @extends AbstractDataTransferObject<ProviderMetadataArrayShape>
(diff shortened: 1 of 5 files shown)


src/Providers/DTO/ProviderMetadata.php
```diff
@@ -52,12 +59,14 @@ class ProviderMetadata extends AbstractDataTransferObject
      * @param string $id The provider's unique identifier.
      * @param string $name The provider's display name.
      * @param ProviderTypeEnum $type The provider type.
+     * @param string|null $credentialsUrl The URL where users can get credentials.
      */
-    public function __construct(string $id, string $name, ProviderTypeEnum $type)
+    public function __construct(string $id, string $name, ProviderTypeEnum $type, ?string $credentialsUrl = null)
```
JasonTheAdams, line 64: Now that this is four parameters long and we're starting to get into optional parameters, I'm curious if we should consider either an array as an optional fourth argument that can extend over time to include more, or if we should instead add a setter for things like this.

Another option is to stick with this for now and then change it to an array later if/when the time arrives and make this parameter backwards-compatible.

What do you think, @felixarntz?

felixarntz: I'd say let's do the latter. It'll be easy to keep things backward compatible in the way you're saying.

JasonTheAdams (approved):
LGTM! ✅

PR #139, opened 2025-11-30
Removing an old Docblock as includeRequiredOption was previously deleted

mpeshev:
Found an orphaned docblock after removing includeRequiredOption from the PromptBuilder class


felixarntz (approved):
@mpeshev Thank you, great catch!

PR #140, opened 2025-11-30
json_last_error() usage for json_decode validation for compliance

mpeshev:
json_last_error() is used across cli.php, but this one was missing a proper check/validation. Pushing a small update for consistency and avoiding warnings.


felixarntz (approved):
@mpeshev LGTM, thanks!

PR #144, opened 2025-12-05
Introduce event dispatching

### LINKED ISSUE #142: Add support for event dispatching
There's currently no way to catch eventful moments within the prompting process, such as before and after a prompt is fulfilled, or if it fails. Having events like this is useful, at the very least for systems such as logging.

We should implement the [PSR-14](https://www.php-fig.org/psr/psr-14/) event dispatching interfaces so implementing systems can attach this to whatever event system they have. For WP AI Client we can connect this to actions.

JasonTheAdams:
Resolves #142 

This introduces an abstract system for event dispatching using PSR-14 interfaces. An event dispatcher is set on the `AiClient` which can then be used throughout the system. There are currently two events:
* `AfterPromptSentEvent`
* `BeforePromptSentEvent`

We could add more, of course, but these seem like the two most sensible. 

Being abstract, it's up to the implementing system to connect this to whatever concrete event system they have. It's also their responsibility to set up the `PSR-14` event listeners, which would be fired when the dispatching is fulfilled. This makes the actual event system within the PHP AI Client pretty lightweight.

An intention here is to have a WP AI Client counterpart which hooks this up to the WordPress actions system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ The AI Client supports PSR-14 event dispatching for prompt lifecycle events. Thi
 - `BeforeGenerateResultEvent` - Dispatched before a prompt is sent to the model
 - `AfterGenerateResultEvent` - Dispatched after a result is received from the model
 
+**Important:** Event listeners should not return a value, as they will be ignored. In order to modify data that is passed with the event object, you need to rely on setters on the event object. Any event data for which there are no setters on the event object is meant to be immutable or, in other words, read-only for the event listener.
+
 ### Connecting Your Event Dispatcher
 
 To enable event dispatching, pass any PSR-14 compatible `EventDispatcherInterface` to the client:
--- a/README.md
+++ b/README.md
@@ -80,6 +80,57 @@ See the [`PromptBuilder` class](https://github.com/WordPress/php-ai-client/blob/
 
 **More documentation is coming soon.**
 
+## Event Dispatching
+
+The AI Client supports PSR-14 event dispatching for prompt lifecycle events. This allows you to hook into the generation process for logging, monitoring, or other integrations.
+
+### Available Events
+
+- `BeforeGenerateResultEvent` - Dispatched before a prompt is sent to the model
+- `AfterGenerateResultEvent` - Dispatched after a result is received from the model
+
+### Connecting Your Event Dispatcher
+
+To enable event dispatching, pass any PSR-14 compatible `EventDispatcherInterface` to the client:
+
+```php
+use WordPress\AiClient\AiClient;
+
+// Set your PSR-14 event dispatcher
+AiClient::setEventDispatcher($yourEventDispatcher);
+
+// Events will now be dispatched during generation
+$text = AiClient::prompt('Hello, world!')
+    ->generateText();
+```
+
+### Example: Logging Events
+
+```php
+use WordPress\AiClient\Events\BeforeGenerateResultEvent;
+use WordPress\AiClient\Events\AfterGenerateResultEvent;
+
+// In your event listener/subscriber
+class AiEventListener
+{
+    public function onBeforeGenerate(BeforeGenerateResultEvent $event): void
+    {
+        $model = $event->getModel();
+        $messages = $event->getMessages();
+        $capability = $event->getCapability();
+
+        // Log, monitor, or perform other actions
+    }
+
+    public function onAfterGenerate(AfterGenerateResultEvent $event): void
+    {
+        $result = $event->getResult();
+
+        // Log the result, track usage, etc.
+    }
+}
+```
+
 ## Further reading
 
 For more information on the requirements and guiding principles, please review:
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -149,15 +149,25 @@ class AiClient
     }
 
     /**
-     * Checks if an event dispatcher is registered.
+     * Dispatches an event if an event dispatcher is registered.
+     *
+     * This is a convenience method that handles the null check internally,
+     * only dispatching if a dispatcher has been set via setEventDispatcher().
      *
      * @since n.e.x.t
      *
-     * @return bool True if an event dispatcher is set, false otherwise.
+     * @template T of object
+     * @param T $event The event to dispatch.
+     * @return T The event (potentially modified by listeners).
      */
-    public static function hasEventDispatcher(): bool
+    public static function dispatchEvent(object $event): object
     {
-        return self::$eventDispatcher !== null;
+        if (self::$eventDispatcher !== null) {
+            /** @var T */
+            return self::$eventDispatcher->dispatch($event);
+        }
+
+        return $event;
     }
 
     /**
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use WordPress\AiClient\Builders\PromptBuilder;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
@@ -91,6 +92,11 @@ class AiClient
     private static ?ProviderRegistry $defaultRegistry = null;
 
     /**
+     * @var EventDispatcherInterface|null The event dispatcher for prompt lifecycle events.
+     */
+    private static ?EventDispatcherInterface $eventDispatcher = null;
+
+    /**
      * Gets the default provider registry instance.
      *
      * @since 0.1.0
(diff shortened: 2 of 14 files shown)


JasonTheAdams:
@felixarntz Do you think it's worth adding our own event interface so subsequent dispatchers can identify any events coming from our system? I think so, yeah? 🤔

src/AiClient.php
```diff
@@ -114,6 +120,56 @@ public static function defaultRegistry(): ProviderRegistry
...
+     */
+    public static function dispatchEvent(object $event): object
+    {
+        if (self::$eventDispatcher !== null) {
+            /** @var T */
+            return self::$eventDispatcher->dispatch($event);
+        }
+
+        return $event;
+    }
```
felixarntz: Instead of having this in `AiClient`, I think we should come up with a dedicated class like this, mirroring the approach of `HttpTransporter`, which receives PSR interfaces for request handling.

That new class could receive an `EventDispatcherInterface` in the constructor.

src/AiClient.php
```diff
@@ -114,6 +120,56 @@ public static function defaultRegistry(): ProviderRegistry
...
+     *
+     * @template T of object
+     * @param T $event The event to dispatch.
+     * @return T The event (potentially modified by listeners).
+     */
+    public static function dispatchEvent(object $event): object
+    {
+        if (self::$eventDispatcher !== null) {
+            /** @var T */
+            return self::$eventDispatcher->dispatch($event);
```
felixarntz: I wonder whether we want to be a bit safer here.

I don't love that an event listener can arbitrarily replace the entire instance with another instance. I think we'll want to remain in control of what can and what cannot be modified.

While PSR-14 doesn't mandate that, it doesn't mean we cannot be stricter. I think we should include here a check to ensure that the return value is `$modified_event === $event`, and otherwise throw.

Yes, that technically means we wouldn't even need to return it, since objects are modified by reference. But I think this is preferable since that leaves us in control.

For example, right now you have a `setMessages` method on the event implementations, but having that doesn't even matter, since a listener could just create a new instance of the class and replace _everything_. I think that's not great.

src/Builders/PromptBuilder.php
```diff
@@ -826,7 +829,39 @@ public function generateResult(?CapabilityEnum $capability = null): GenerativeAi
 
         $model = $this->getConfiguredModel($capability);
 
+        // Dispatch BeforePromptSentEvent (allows message modification)
+        $beforeEvent = AiClient::dispatchEvent(
```
felixarntz: Following up to my above comment: Can we instead access an instance of our new `EventDispatcherInterface` wrapper class here?

src/Builders/PromptBuilder.php
```diff
@@ -826,7 +829,39 @@ public function generateResult(?CapabilityEnum $capability = null): GenerativeAi
...
+            new BeforePromptSentEvent($this->messages, $model, $capability)
+        );
+        $messages = $beforeEvent->getMessages();
```
felixarntz: Why do we want to allow modifying the messages here? I would prefer if we didn't allow that, at least not from the start. We can always add it later if there's a need for it.

I think for now I'd prefer that we kept these events "immutable", I.e. more like an action in WordPress, not a filter.

src/Builders/PromptBuilder.php
```diff
@@ -872,15 +907,13 @@ public function generateResult(?CapabilityEnum $capability = null): GenerativeAi
...
-            return $model->generateSpeechResult($this->messages);
+            return $model->generateSpeechResult($messages);
         }
 
         if ($capability->isVideoGeneration()) {
-            // Video generation is not yet implemented
             throw new RuntimeException('Output modality "video" is not yet supported.');
         }
 
-        // TODO: Add support for other capabilities when interfaces are available
```
felixarntz, line 895: Why remove these comments?

src/Events/AfterPromptSentEvent.php
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Event dispatched after a prompt has been sent to the AI model and a response received.
+ *
+ * This event allows listeners to inspect the result of the model call for logging,
+ * analytics, or other post-processing purposes. The result object is immutable.
+ *
+ * @since n.e.x.t
+ */
+class AfterPromptSentEvent
```
felixarntz: How about instead naming this `AfterGenerateResultEvent`? I think that fits better with the logic as part of which is run, and is less low-level than whether "prompts were sent".

src/Events/BeforePromptSentEvent.php
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Event dispatched before a prompt is sent to the AI model.
+ *
+ * This event allows listeners to inspect and modify the messages before they
+ * are sent to the model. The event is not stoppable, meaning the model call
+ * will always proceed regardless of listener actions.
+ *
+ * @since n.e.x.t
+ */
+class BeforePromptSentEvent
```
felixarntz: See above, how about `BeforeGenerateResultEvent`?

src/Events/BeforeGenerateResultEvent.php
```diff
@@ -0,0 +1,103 @@
...
+     * Sets the messages to be sent to the model.
+     *
+     * This allows listeners to modify the messages before they are sent.
+     *
+     * @since n.e.x.t
+     *
+     * @param list<Message> $messages The modified messages.
+     * @return void
+     */
+    public function setMessages(array $messages): void
```
felixarntz: See above, I think we should remove this, at least for now.

Another reason this is better for now: If we allow altering the input here, we would be required to have the same hook run before the corresponding `is_supported_*` check, otherwise the outcome could go out of sync. We would need to ensure the data is altered in the same way in both places.

I think we can separately open an issue to discuss this. Maybe a separate event just to allow altering messages would be ideal, and that event could have a different name so that it could be fired in multiple places (both before generation and before support check).

felixarntz (changes requested):
@JasonTheAdams I think it's a solid start, but I think we can refine a couple things.

felixarntz:
@JasonTheAdams 
> Do you think it's worth adding our own event interface so subsequent dispatchers can identify any events coming from our system? I think so, yeah? 🤔

I like that idea, yeah. But it probably should simply be an empty interface. It'll allow to do exactly what you're saying, but otherwise we can adhere to the regular constraint of just `object`.

In the WP AI Client for example, we can check for the class name of the event and based on that fire specific action hooks with names that make sense according to the event class name.

src/Events/BeforePromptSentEvent.php
JasonTheAdams: Resolved in 5d1d568

src/Events/AfterPromptSentEvent.php
JasonTheAdams: Resolved in 5d1d568

src/Builders/PromptBuilder.php
JasonTheAdams: Ugh. That was not intentional. Thanks for catching that! Resolved in 069c8d3

src/Events/BeforeGenerateResultEvent.php
JasonTheAdams: Resolved in 069c8d3

src/Builders/PromptBuilder.php
JasonTheAdams: Apparently the AI decided we were done with these. 😆 

Resolved in 626979e

src/AiClient.php
JasonTheAdams: After some thought, I agree with your reasoning but not the placement. Hahah!

I think this should be a decision of the actual event dispatcher. Seeing as we're trying to be framework agnostic, we don't want to assume that our WP use case reflects how everyone wants event dispatching to work. This is something we can put in our own dispatcher in WP AI Client, but I don't think should be done here.

tests/unit/AiClientTest.php
```diff
@@ -30,7 +30,8 @@ protected function setUp(): void
 
     protected function tearDown(): void
     {
-        // Tests use dependency injection - registry instances passed directly to methods
+        // Clean up static event dispatcher after each test
+        AiClient::setEventDispatcher(null);
```
felixarntz: Why does this need to be cleaned up here? I don't see anything in this test class where it's set?

README.md
```diff
@@ -80,6 +80,57 @@ See the [`PromptBuilder` class](https://github.com/WordPress/php-ai-client/blob/
...
 **More documentation is coming soon.**
 
+## Event Dispatching
+
+The AI Client supports PSR-14 event dispatching for prompt lifecycle events. This allows you to hook into the generation process for logging, monitoring, or other integrations.
+
+### Available Events
+
+- `BeforeGenerateResultEvent` - Dispatched before a prompt is sent to the model
+- `AfterGenerateResultEvent` - Dispatched after a result is received from the model
```
felixarntz, line 90: While it's part of PSR-14, maybe we should still include an explanation here about maintaining the same event object instance?

```suggestion
- `AfterGenerateResultEvent` - Dispatched after a result is received from the model

**Important:** Event listeners should not return a value, as they will be ignored. In order to modify data that is passed with the event object, you need to rely on setters on the event object. Any event data for which there are no setters on the event object is meant to be immutable or, in other words, read-only for the event listener.
```

felixarntz (approved):
@JasonTheAdams Looks great! Two minor points.

src/AiClient.php
JasonTheAdams: We're now passing the `EventDispatcherInterface` and using the `AiClient` only for setting the dispatcher.

src/Builders/PromptBuilder.php
JasonTheAdams: We've removed this.

tests/unit/AiClientTest.php
JasonTheAdams: Resolved in 943a3e55f6

PR #145, opened 2025-12-08
Enhance error message extraction and move class to more appropriate namespace

felixarntz:
Some provider API endpoints, such as those from Google may return an array of error objects with message inside. This PR enhances the `ErrorMessageExtractor` to account for that shape, leading to more helpful error messages.

Additionally, I noticed there's both a `Providers\Http\Util` namespace and a `Providers\Http\Utilities` namespace. That doesn't make any sense, so they are being combined into one as part of this PR.

### DIFF
--- a/src/Providers/Http/Util/ErrorMessageExtractor.php
+++ b/src/Providers/Http/Util/ErrorMessageExtractor.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace WordPress\AiClient\Providers\Http\Utilities;
+namespace WordPress\AiClient\Providers\Http\Util;
 
 /**
  * Utility for extracting error messages from API response data.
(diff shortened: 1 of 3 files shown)


src/Providers/Http/Util/ErrorMessageExtractor.php
```diff
@@ -33,6 +34,18 @@ public static function extractFromResponseData($data): ?string
...
 
+        // Handle [ { "error": { "message": "Error text" } } ]
+        if (
+            isset($data[0]) &&
+            is_array($data[0]) &&
+            isset($data[0]['error']) &&
+            is_array($data[0]['error']) &&
+            isset($data[0]['error']['message']) &&
+            is_string($data[0]['error']['message'])
+        ) {
```
JasonTheAdams, line 45: You don't need to check each layer. PHP supports checking multiple layers down. 👍 
```suggestion
        if (isset($data[0]['error']['message']) && is_string($data[0]['error']['message'])) {
```

JasonTheAdams (approved):
Small suggestion, but otherwise looks good!

src/Providers/Http/Util/ErrorMessageExtractor.php
felixarntz: But PHPStan hates it 😢

PR #147, opened 2025-12-11
Fix `getProviderClassName()` return type

felixarntz:


felixarntz:
@JasonTheAdams There's some weird PHPStan issue when I change it, see https://github.com/WordPress/php-ai-client/actions/runs/20122278394/job/57744921021?pr=147: I don't see why it complains about the value being a string here. As far as I can tell, the type annotations for the underlying class properties correctly use `class-string<ProviderInterface>`. Any idea what's going on? 🤔

JasonTheAdams:
@felixarntz check out 30b3ffd

So the issue is that we're have a union `string|class-string` type being returned, and PHPStan isn't particularly good at following type narrowing by tracking the `isset()` logic the way we were going about it. So PHPStan was never confident when it logically should have been a `class-string`.

Fortunately, PHPStan has `@phpstan-assert-if-true` tags that we can use here which helps guide the static analysis system. Happily, it also makes our checks a bit more declarative. This now works! 😄

JasonTheAdams (approved):
Approving this, but you should as well, @felixarntz, since I ended up changing more code than you. Hahah!

felixarntz (commented):
@JasonTheAdams Awesome! Really appreciate learning more gotchas about PHPStan from you :)

PR #149, opened 2025-12-12
Add provider awareness of their authentication method, decoupling from the assumption of all using API keys

### LINKED ISSUE #143: Add provider association with specific authentication method
We already have `RequestAuthenticationInterface`, with `ApiKeyRequestAuthentication` currently being the only implementation of that interface.

Since pretty much all cloud providers today require an API key for authentication, we don't need to expand the available implementations just yet.

However, what we should do is to properly set up an association between each provider and which authentication method they use. Today, all our code can simply just _assume_ that a cloud provider needs an API key, which is not a proper solution.

Implementing this should be pretty straightforward. Authentication classes can already provide context about its necessary input, via the `getJsonSchema()` method. They can also be dynamically instantiated without coupling things to a concrete constructor, via the `fromArray()` method.

## Proposed approach

- Add a new optional `authenticationMethod` field / key to `ProviderMetadata`, which will contain an identifier of an available authentication method.
    - For now, the only possible values will be `NONE` and `API_KEY`.
    - As reasonable defaults, use `API_KEY` if the provider type is `CLOUD`, and use `NONE` otherwise.
    - That being said, we should still update all provider implementations to explicitly set those values, as a good example.
- Add a way to get the authentication class for a given authentication identifier.
    - Either use an enum class, with a custom additional method to get the class name associated with the identifier.
    - Or implement an authentication registry where authentication classes can be registered with an identifier.
    - The latter is probably the proper long-term solution, but we could start with the former to keep things simple for now. It could be iterated on later without breaking changes.
- Update `ProviderRegistry::setRequestAuthenticationForProvider` to include a check in the beginning to ensure the given authentication instance is of the class expected by the provider. If not, throw an exception.
- Update `ProviderRegistry::createDefaultProviderRequestAuthentication` to look up which class is needed, instead of having `ApiKeyRequestAuthentication` hard-coded.

felixarntz:
Fixes #143


src/Providers/DTO/ProviderMetadata.php
```diff
@@ -60,13 +68,24 @@ class ProviderMetadata extends AbstractDataTransferObject
...
+        ?RequestAuthenticationMethod $authenticationMethod = null
+    ) {
         $this->id = $id;
         $this->name = $name;
         $this->type = $type;
         $this->credentialsUrl = $credentialsUrl;
+
+        if ($authenticationMethod === null && $type->isCloud()) {
+            $authenticationMethod = RequestAuthenticationMethod::apiKey();
+        }
```
JasonTheAdams: Are we just considering this a safe assumption?

src/Providers/Http/Enums/RequestAuthenticationMethod.php
```diff
@@ -0,0 +1,45 @@
...
+     *                                                                                            none.
+     * @phpstan-ignore missingType.generics
+     */
+    public function getImplementationClass(): ?string
+    {
+        if ($this->isApiKey()) {
+            return ApiKeyRequestAuthentication::class;
+        }
+
+        return null;
```
JasonTheAdams: Seems logically impossible to arrive here? How is this method nullable?

src/Providers/ProviderRegistry.php
```diff
@@ -443,11 +442,37 @@ private function setHttpTransporterForProvider(
...
+
+        if ($expectedClass === null) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Provider %s does not expect any authentication, but got %s.',
+                    $className,
+                    get_class($requestAuthentication)
+                )
+            );
+        }
```
JasonTheAdams, line 461: I'm a bit confused by this logic. We get the authentication method enum from the provider class, then we get the authentication class derived from the enum, null if the enum is null, and then we throw an exception if the authentication class is null saying that it _does not_ expect authentication but has an authentication method enum.

The only way this would get here would be if the `getImplementationClass()` method returned null, but that's logically impossible (see my other comment).

I also notice this down below, too. I guess it all hinges on whether the `getImplementationClass()` return type is in fact nullable or not.

JasonTheAdams (changes requested):
Nice work, @felixarntz! A couple points I'm a bit confused on.

src/Providers/DTO/ProviderMetadata.php
felixarntz: Fair point. I added this for backward compatibility, but I already started thinking yesterday "what if there's a cloud provider that doesn't need any authentication and therefore _wants_ this to be `null`?

Maybe we just get rid of it and always use `null` as a default? It's a BC break, but probably not that important given almost nobody is implementing providers.

src/Providers/Http/Enums/RequestAuthenticationMethod.php
felixarntz: Good point. I suppose for now this can unconditionally return `ApiKeyRequestAuthentication` because that's the only value for the enum. I'll update this.

src/Providers/ProviderRegistry.php
felixarntz: This clause is there because one shouldn't attempt to set a `RequestAuthentication` instance for a provider that doesn't need any authentication (i.e. where `authenticationMethod` is `null`).

I agree with your feedback that the `getImplementationClass()` return type shouldn't be nullable, so I'll update, and then the logic here has to be adjusted accordingly. But at its core the condition will still be needed - it'll instead just check whether the `authenticationMethod` is `null` instead of whether the `expectedClass` is `null`. Makes sense?

felixarntz:
@JasonTheAdams I've updated the PR based on the comments feedback.

JasonTheAdams (approved):

PR #152, opened 2025-12-22
Implement proper Google provider using its first-class API

### LINKED ISSUE #106: Use Google's primary Gemini API endpoints instead of its OpenAI compatibility layer
This is similar to #96, but for the Google AI provider.

felixarntz:
Fixes #106.

This unlocks significant benefits, such as finally being able to use Google's Nano Banana models for image generation, as well as additional supported parameters that weren't possible to support via the OpenAI API compatible implementation.

### DIFF
--- a/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
+++ b/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
@@ -4,28 +4,338 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\ProviderImplementations\Google;
 
+use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+use WordPress\AiClient\Common\Exception\RuntimeException;
+use WordPress\AiClient\Files\DTO\File;
+use WordPress\AiClient\Files\Enums\MediaOrientationEnum;
+use WordPress\AiClient\Messages\DTO\Message;
+use WordPress\AiClient\Messages\DTO\MessagePart;
+use WordPress\AiClient\Messages\Enums\MessageRoleEnum;
+use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModel;
 use WordPress\AiClient\Providers\Http\DTO\Request;
+use WordPress\AiClient\Providers\Http\DTO\Response;
 use WordPress\AiClient\Providers\Http\Enums\HttpMethodEnum;
-use WordPress\AiClient\Providers\OpenAiCompatibleImplementation\AbstractOpenAiCompatibleImageGenerationModel;
+use WordPress\AiClient\Providers\Http\Exception\ResponseException;
+use WordPress\AiClient\Providers\Http\Util\ResponseUtil;
+use WordPress\AiClient\Providers\Models\ImageGeneration\Contracts\ImageGenerationModelInterface;
+use WordPress\AiClient\Results\DTO\Candidate;
+use WordPress\AiClient\Results\DTO\GenerativeAiResult;
+use WordPress\AiClient\Results\DTO\TokenUsage;
+use WordPress\AiClient\Results\Enums\FinishReasonEnum;
 
 /**
  * Class for a Google image generation model.
  *
+ * This caters for Gemini models that can generate images as part of their multimodal output capabilities as well
+ * as for more traditional image generation models such as Imagen.
+ *
  * @since 0.1.0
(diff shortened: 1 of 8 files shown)


src/ProviderImplementations/Google/GoogleImageGenerationModel.php
```diff
@@ -4,28 +4,338 @@
...
+        return $this->parseResponseToGenerativeAiResult(
+            $response,
+            isset($params['parameters']) &&
+            is_array($params['parameters']) &&
+            isset($params['parameters']['outputOptions']) &&
+            is_array($params['parameters']['outputOptions']) &&
+            isset($params['parameters']['outputOptions']['mimeType']) &&
+            is_string($params['parameters']['outputOptions']['mimeType']) ?
+                $params['parameters']['outputOptions']['mimeType'] :
+                'image/png'
```
JasonTheAdams, line 96: So I dug into why PHPStan doesn't like this and it actually has nothing to do with it's ability to handled nested `isset()` checks, instead having to do with the lines below where arbitrary custom options are added to the `$params` array. Even if we set a type, PHPStan can't validate that the code matches the type because that code modifies it. Now, the runtime checks make it such that this is safe. It's just hard to follow from a type checking perspective.

There are options on how to go about this. The simplest, from looking at this, would actually be to not use this array here at all and instead to do: `$this->getConfig()->getOutputMimeType() ?? 'image/png'. From what I can tell, the `mimeType` array key is only set if that config provides it. A person _could_ set it by not including it in the config and then passing it as a "custom option", but this seems more like they're deviating from how it's supposed to work.

What do you think?

src/ProviderImplementations/Google/GoogleImageGenerationModel.php
```diff
@@ -4,28 +4,338 @@
...
+        foreach ($message->getParts() as $part) {
+            $text = $part->getText();
+            if ($text !== null) {
+                break;
+            }
+        }
+
+        if ($text === null) {
+            throw new InvalidArgumentException(
+                'The API only supports a single text message part as prompt.'
```
JasonTheAdams, line 215:
```suggestion
                'The API requires a single text message part as prompt.'
```

JasonTheAdams (approved):
Great work, @felixarntz! I left a couple minor suggestions, but they're not dealbreakers so I'll go ahead and approve! 👍

src/ProviderImplementations/Google/GoogleImageGenerationModel.php
felixarntz: I don't really like that approach, because it moves the dependency away to a transitive one instead of the correct _direct_ dependency:
* `params.parameters.outputOptions.mimeType` depends on `$this->getConfig()->getOutputMimeType()`.
* The value here depends on `params.parameters.outputOptions.mimeType`.
* Now, in practice the result will likely be the same no matter what you depend on, but there could be a future integration point that e.g. allows to filter `params` and then you would no longer have that guarantee.
* So it's a theoretical concern at the moment, but it's safe to say that relying on `$this->getConfig()->getOutputMimeType()` would make this code less error-proof than it is now.

I don't think we should trait actual code quality for a workaround just for PHPStan sake. I rather deal with annoying nested `isset`s then :)

JasonTheAdams: Totally get it. I'm with you there. The last option here, that I'd endorse, for typing would be to take the code for applying custom options and extract it to a private `applyCustomOptions` method which asserts the return type. It's really that code which is the most open-ended and skeeves out PHPStan in a reasonable way. This would isolate that code I think in a helpful way. I'm happy to do a commit for it if you'd like to show you what I mean. 😄

felixarntz: Sure! Please feel free to add it to this PR :)

JasonTheAdams: So I made the commit locally, but I can't push to this PR since the branch name is prefixed with "feature/" and therefore protected. 😅

felixarntz: Ahhh... oops

JasonTheAdams: Let me know how you'd like me to proceed. 😁 

We could always merge this and then I make a subsequent PR.

PR #153, opened 2025-12-22
Implement proper Anthropic provider using its first-class API

### LINKED ISSUE #107: Use Anthropic's primary API endpoints instead of its OpenAI compatibility layer
This is similar to #96, but for the Anthropic AI provider.

felixarntz:
Fixes #107.

This unlocks significant benefits, such as being able to use Anthropic's web search features, its recently introduced JSON output (boy, were they late to that party - but it's awesome!), as well as additional supported parameters that weren't possible to support via the OpenAI API compatible implementation.

### DIFF
--- a/src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
+++ b/src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
@@ -4,30 +4,516 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\ProviderImplementations\Anthropic;
 
+use Generator;
+use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+use WordPress\AiClient\Common\Exception\RuntimeException;
+use WordPress\AiClient\Messages\DTO\Message;
+use WordPress\AiClient\Messages\DTO\MessagePart;
+use WordPress\AiClient\Messages\Enums\MessagePartChannelEnum;
+use WordPress\AiClient\Messages\Enums\MessageRoleEnum;
+use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModel;
 use WordPress\AiClient\Providers\Http\DTO\Request;
+use WordPress\AiClient\Providers\Http\DTO\Response;
 use WordPress\AiClient\Providers\Http\Enums\HttpMethodEnum;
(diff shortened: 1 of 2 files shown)


src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
```diff
@@ -4,30 +4,516 @@
...
+                ($usage['cache_creation_input_tokens'] ?? 0) +
+                ($usage['cache_read_input_tokens'] ?? 0);
+
+            $tokenUsage = new TokenUsage(
+                $inputTokens,
+                $usage['output_tokens'] ?? 0,
+                $inputTokens + ($usage['output_tokens'] ?? 0)
+            );
+        } else {
+            $tokenUsage = new TokenUsage(0, 0, 0);
```
JasonTheAdams, line 445: Hmm. This feels odd as it would be difficult downstream to know the difference between zeroes as values versus token usage not being provided. I feel like we should do two things:
1. Make this a nullable parameter of `GenerativeAiResult`
2. Make the `TokenUsage` parameters nullable in case some systems provide some usage information but not all.

This way a downstream system knows the difference between unprovided data and real values.

JasonTheAdams (changes requested):
Looks good overall! One question and a suggestion.

My question is whether the API URL should be changed for the provider or if it's actually exactly the same? I was surprised to not see that.

src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
felixarntz: @JasonTheAdams Can you raise a separate issue for that? We already have existing logic where if the response is missing the relevant data, we set the values to `0`.

I agree with you that supporting `null` values for those cases would be better. But since that's not a problem this PR technically introduces, I think it's best to wave it through here and address holistically in another issue and PR.

felixarntz:
@JasonTheAdams 
> My question is whether the API URL should be changed for the provider or if it's actually exactly the same? I was surprised to not see that.

This is not necessary. It's the same API base URL, it's only the specific route that changes, which this PR does.

src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
JasonTheAdams: Sure! Created #158

PR #156, opened 2025-12-30
Implement proper Anthropic provider using its first-class API

### LINKED ISSUE #107: Use Anthropic's primary API endpoints instead of its OpenAI compatibility layer

felixarntz:
Fixes #107

Replaces #153 (which was accidentally using the protected `feature/*` branch prefix)


felixarntz:
@JasonTheAdams Can you give this a re-review? It's the same as #153, but from a proper branch. I've replied to your feedback in #153, I don't think any updates are needed based on that.

JasonTheAdams (approved):
LGTM! 🚀

PR #157, opened 2025-12-30
Adds caching support via PSR-16

### LINKED ISSUE #141: Add support for caching using PSR-16
There are moments where Providers will want to cache results and not make a request every time. An example of this is that our current Provider implementations request the list of models for every fulfilled prompt. This means every prompt is at least two round trips when it could be one in subsequent calls.

We should implement the [PSR-6](https://www.php-fig.org/psr/psr-6/) interfaces to make this work. This allows any system implementing this to provide its own caching layer.

I did consider the [PSR-16](https://www.php-fig.org/psr/psr-16/) route, especially since maps pretty well to the `wp_cache_*` functions in WordPress, for us in the WP AI Client, but in the spirit of keeping this more framework agnostic, I think we should use PSR-6. It's really not much more complex, but allows for more verbose caching systems.

JasonTheAdams:
Resolves #141 

This PR adds optional PSR-16 caching support to the AI Client. When a cache is configured, model metadata from provider APIs (like the list of available models) is stored and reused instead of making repeated API calls. This improves performance and reduces API usage, especially useful in applications that frequently initialize the client or query model availability.

  What's new:
  - AiClient::setCache() and AiClient::getCache() for configuring a PSR-16 cache
  - psr/simple-cache added as a dependency
  - AbstractOpenAiCompatibleModelMetadataDirectory now caches /models API responses
  - Cache keys are unique per provider to prevent data conflicts
  - MockCache test helper for verifying cache behavior

### DIFF
--- a/src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
+++ b/src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\Providers\ApiBasedImplementation;
 
+use WordPress\AiClient\AiClient;
+use WordPress\AiClient\Common\Contracts\CachesDataInterface;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+use WordPress\AiClient\Common\Traits\WithDataCachingTrait;
 use WordPress\AiClient\Providers\Contracts\ModelMetadataDirectoryInterface;
 use WordPress\AiClient\Providers\Http\Contracts\WithHttpTransporterInterface;
 use WordPress\AiClient\Providers\Http\Contracts\WithRequestAuthenticationInterface;
@@ -20,15 +23,21 @@ use WordPress\AiClient\Providers\Models\DTO\ModelMetadata;
 abstract class AbstractApiBasedModelMetadataDirectory implements
     ModelMetadataDirectoryInterface,
     WithHttpTransporterInterface,
-    WithRequestAuthenticationInterface
+    WithRequestAuthenticationInterface,
+    CachesDataInterface
 {
     use WithHttpTransporterTrait;
     use WithRequestAuthenticationTrait;
+    use WithDataCachingTrait;
 
     /**
-     * @var ?array<string, ModelMetadata> Map of model ID to model metadata, effectively for caching.
+     * The cache key suffix for the models list.
+     *
+     * @since n.e.x.t
+     *
+     * @var string
      */
-    private ?array $modelMetadataMap = null;
+    private const MODELS_CACHE_KEY = 'models';
 
     /**
      * {@inheritDoc}
@@ -77,10 +86,32 @@ abstract class AbstractApiBasedModelMetadataDirectory implements
      */
     private function getModelMetadataMap(): array
     {
-        if ($this->modelMetadataMap === null) {
-            $this->modelMetadataMap = $this->sendListModelsRequest();
-        }
-        return $this->modelMetadataMap;
+        /** @var array<string, ModelMetadata> */
+        return $this->cached(
+            self::MODELS_CACHE_KEY,
+            fn () => $this->sendListModelsRequest(),
+            86400
+        );
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    protected function getCachedKeys(): array
+    {
+        return [self::MODELS_CACHE_KEY];
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    protected function getBaseCacheKey(): string
+    {
+        return 'ai_client_' . AiClient::VERSION . '_' . md5(static::class);
     }
 
     /**
--- a/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
+++ b/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
@@ -81,7 +81,7 @@ abstract class AbstractOpenAiCompatibleModelMetadataDirectory extends AbstractAp
      */
     protected function getCacheKey(): string
     {
-        return 'ai_client_models_' . md5(static::class);
+        return 'ai_client_models_' . AiClient::VERSION . '_' . md5(static::class);
     }
 
     /**
--- a/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
+++ b/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\Providers\OpenAiCompatibleImplementation;
 
+use WordPress\AiClient\AiClient;
 use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModelMetadataDirectory;
 use WordPress\AiClient\Providers\Http\DTO\Request;
 use WordPress\AiClient\Providers\Http\DTO\Response;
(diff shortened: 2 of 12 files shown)


src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -44,6 +60,61 @@ protected function sendListModelsRequest(): array
...
+     * The cache key is unique per child class to ensure different providers
+     * don't share cached model data.
+     *
+     * @since n.e.x.t
+     *
+     * @return string The cache key.
+     */
+    protected function getCacheKey(): string
+    {
+        return 'ai_client_models_' . md5(static::class);
```
felixarntz: It would be great if we could include the current AI Client version in the cache key. A version update may come with an updated DTO shape at some point, so it's always safest to invalidate all caches upon update. Or there may be a fix in a specific `ModelMetadataDirectory` implementation.

This was relevant quite a few times while I worked on AI Services, so I suggest we incorporate that here too.

We don't store the current version anywhere in the codebase, but it's easy to do that in a maintainable way. We could have a specific constant somewhere which we replace its value for as part of the release process. It could even be via a simple vibe coded script.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -44,6 +60,61 @@ protected function sendListModelsRequest(): array
         foreach ($modelsMetadataList as $modelMetadata) {
             $modelMetadataMap[$modelMetadata->getId()] = $modelMetadata;
         }
+
+        // Store in cache.
+        if ($cache !== null) {
+            $cache->set($cacheKey, $this->dehydrateModelMetadataMap($modelMetadataMap));
```
felixarntz: What's the TTL? We'll need a mechanism for the cache to invalidate.

My suggestion is to use a mix of both:
- Set a reasonable TTL, e.g. for a couple hours, maybe a day.
- Add some public method to invalidate the cache. We could introduce an interface like `WithCachedDataInterface` that has a single method `invalidateCaches()` that would invalidate all relevant caches for the class. And our `ModelMetadataDirectory` implementations for cloud providers could make use of that.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -20,6 +21,8 @@
  * providers that have adopted OpenAI's models API specification as a standard interface.
  *
  * @since 0.1.0
+ *
+ * @phpstan-import-type ModelMetadataArrayShape from ModelMetadata
  */
 abstract class AbstractOpenAiCompatibleModelMetadataDirectory extends AbstractApiBasedModelMetadataDirectory
```
felixarntz, line 24: Note: As we have merged and are further merging the updated provider implementations using their primary APIs, we will also need to apply a similar solution to the other `ModelMetadataDirectory` implementations.

Related to my other comment about a `WithCachedDataInterface`, maybe we could move some of the logic to a `WithCachedDataTrait`, so that it's fewer code to duplicate?

felixarntz (commented):
@JasonTheAdams Looks great! A few additional points to consider, but this is close.

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
JasonTheAdams: I added an `AiClient::VERSION`. It seemed like a sensible place to put it since this a client package and that's the main class. How does that sound?

Tentatively resolved in 5ccaee3

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
JasonTheAdams: I added an interface and trait in 6d0e4f7

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
JasonTheAdams: I updated the trait in 3981997 to handle the `invalidateCaches()` method.

I also set a 1-day TTL. 👍

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -20,16 +23,36 @@
...
  *
  * @since 0.1.0
+ *
+ * @phpstan-import-type ModelMetadataArrayShape from ModelMetadata
  */
-abstract class AbstractOpenAiCompatibleModelMetadataDirectory extends AbstractApiBasedModelMetadataDirectory
+abstract class AbstractOpenAiCompatibleModelMetadataDirectory extends AbstractApiBasedModelMetadataDirectory implements
+    CachesDataInterface
 {
+    use WithDataCachingTrait;
```
felixarntz: Wouldn't it make more sense to implement the interface and use the trait in `AbstractApiBasedModelMetadataDirectory`?

* That's the class likely any cloud based provider (i.e. where caching responses is relevant) will use.
* `AbstractOpenAiCompatibleModelMetadataDirectory` is currently used by all our model metadata directories, but it's much more specific.

Additionally, the way it's implemented, it's easy to accidentally get rid of the caching by overriding `sendListModelsRequest`. I would suggest to implement the cache lookup in `AbstractApiBasedModelMetadataDirectory::getModelMetadataMap()` --> more broadly reusable, and private, so impossible to bypass, which I think makes sense here.

felixarntz (commented):
@JasonTheAdams Almost good to go, only one important architectural consideration.

felixarntz:
@JasonTheAdams Since you're adding the `VERSION` constant here, I thought it'd be a good idea to document the prepublish requirements, as basic as they are, since now they are becoming _slightly_ less basic. See #172.

(In the future it would be great to automate this, but I think for now this is useful to at least document.)

src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
JasonTheAdams: Good idea! Resolved in d6b1a6f

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -77,10 +86,32 @@ final public function getModelMetadata(string $modelId): ModelMetadata
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
+     */
+    protected function getBaseCacheKey(): string
+    {
+        return 'ai_client_' . AiClient::VERSION . '_' . md5(static::class);
     }
```
felixarntz, line 115: Why not put this in the trait? This implementation here does nothing that wouldn't be suitable for the trait itself - it's not tied to this specific class.

felixarntz (approved):
@JasonTheAdams LGTM, just one thing that I think is worth improving. Please check before merging, feel free to merge after updating, or let me know if you disagree.

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
JasonTheAdams: I thought about this, but I didn't want to assume that we always need the class on the end of the key. We really only need it here because this is an abstract class. But another class may not want that in the key, which would also also keys to be shared between contexts.

felixarntz: Agreed, it doesn't _always_ need to be like that, but a protected method would still allow changing it. So IMO having this implementation in the trait as "works for ~90% of cases" would still be helpful.

But not a big deal, we can also iterate later, as changing that would not break anything.

PR #155, opened 2025-12-30
Implement proper Google provider using its first-class API

felixarntz:
Fixes #106

Replaces #152 (which was accidentally using the protected `feature/*` branch prefix)

PR #159, opened 2025-12-30
Fix proper Google API implementation

felixarntz:
Follow up to #155.

I hadn't yet tested things when it was merged, so this PR has the fixes needed to actually make things work.

**Note:** A workaround is included to return the image generation specific class when using Gemini multimodal output models that _primarily_ are used for image generation. This is only since we don't have model class implementations yet that are actually multimodal. Let's discuss in a separate issue what's the best way to go about that, for a proper solution. Doesn't have to block this work.

### DIFF
--- a/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
+++ b/src/ProviderImplementations/Google/GoogleImageGenerationModel.php
@@ -12,6 +12,8 @@ use WordPress\AiClient\Messages\DTO\Message;
 use WordPress\AiClient\Messages\DTO\MessagePart;
 use WordPress\AiClient\Messages\Enums\MessageRoleEnum;
 use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModel;
+use WordPress\AiClient\Providers\Http\Contracts\RequestAuthenticationInterface;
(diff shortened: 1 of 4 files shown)


felixarntz:
Quick experiment I did to test this:

### Imagen 4 image generation

```bash
php cli.php 'Photo of a tricolor Cavalier King Charles Spaniel on an airfield in the desert of Peru' --outputFormat=image-base64 --providerId=google
```

Output:
cavalier-imagen-4-generate-001

### Gemini Nano Banana image generation

```bash
php cli.php 'Photo of a tricolor Cavalier King Charles Spaniel on an airfield in the desert of Peru' --outputFormat=image-base64 --providerId=google --modelId=gemini-2.5-flash-image
```

Output:
cavalier-gemini-2.5-flash-image

### Summary

Confirms what's widely known: multimodal output models that do image generation understand things much better. Both images look solid, but only the Gemini Nano Banana image has a _**tricolor**_ Cavalier King Charles Spaniel, like I asked for in both cases :)

The other thing is that those models create more realistic looking images, while classic diffusion models create more "artsy" images. The depth of field in the Imagen-generated image is way too extreme - it looks cool, but not realistic.

JasonTheAdams (approved):
Glad you tested! Let me know which Issue you open to discuss multi-modal models.

Issue #160, opened 2025-12-30
Implement proper multimodal output model classes


felixarntz:
See #159: For now, Google is the only provider with multimodal output models that we have implemented support for (since other multimodal output models are for audio generation, which we generally don't support yet).

In general, we need to figure out the best approach for implementing classes for those models. Technically, they are pretty much using the exact implementation that our current text generation model class implementations rely on. But because they support both text generation and image generation, we need a class that implements both `TextGenerationModelInterface` and `ImageGenerationModelInterface`.

Looking at the Google implementation, we could combine both classes `GoogleTextGenerationModel` and `GoogleImageGenerationModel` into one, but that IMO would make things very messy and crowded. So my initial suggestion would be to leave those classes as is and instead introduce another class like `GoogleTextAndImageGenerationModel`, which simply forwards to a new instance of the respective other one, depending on which method is called. This would be almost similar to how currently `GoogleImageGenerationModel` will instantiate a `GoogleTextGenerationModel` for Gemini multimodal image output models.

felixarntz:
For a more close-to-code perspective: What we'll want to fix with this is https://github.com/WordPress/php-ai-client/blob/trunk/src/ProviderImplementations/Google/GoogleProvider.php#L45.

JasonTheAdams:
Hmm, this is really interesting. I'd like to think about this from the DX perspective, too. Thinking on this for a moment:

1. Use `PromptBuilder::generateTextResult()`
2. Use `PromptBuilder::generateImageResult()`
3. Use `PromptBuilder::generateResult()`

The first two set the capability intended for the result, while the latter accepts a capability or none at all. It looks like it assumes one capability at a time. So I think we'd need to also work here to support multiple capabilities. Otherwise, if none is provided it would just grab the first interface and assume that's it.

I think my inclination is to start here, as the point of DX entry, and then determine how that influences the model design.

felixarntz:
The more I think about it, the more I think that in the majority of cases, the concept of capabilities as a whole is pointless. The capability of "text generation" for example means nothing more than "output modality is or includes text", similarly "image generation" means "output modality is or includes image".

The problem, as we've identified several times before, is that it's not always as simple, for example "speech generation" and "text to speech conversion" are distinct capabilities that both imply "output modality is or includes audio". So we can't just remove the concept of capabilities, at least not without partially replacing it.

[...]

PR #161, opened 2025-12-30
Updates OpenAI to use Responses API

### LINKED ISSUE #96: Use OpenAI Responses API instead of the Chat Completions API
OpenAI has introduced the **Responses API** (`POST /v1/responses`) as the new, unified endpoint that consolidates functionality from both **Chat Completions** and **Assistants/Threads** APIs.

### Advantages of using the Responses API

* **Unified design**: One endpoint for chat, tools, and multimodal interactions.
* **Built-in memory**: Supports `store:true` and `conversation.id`, which means conversations can persist without resending the full message history. This can significantly reduce token usage and costs for long-running chats.
* **Native tool handling**: Cleaner definition of tools and structured `tool_call` outputs, easier to validate and safer than current function-calling prompts.
* **Multimodal ready**: Accepts text, images, and audio in the same request.
* **Consistent output**: Always returns a `response` object with `output[]`, `usage`, `status`, etc., which simplifies logging and debugging.
* **Future direction**: OpenAI is evolving the Responses API, while Chat Completions is effectively a legacy path for new projects.

### Disadvantages / Migration Considerations

* **Newer API**: Responses API is less mature; not all SDKs or plugins in the ecosystem support it fully yet.
* **Learning curve**: Output parsing is slightly more complex (`output[]` vs. `choices[0].message.content`).
* **Migration cost**: Projects already using Chat Completions will need refactoring to adapt request/response handling.
* **Community ecosystem**: Most tutorials and sample code still target Chat Completions.

### Proposal

Evaluate adding support for the **Responses API** in this client library, either:

1. As a replacement for Chat Completions, or
2. As an additional option, allowing developers to choose between `chat.completions` and `responses`.

This would ensure the library stays aligned with OpenAI’s current API direction and provides WordPress developers with a future-proof integration option.

JasonTheAdams:
Resolves #96 

This PR migrates the OpenAI provider implementation from the legacy Chat Completions API (/v1/chat/completions) to the newer Responses API (/v1/responses) for text generation and Images API (/v1/images/generations) for image generation. The OpenAiTextGenerationModel and OpenAiImageGenerationModel classes have been completely rewritten to extend AbstractApiBasedModel directly, following the same implementation pattern used by the Google and Anthropic providers. The OpenAiCompatible helper classes remain available for other providers that use OpenAI-compatible APIs.

### What's New

  - Built-in web search tool — GPT models can now use OpenAI's native web_search tool via the webSearch config option
  - Built-in code interpreter — Access the code_interpreter tool via customOptions for code execution capabilities
  - Continuing a previous chat — Use `previous_response_id` with a previous response id to continue the conversation
  - Responses API for image generation — gpt-image-* models now use the Responses API with the image_generation tool, providing a consistent API surface
  - Structured output support — JSON schema output via the text.format parameter
  - Function calling — Full support for custom function declarations with the new output format
  - Comprehensive test coverage — 39 new unit tests covering both text and image generation models
  - Document support — the text generation models can receive documents (e.g. PDFs) now

The image generation tool means the output is multi-modal, so for now I'm not adding support for this. We're discussing a solution in #160 and then can circle back on adding support for this.

### Bonus

I updated the `cli.php` to support stdin and file references. This was useful for testing having it examine an image and describe it to me (multi-modal input).

### Testing

I tested out the text generation as well as image generation, web search tool, code interpretation, and making an accurate picture of my crazy dog jumping out of a helicopter:

image

### DIFF
--- a/src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
+++ b/src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
@@ -4,23 +4,14 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\ProviderImplementations\OpenAi;
 
-use WordPress\AiClient\Common\Exception\InvalidArgumentException;
-use WordPress\AiClient\Files\DTO\File;
 use WordPress\AiClient\Files\Enums\MediaOrientationEnum;
-use WordPress\AiClient\Messages\DTO\Message;
-use WordPress\AiClient\Messages\DTO\MessagePart;
-use WordPress\AiClient\Messages\Enums\MessageRoleEnum;
-use WordPress\AiClient\Providers\ApiBasedImplementation\AbstractApiBasedModel;
 use WordPress\AiClient\Providers\Http\DTO\Request;
 use WordPress\AiClient\Providers\Http\DTO\Response;
 use WordPress\AiClient\Providers\Http\Enums\HttpMethodEnum;
 use WordPress\AiClient\Providers\Http\Exception\ResponseException;
-use WordPress\AiClient\Providers\Http\Util\ResponseUtil;
-use WordPress\AiClient\Providers\Models\ImageGeneration\Contracts\ImageGenerationModelInterface;
-use WordPress\AiClient\Results\DTO\Candidate;
+use WordPress\AiClient\Providers\OpenAiCompatibleImplementation\AbstractOpenAiCompatibleImageGenerationModel;
 use WordPress\AiClient\Results\DTO\GenerativeAiResult;
 use WordPress\AiClient\Results\DTO\TokenUsage;
-use WordPress\AiClient\Results\Enums\FinishReasonEnum;
 
 /**
  * Class for an OpenAI image generation model using the Images API.
@@ -30,133 +21,125 @@ use WordPress\AiClient\Results\Enums\FinishReasonEnum;
  *
  * @since n.e.x.t
  *
- * @phpstan-type ImageData array{
- *     b64_json?: string,
- *     url?: string,
- *     revised_prompt?: string
- * }
- * @phpstan-type ResponseData array{
+ * @phpstan-type ImageResponseData array{
  *     created?: int,
- *     data?: list<ImageData>
+ *     data?: list<array{url?: string, b64_json?: string}>
  * }
  */
-class OpenAiImageGenerationModel extends AbstractApiBasedModel implements ImageGenerationModelInterface
+class OpenAiImageGenerationModel extends AbstractOpenAiCompatibleImageGenerationModel
 {
     /**
      * {@inheritDoc}
      *
      * @since n.e.x.t
      */
-    public function generateImageResult(array $prompt): GenerativeAiResult
-    {
-        $httpTransporter = $this->getHttpTransporter();
-
-        $params = $this->prepareGenerateImageParams($prompt);
-
-        $request = new Request(
-            HttpMethodEnum::POST(),
-            OpenAiProvider::url('images/generations'),
-            ['Content-Type' => 'application/json'],
-            $params,
+    protected function createRequest(
+        HttpMethodEnum $method,
+        string $path,
+        array $headers = [],
+        $data = null
+    ): Request {
+        return new Request(
+            $method,
+            OpenAiProvider::url($path),
+            $headers,
+            $data,
             $this->getRequestOptions()
         );
-
-        // Add authentication credentials to the request.
-        $request = $this->getRequestAuthentication()->authenticateRequest($request);
-
-        // Send and process the request.
-        $response = $httpTransporter->send($request);
-        ResponseUtil::throwIfNotSuccessful($response);
-        return $this->parseResponseToGenerativeAiResult($response);
     }
 
     /**
-     * Prepares the given prompt and the model configuration into parameters for the API request.
+     * {@inheritDoc}
      *
      * @since n.e.x.t
-     *
-     * @param list<Message> $prompt The prompt to generate an image for. Should be a single user message.
-     * @return array<string, mixed> The parameters for the API request.
      */
     protected function prepareGenerateImageParams(array $prompt): array
     {
-        $config = $this->getConfig();
+        $params = parent::prepareGenerateImageParams($prompt);
         $modelId = $this->metadata()->getId();
 
-        $params = [
-            'model' => $modelId,
-            'prompt' => $this->preparePromptParam($prompt),
-        ];
-
-        // Add size configuration if available.
-        $outputMediaOrientation = $config->getOutputMediaOrientation();
-        $outputMediaAspectRatio = $config->getOutputMediaAspectRatio();
-        if ($outputMediaOrientation !== null || $outputMediaAspectRatio !== null) {
-            $params['size'] = $this->prepareSizeParam($modelId, $outputMediaOrientation, $outputMediaAspectRatio);
-        }
-
-        // Add model-specific parameters.
+        // GPT image models use output_format, DALL-E uses response_format.
         if ($this->isGptImageModel($modelId)) {
-            $this->addGptImageModelParams($params, $config);
-        } else {
-            $this->addDalleModelParams($params);
-        }
-
-        /*
-         * Any custom options are added to the parameters as well.
-         * This allows developers to pass other options that may be more niche or not yet supported by the SDK.
-         */
-        $customOptions = $config->getCustomOptions();
-        foreach ($customOptions as $key => $value) {
-            if (isset($params[$key])) {
-                throw new InvalidArgumentException(
-                    sprintf(
-                        'The custom option "%s" conflicts with an existing parameter.',
-                        $key
-                    )
-                );
-            }
-            $params[$key] = $value;
+            // For GPT image models, convert response_format to the appropriate format.
+            // The parent sets response_format, but GPT models don't need it.
+            unset($params['response_format']);
         }
 
         return $params;
     }
 
     /**
-     * Adds GPT image model specific parameters to the request.
+     * {@inheritDoc}
      *
      * @since n.e.x.t
-     *
-     * @param array<string, mixed> $params The parameters array to modify.
-     * @param \WordPress\AiClient\Providers\Models\DTO\ModelConfig $config The model configuration.
      */
-    protected function addGptImageModelParams(array &$params, $config): void
+    protected function prepareSizeParam(?MediaOrientationEnum $orientation, ?string $aspectRatio): string
     {
-        // Add output format configuration if available.
-        $outputMimeType = $config->getOutputMimeType();
-        if ($outputMimeType !== null) {
-            $formatMap = [
-                'image/png' => 'png',
-                'image/jpeg' => 'jpeg',
-                'image/webp' => 'webp',
-            ];
-            if (isset($formatMap[$outputMimeType])) {
-                $params['output_format'] = $formatMap[$outputMimeType];
-            }
+        $modelId = $this->metadata()->getId();
+
+        if ($this->isGptImageModel($modelId)) {
+            return $this->prepareGptImageSizeParam($orientation, $aspectRatio);
         }
+
+        return $this->prepareDalleSizeParam($modelId, $orientation, $aspectRatio);
     }
 
     /**
-     * Adds DALL-E model specific parameters to the request.
+     * {@inheritDoc}
+     *
+     * Overrides the parent to handle OpenAI's `created` timestamp instead of `id`.
      *
      * @since n.e.x.t
-     *
-     * @param array<string, mixed> $params The parameters array to modify.
      */
-    protected function addDalleModelParams(array &$params): void
-    {
-        // DALL-E models need response_format set to b64_json to get base64 data.
-        $params['response_format'] = 'b64_json';
+    protected function parseResponseToGenerativeAiResult(
+        Response $response,
+        string $expectedMimeType = 'image/png'
+    ): GenerativeAiResult {
+        /** @var ImageResponseData $responseData */
+        $responseData = $response->getData();
+
+        if (!isset($responseData['data']) || !$responseData['data']) {
+            throw ResponseException::fromMissingData($this->providerMetadata()->getName(), 'data');
+        }
+        if (!is_array($responseData['data']) || !array_is_list($responseData['data'])) {
+            throw ResponseException::fromInvalidData(
+                $this->providerMetadata()->getName(),
+                'data',
+                'The value must be an indexed array.'
+            );
+        }
+
+        $candidates = [];
+        foreach ($responseData['data'] as $index => $choiceData) {
+            if (!is_array($choiceData) || array_is_list($choiceData)) {
+                throw ResponseException::fromInvalidData(
+                    $this->providerMetadata()->getName(),
+                    "data[{$index}]",
+                    'The value must be an associative array.'
+                );
+            }
+
+            $candidates[] = $this->parseResponseChoiceToCandidate($choiceData, $index, $expectedMimeType);
+        }
+
+        // The Images API returns `created` timestamp instead of `id`.
+        $id = isset($responseData['created']) ? 'img-' . $responseData['created'] : '';
+
+        // The Images API doesn't return token usage.
+        $tokenUsage = new TokenUsage(0, 0, 0);
+
+        // Use any other data from the response as provider-specific response metadata.
+        $additionalData = $responseData;
+        unset($additionalData['data'], $additionalData['created']);
+
+        return new GenerativeAiResult(
+            $id,
+            $candidates,
+            $tokenUsage,
+            $this->providerMetadata(),
+            $this->metadata(),
+            $additionalData
+        );
     }
 
     /**
@@ -170,67 +153,6 @@ class OpenAiImageGenerationModel extends AbstractApiBasedModel implements ImageG
     protected function isGptImageModel(string $modelId): bool
     {
         return str_starts_with($modelId, 'gpt-image-');
-    }
-
-    /**
-     * Prepares the prompt parameter for the API request.
-     *
-     * @since n.e.x.t
-     *
-     * @param list<Message> $messages The messages to prepare. Should be a single user message.
-     * @return string The prepared prompt string.
-     */
-    protected function preparePromptParam(array $messages): string
-    {
-        if (count($messages) !== 1) {
-            throw new InvalidArgumentException(
-                'The API requires a single user message as prompt.'
-            );
-        }
-        $message = $messages[0];
-        if (!$message->getRole()->isUser()) {
-            throw new InvalidArgumentException(
-                'The API requires a user message as prompt.'
-            );
-        }
-
-        $text = null;
-        foreach ($message->getParts() as $part) {
-            $text = $part->getText();
-            if ($text !== null) {
-                break;
-            }
-        }
-
-        if ($text === null) {
-            throw new InvalidArgumentException(
-                'The API requires a text message part as prompt.'
-            );
-        }
-
-        return $text;
-    }
-
-    /**
-     * Prepares the size parameter for the image generation request.
-     *
-     * @since n.e.x.t
-     *
-     * @param string $modelId The model ID.
-     * @param MediaOrientationEnum|null $orientation The desired media orientation.
-     * @param string|null $aspectRatio The desired media aspect ratio.
-     * @return string The size parameter value.
-     */
-    protected function prepareSizeParam(
-        string $modelId,
-        ?MediaOrientationEnum $orientation,
-        ?string $aspectRatio
-    ): string {
-        if ($this->isGptImageModel($modelId)) {
-            return $this->prepareGptImageSizeParam($orientation, $aspectRatio);
-        }
-
-        return $this->prepareDalleSizeParam($modelId, $orientation, $aspectRatio);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -4,48 +4,386 @@
...
+use WordPress\AiClient\Results\DTO\GenerativeAiResult;
+use WordPress\AiClient\Results\DTO\TokenUsage;
+use WordPress\AiClient\Results\Enums\FinishReasonEnum;
 
 /**
- * Class for an OpenAI image generation model.
+ * Class for an OpenAI image generation model using the Responses API.
  *
- * @since 0.1.0
+ * This uses the Responses API with the built-in image_generation tool.
```
felixarntz: I didn't expect to see the Responses API being used for image generation. Given you did that, I assume it's possible, also given your example, but is that the case for all image generation models? Have you tried the implementation with the older models like `dall-e-3`? Just want to make sure they can also be used with the Responses API.

My original expectation was that we would only need to update the `OpenAiTextGenerationModel` to use the Responses API.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -4,48 +4,386 @@
...
+        $modelId = $this->metadata()->getId();
+
+        // The Responses API with image_generation tool requires a model that supports it.
+        // We use a capable model like gpt-4o to process the request with the image_generation tool.
+        $params = [
+            'model' => $this->getHostModelForImageGeneration($modelId),
+            'input' => $this->preparePromptParam($prompt),
+            'tools' => [
+                $this->prepareImageGenerationTool($modelId),
+            ],
```
felixarntz: I'm not sure how to best handle this, I think it needs some more thought.

* Do we handle this under the hood? If so, I think this is technically pretty much a multimodal output implementation, right? It still primarily generates text, but can generate images via the tool.
* Or do we expose this explicitly as a tool? If so, it would be similar in concept to e.g. `WebSearch`, which maps to a provider-specific tool that some providers support.

Given that actually another model generates the image than the one that you prompt, I feel like the latter would be more appropriate. This then leads me to think: Why don't we just implement _all_ of it as part of `OpenAiTextGenerationModel` and leave `OpenAiImageGenerationModel` untouched as purely a traditional image generation model? No need for that to use the Responses API.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -4,48 +4,386 @@
...
+     * @param string $modelId The requested model ID.
+     * @return string The host model ID to use for the request.
+     */
+    protected function getHostModelForImageGeneration(string $modelId): string
+    {
+        // If this is a gpt-image-* model, we need a host model to run the tool.
+        // Otherwise, the model itself can host the image_generation tool.
+        if (str_starts_with($modelId, 'gpt-image-')) {
+            return 'gpt-4o';
+        }
```
felixarntz: Related to the above, this is not great. Since it's a tool and not multimodal output, I think we should expose it as a tool.

Or, at the very least, we should continue to offer an actual image generation model to prompt directly, via the existing `images/generations` endpoint from `AbstractOpenAiCompatibleImageGenerationModel`, without going through the Responses API - which for the scenario of purely image generation seems like an unnecessary proxy in between.

felixarntz (changes requested):
@JasonTheAdams Thanks for working on this!

I haven't done a full review yet, but my early feedback focuses mostly on concerns with the image generation implementation. I'm not convinced why that needs to use the Responses API.

JasonTheAdams:
@felixarntz So you're completely reasonable to wonder why I'm using the Responses API and not the Images API. After **much** research, I come out the other side wondering if OpenAI could have made it any more convoluted. 😂 😭 

I'm going to write this out so we're on the same page, though I'm sure you know a good bit of it.

So images can be generated using both the Images API and the Responses API. The Responses API is, as a whole, the more capable API, providing things like message history, caching (to help with cost), stateful history, multimodal output, and so forth. The only thing it can't currently do is use `gpt-image-1.5` (only v1). For all intents and purposes, it's the more capable API. But it has trade-offs:

* Image generation is a tool of the base model. So you can't actually use `gpt-image-1` directly; you have to use something like `gpt-5` which then calls the image model. This means you effectively need to specify two models.
* The image prompt is adjusted by the base model. So if you put "a king charles spaniel" the base model will attempt to rewrite that into a "better" image prompt. This may be useful, but it may also adjust it in a surprising way.
* As mentioned, you can't use the `gpt-image-1.5` model yet.

It's weird. I don't love my first pass, but it does work. I'm thinking of making a decision matrix so it conditionally uses the most "optimal" API. Something like:

```mermaid
flowchart TD
  A([Start: intent = generateImage]) --> B{Has message history}
  B -- Yes --> R[Use Responses API]
  B -- No --> C{Has previous_response_id}
  C -- Yes --> R
  C -- No --> D{Specified model is non-image}
  D -- Yes --> R
  D -- No --> E{Needs multimodal output}
  E -- Yes --> R
  E -- No --> F{Needs pre-render reasoning}
  F -- Yes --> R
  F -- No --> I[Use Images API]
```

I'd also consider adding a custom option so someone can specify the base and/or image model if the Responses API is used.

What do you think?

felixarntz:
@JasonTheAdams Thanks for outlining this in so much detail! Your rationale makes a lot of sense to me.

Two follow up questions:
1. "Has previous_response_id" - what does this mean? What is a "previous response ID"?
2. "Specified model is non-image" - unless we intentionally make that possible, right now it wouldn't be. We wouldn't assign a non-image model the image generation capability, at least not now. And I'm not sure why that would make sense. If we want to enable someone to use the Responses API for image generation, I think that should rather be a conscious decision where you can specify the image generation model as usual, but then through some other more advanced option the "host" model for the Responses API.

For now, as of for this PR, I would suggest to keep things simple for image generation and stick with only the Images API: While the Responses API is better in many ways, it's also a lot more complex to think through and properly implement for the image generation use-case, and while most branches of the decision tree you shared end up in the Responses API, I think in 95% of actual usage it'll end up in the one branch that goes to the Images API.

The most valuable benefit of the Responses API for image generation IMO is that it can deal with messages history.

But we can add support for that separately, let's open an issue. The purpose of this PR is primarily to migrate `chat/completions` usage to the Responses API, since for that the latter is clearly recommended, so I'd say let's leave image generation (almost?) untouched.

JasonTheAdams:
> "Has previous_response_id" - what does this mean? What is a "previous response ID"?

The Responses API supports stateful prompting. You can connect it with the Conversations API or [use this previous response ID](https://platform.openai.com/docs/guides/conversation-state#passing-context-from-the-previous-response).

In our system, the response ID would be sent back as "additional data" in the `GenerativeAiResult`, which could then be passed as a custom config option.

> "Specified model is non-image"

I see what you mean. I'd be fine with that for now. 👍 

I like the idea of keeping this simple for images and opening an Issue afterwards. I'll update this to go that route!

JasonTheAdams:
Back to you, @felixarntz! I switched the Image Generation portion to using the Images API. Here's some fun tests of "A happy Belgian Malinois dog playing in a sunny field":

**Model: `gpt-image-1`**
image

**Model: `dall-e-3`**
image

JasonTheAdams:
I did add support for `previous_response_id` to the text generation because it was very simple to add and could certainly be useful for some folks.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -4,48 +4,410 @@
...
+ *     url?: string,
+ *     revised_prompt?: string
+ * }
+ * @phpstan-type ResponseData array{
+ *     created?: int,
+ *     data?: list<ImageData>
+ * }
  */
-class OpenAiImageGenerationModel extends AbstractOpenAiCompatibleImageGenerationModel
+class OpenAiImageGenerationModel extends AbstractApiBasedModel implements ImageGenerationModelInterface
```
felixarntz: I don't understand why you're not using `AbstractOpenAiCompatibleImageGenerationModel`. It relies on the same `images/generations` endpoint your custom implementation does now, so why reinvent the wheel?

If there are some enhancements to make for the OpenAI implementation, either it should be made to the abstract base class, or it can be made here but still on top of the same base class. Unless there is a strong reason _not_ to use that?

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,693 @@
...
+         *
+         * This allows chaining responses together for multi-turn conversations without
+         * resending the entire conversation history. Pass the response ID from a previous
+         * call to continue the conversation.
+         *
+         * @see https://platform.openai.com/docs/guides/conversation-state
+         */
+        if (!empty($customOptions['previous_response_id'])) {
+            $params['previous_response_id'] = $customOptions['previous_response_id'];
+        }
```
felixarntz: Why is this needed? To be clear: My question is not whether or not to support `previous_response_id`, but about why not simply rely on the regular custom options handling? That would have the same outcome as far as I can tell.

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,693 @@
...
+                // This should be impossible due to class internals, but still needs to be checked.
+                throw new RuntimeException(
+                    'The inline file must contain base64 data.'
+                );
+            }
+            $mimeType = $file->getMimeType();
+            if ($file->isImage()) {
+                return [
+                    'type' => 'input_image',
+                    'image_url' => "data:{$mimeType};base64,{$fileBase64Data}",
```
felixarntz: This should use `File::getDataUri`.

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,693 @@
...
+                ];
+            }
+            // For other file types (like PDF), use input_file.
+            return [
+                'type' => 'input_file',
+                'filename' => 'file',
+                'file_data' => "data:{$mimeType};base64,{$fileBase64Data}",
```
felixarntz: Same here.

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,693 @@
...
+            // Function calls in input are typically from assistant messages in conversation history.
+            // The Responses API handles this differently - we include them as part of the message.
+            $functionCall = $part->getFunctionCall();
+            if (!$functionCall) {
+                throw new RuntimeException(
+                    'The function_call typed message part must contain a function call.'
+                );
+            }
+            // Skip function calls in input - they're part of the conversation flow.
+            return null;
```
felixarntz: Are you certain this is how it works? Not that I know any better, but if they're skipped when sending the full input, that seems like it could lead to problems? Wouldn't it mean missing context when conversation history includes a function call output, but not the function call it is in response to?

Is there any documentation that says this is how it should be done?

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,693 @@
...
+            // For now, we use the basic form.
+            $tools[] = $webSearchTool;
+        }
+
+        if ($codeInterpreter) {
+            $tools[] = [
+                'type' => 'code_interpreter',
+                'container' => ['type' => 'auto'],
+            ];
+        }
```
felixarntz: Fine for now, but ideally this should become an actual tool class in our foundation. Other providers also support some kind of code execution, and even if that wasn't the case, it could be a tool class that for now is only supported by one provider.

felixarntz (changes requested):
@JasonTheAdams The Responses API implementation looks solid, a few small points there.

For image generation, now I'm confused why you're not reusing the existing abstract OpenAI compatible implementation, which already relies on that same endpoint.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
JasonTheAdams: Good catch! I just didn't realize it used that endpoint. Resolved in 32291e8

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: Good catch! Resolve in 932f26d

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: Resolved in 932f26d

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: Resolved in 932f26d

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: Oof, I'm embarrassed on this one. I was so focused on the Responses and Image API nuances that this completely slipped by me. No, this is completely wrong. I updated this to work properly, also enforcing that a `Message` can only contain a single part if it's a function call or response.

Resolved in bf04bc6

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: I really like this so I went ahead and added it in 1f41f2b

As a note, there's not much commonality between how this works in OpenAI, Anthropic, and Google. OpenAI has the most configuration options, Anthropic supports passing a container ID (which OpenAI supports), and Google has no parameters right now. So I took the container id as a common denominator and added support for open-ended custom options. Curious what you think of this!

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -30,22 +49,199 @@ protected function createRequest(HttpMethodEnum $method, string $path, array $he
...
-         * Conversely, they do not support 'response_format', but always return a base64 encoded image.
-         */
-        if (isset($params['model']) && is_string($params['model']) && str_starts_with($params['model'], 'gpt-image-')) {
+        // GPT image models use output_format, DALL-E uses response_format.
+        if ($this->isGptImageModel($modelId)) {
+            // For GPT image models, convert response_format to the appropriate format.
+            // The parent sets response_format, but GPT models don't need it.
             unset($params['response_format']);
-        } else {
-            unset($params['output_format']);
```
felixarntz, line 46: Why is this removed? Was there something wrong with it?

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -4,22 +4,41 @@
...
+ *     data?: list<array{url?: string, b64_json?: string}>
+ * }
  */
 class OpenAiImageGenerationModel extends AbstractOpenAiCompatibleImageGenerationModel
 {
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
```
felixarntz: I don't think this method is new? So it should use `@since 1.0.0`.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -30,22 +49,199 @@ protected function createRequest(HttpMethodEnum $method, string $path, array $he
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     *
+     * @since n.e.x.t
```
felixarntz: See above.

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -30,22 +49,199 @@ protected function createRequest(HttpMethodEnum $method, string $path, array $he
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Overrides the parent to handle OpenAI's `created` timestamp instead of `id`.
+     *
+     * @since n.e.x.t
+     */
+    protected function parseResponseToGenerativeAiResult(
```
felixarntz: Why are you overriding the parent method here? Shouldn't this simply use the parent method as is?

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
```diff
@@ -4,30 +4,697 @@
...
+                    throw new RuntimeException(
+                        'The function_response typed message part must contain a function response.'
+                    );
+                }
+                return [
+                    'type' => 'function_call_output',
+                    'call_id' => $functionResponse->getId(),
+                    'output' => json_encode($functionResponse->getResponse()),
+                ];
+            }
```
felixarntz: Why not have this code be part of `getMessagePartData()`? It seems like logically the right place, so why implement it outside of that method? Function call and response are also message parts.

src/Tools/DTO/CodeExecution.php
```diff
@@ -0,0 +1,133 @@
...
+ * This DTO defines configuration for code execution/interpreter tools
+ * that AI models can use to run code.
+ *
+ * @since n.e.x.t
+ *
+ * @phpstan-type CodeExecutionArrayShape array{containerId?: string|null, customOptions?: array<string, mixed>}
+ *
+ * @extends AbstractDataTransferObject<CodeExecutionArrayShape>
+ */
+class CodeExecution extends AbstractDataTransferObject
```
felixarntz: Can we introduce this in a separate PR? I think it requires a bit more thought, including exploration of how other providers handle code execution, to find the right abstraction.

This is an entirely new feature so IMO it shouldn't need to be a blocker to this PR.

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
felixarntz: I'm generally on board with adding it, but I think doing that in this PR is preliminary. I would prefer that we don't include code execution here because it's something completely new and requires a bit more exploration.

src/Tools/DTO/CodeExecution.php
JasonTheAdams: Sure! I made a new `code-execution-dto` branch with this DTO change, and then moved this branch back.

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: I see what you mean. There are really two things happening here:
* Validating that a message only has a single part if that part is a function call or response
* Converting it into the needed part

For now, I'll separate out the conversion and keep the validation here. Open to suggestions. We could even validate inside of `Message` itself to make sure of this.

I took a stab in 037e873. I'm curious what you think!

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
JasonTheAdams: Resolved in 97a5d6a

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
JasonTheAdams: Resolved in dfb63a7

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
JasonTheAdams: Resolved in dfb63a7

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
JasonTheAdams: Good catch. Resolved in da89e4d

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: I'm fine with that. I made another branch we can continue from. Removed code interpretation in 603b98b

src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
JasonTheAdams: Per our discussion I moved the validation back to be OpenAI specific: 8a563be

src/Tools/DTO/CodeExecution.php
JasonTheAdams: I made a draft PR for this: https://github.com/WordPress/php-ai-client/pull/167

src/ProviderImplementations/OpenAi/OpenAiImageGenerationModel.php
```diff
@@ -40,12 +52,144 @@ protected function prepareGenerateImageParams(array $prompt): array
          * Only the newer 'gpt-image-' models support passing a MIME type ('output_format').
          * Conversely, they do not support 'response_format', but always return a base64 encoded image.
          */
-        if (isset($params['model']) && is_string($params['model']) && str_starts_with($params['model'], 'gpt-image-')) {
+        if ($this->isGptImageModel($this->metadata()->getId())) {
```
felixarntz: Similar to some other comment I left a while ago, it is safer to use the actual data it directly depends on, not the source of _that_ data:
```suggestion
        if ($this->isGptImageModel($params['model'])) {
```

felixarntz (approved):
@JasonTheAdams LGTM - awesome work! One small last nit-pick, but good to go.

PR #163, opened 2026-01-05
Add anthropic-beta header for structured outputs.

### LINKED ISSUE #151: Anthropic Structured Output Issue
## Problem

When using `->asJsonResponse($schema)` with the Anthropic provider, the API returns errors:

```
Bad Request (400) - response_format.json_schema.name: Field required
# Added `name` field to schema wrapper, then got:

Bad Request (400) - response_format.json_schema.strict: Field required
# Added `strict: true` to schema wrapper, then got:

Bad Request (400) - output_config: Extra inputs are not permitted
# Anthropic doesn't recognize `response_format` at all—it's an OpenAI parameter
```

The library sends OpenAI's `response_format` parameter to Anthropic's API, which now rejects it. This appears to have started around December 4, 2025 when Anthropic launched their structured outputs feature—prior to this, the parameter was likely ignored.

## Proposed Solution

Anthropic's structured outputs use a different format than OpenAI:

| | OpenAI | Anthropic |
|---|--------|-----------|
| Parameter | `response_format` | `output_format` |
| Beta header | None | `anthropic-beta: structured-outputs-2025-11-13` |

The Anthropic provider should override `prepareGenerateTextParams()` to:
1. Use `output_format` instead of `response_format`
2. Add the required beta header

## References

- [Anthropic Structured Outputs Documentation](https://docs.anthropic.com/en/docs/build-with-claude/structured-outputs)
- [Anthropic API Release Notes](https://docs.anthropic.com/en/release-notes/api)

johnhooks:
When requesting structured JSON output from the Anthropic API using the `output_format` parameter with a JSON schema, the API requires the `anthropic-beta: structured-outputs-2025-11-13` header to be present in the request. Without this header, Anthropic rejects the request because structured outputs is currently a beta feature that must be explicitly opted into. This fix detects when JSON schema output is configured via `getOutputMimeType()` and `getOutputSchema()` and automatically adds the required beta header, allowing structured output requests to succeed.

See [Claude Docs: Structured outputs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)

Fixes #151.

### DIFF
--- a/src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
+++ b/src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php
@@ -79,10 +79,18 @@ class AnthropicTextGenerationModel extends AbstractApiBasedModel implements Text
 
         $params = $this->prepareGenerateTextParams($prompt);
 
+        $headers = ['Content-Type' => 'application/json'];
+
+        // Add beta header for structured outputs if JSON schema output is requested.
+        $config = $this->getConfig();
+        if ('application/json' === $config->getOutputMimeType() && $config->getOutputSchema()) {
+            $headers['anthropic-beta'] = 'structured-outputs-2025-11-13';
+        }
+
         $request = new Request(
             HttpMethodEnum::POST(),
             AnthropicProvider::url('messages'),
-            ['Content-Type' => 'application/json'],
+            $headers,
             $params,
             $this->getRequestOptions()
         );


johnhooks:
@JasonTheAdams I was going to add a test or two, but at the moment there aren't any tests for `AnthropicTextGenerationModel`, or am I missing something?

JasonTheAdams:
Thanks, @johnhooks! You're right. There aren't currently any tests. I'll let @felixarntz decide if he'd like to add tests in this PR since he built out #156. I'm also adding him as a reviewer.

johnhooks:
I had tests in my implementation, before #156 was created [AnthropicTextGenerationModelTest.php](https://github.com/johnhooks/php-ai-client/blob/fix/issue-151/tests/unit/ProviderImplementations/Anthropic/AnthropicTextGenerationModelTest.php). I could attempt to port it over and add tests for all of `AnthropicTextGenerationModel`.

UPDATE: I updated the test for the current implementation, and intended to stack the PRs, but I don't think I can since the incoming branch is from my fork. But I can either do separate PR, or merge into this one. https://github.com/johnhooks/php-ai-client/blob/cbaf7dc4f634bc7e6bd8676a393d600eb05034bc/tests/unit/ProviderImplementations/Anthropic/AnthropicTextGenerationModelTest.php

felixarntz (approved):
@johnhooks I think this is good to go. While we should probably have tests for these, I think this is its own task to do, and this change is simple enough to not require being blocked to add tests.

Great catch, thank you for fixing this!

PR #164, opened 2026-01-08
Make HTTP transporter setup in registry more flexible to not throw discovery exception when no HTTP request is even being attempted

### LINKED ISSUE #41: Registering custom providers is not possible due to uninitialized HTTP client discovery
## Context

We are trying to build a WordPress plugin that registers a custom AI provider (see https://github.com/mittwald/wordpress-ai-provider for context).

Currently, registering a custom AI provider is not possible without adding an additional action hook to the `WordPress\AI_Client\AI_Client::init()` function (or other, equally dirty, workarounds). This is because the current initialization sequence required the HTTP client discovery to be set up before provider registrations, but providers must be registered before the API Credentials Manager is initialized.

_Note: This issue is based on a Slack thread in `#core-ai` (https://wordpress.slack.com/archives/C08TJ8BPULS/p1765290849882899)._

## Problem description

1. Registering a provider using the `init` hook _before_ the `WordPress\AI_Client\AI_Client::init` invocation fails with an `Http\Discovery\Exception\DiscoveryFailedException`, because the HTTP client discovery has not been initialized yet. 

    ```
    Fatal error: Uncaught Http\Discovery\Exception\DiscoveryFailedException: Could not find resource using any discovery strategy. Find more information at http://docs.php-http.org/en/latest/discovery.html#common-errors - No valid candidate found using strategy "Http\Discovery\Strategy\CommonClassesStrategy". We tested the following candidates: Http\Discovery\Strategy\CommonClassesStrategy::symfonyPsr18Instantiate, GuzzleHttp\Client, Http\Discovery\Strategy\CommonClassesStrategy::buzzInstantiate. - No valid candidate found using strategy "Http\Discovery\Strategy\CommonPsr17ClassesStrategy". We tested the following candidates: . in /var/www/html/wp-content/plugins/ai/vendor/php-http/discovery/src/Exception/DiscoveryFailedException.php:41 

    Stack trace: 
    #0 /var/www/html/wp-content/plugins/ai/vendor/php-http/discovery/src/ClassDiscovery.php(94): Http\Discovery\Exception\DiscoveryFailedException::create() 
    #1 /var/www/html/wp-content/plugins/ai/vendor/php-http/discovery/src/Psr18ClientDiscovery.php(26): Http\Discovery\ClassDiscovery::findOneByType() 
    #2 /var/www/html/wp-content/plugins/ai/vendor/wordpress/php-ai-client/src/Providers/Http/HttpTransporterFactory.php(34): Http\Discovery\Psr18ClientDiscovery::find() 
    #3 /var/www/html/wp-content/plugins/ai/vendor/wordpress/php-ai-client/src/AiClient.php(106): WordPress\AiClient\Providers\Http\HttpTransporterFactory::createTransporter() 
    #4 /var/www/html/wp-content/plugins/wp-ai-provider/wp-ai-provider.php(77): WordPress\AiClient\AiClient::defaultRegistry()
    ...
    ```

    This initialization happens in the `AI_Client::init()` function in line 56:

    https://github.com/WordPress/wp-ai-client/blob/dacb6e8b73ff3097c41a256dfe26fb3d0cea9f48/includes/AI_Client.php#L56

2. Registering a provider using the `init` or `wp_loaded` hook _after_ the `WordPress\AI_Client\AI_Client::init` invocation fails because the API Credentials Manager has already been initialized, and providers must be registered before that happens. Otherwise, the provider will not appear in the settings page and will not be recognized for the experiments.

    This initialization happens in the `AI_Client::init()` function in line 66:

    https://github.com/WordPress/wp-ai-client/blob/dacb6e8b73ff3097c41a256dfe26fb3d0cea9f48/includes/AI_Client.php#L66

## Possible workarounds/solutions

1. Manually initialize the HTTP client discovery before registering the provider. This can be done by calling the HTTP client discovery initialization code directly in the plugin before the `init` hook is fired. However, this is not ideal as it requires knowledge of the internal workings of the AI Client library.

    Also, the `Psr18ClientDiscovery::prependStrategy` call that is run during client discovery initialization is not idempotent, and may cause unintended side effects on repeated (identical) invocations.
   
2. Injecting directly into the `$wp_ai_client_providers_meta` global, based on a suggestion by @justlevine, although he himself stated that this might only be a fragile solution.[^injecting] In any case, injecting into global variables does not feel like the intended way to register custom providers.

3. Add a new action hook to the `WordPress\AI_Client\AI_Client::init` function at some point between lines 56 and 66.
   
4. According to a Slack comment by @felixarntz,[^http-fix] provider registration is intended to work before the `WP_AI_Client_Discovery_Strategy::init` call... which it doesn't, in this case. If this is intended to work, it would make sense to fix the initialization order, so that providers can regularly be registered by using the `init` hook _before_ the AI client initialization.

I would gravitate towards either solution 3 or 4. We would also be willing to open a respective PR as soon as there is a consensus on a solution.

[^injecting]: https://wordpress.slack.com/archives/C08TJ8BPULS/p1765366811625229?thread_ts=1765290849.882899&cid=C08TJ8BPULS
[^http-fix]:  https://wordpress.slack.com/archives/C08TJ8BPULS/p1765415337195029?thread_ts=1765290849.882899&cid=C08TJ8BPULS

felixarntz:
This partially will fix https://github.com/WordPress/wp-ai-client/issues/41 - specifically, when a provider is registered before an HTTP client has been configured in the PSR-17 discovery mechanism.

This is probably not sufficient by itself as a proper fix for that issue, but it's already a partial fix, which means a workaround will no longer be needed. And since this library is not meant to be _only_ used with WordPress, it makes sense to optimize for general flexibility:

* The problem is that currently, whenever someone tries to register a provider, the default registry is initialized.
* And since the HTTP transporter is set as part of that, we run into an exception if no 3P code has provided an HTTP client yet. This is bad DX because there's no point in throwing an exception at this point, given that no HTTP request is even being attempted.

This PR fixes that, by dynamically attempting to set the default HTTP transporter when no (other) HTTP transporter has already been set, and doing that in a way that is forgiving in terms of the discovery exception.

cc @martin-helmich @LukasFritzeDev

### DIFF
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -13,7 +13,6 @@ use WordPress\AiClient\ProviderImplementations\Google\GoogleProvider;
 use WordPress\AiClient\ProviderImplementations\OpenAi\OpenAiProvider;
 use WordPress\AiClient\Providers\Contracts\ProviderAvailabilityInterface;
 use WordPress\AiClient\Providers\Contracts\ProviderInterface;
-use WordPress\AiClient\Providers\Http\HttpTransporterFactory;
 use WordPress\AiClient\Providers\Models\Contracts\ModelInterface;
 use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
 use WordPress\AiClient\Providers\ProviderRegistry;
@@ -109,7 +108,6 @@ class AiClient
             $registry = new ProviderRegistry();
 
             // Set up default HTTP transporter and register built-in providers.
-            $registry->setHttpTransporter(HttpTransporterFactory::createTransporter());
             $registry->registerProvider(AnthropicProvider::class);
             $registry->registerProvider(GoogleProvider::class);
             $registry->registerProvider(OpenAiProvider::class);
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -107,7 +107,7 @@ class AiClient
         if (self::$defaultRegistry === null) {
             $registry = new ProviderRegistry();
 
-            // Set up default HTTP transporter and register built-in providers.
+            // Register built-in providers.
             $registry->registerProvider(AnthropicProvider::class);
             $registry->registerProvider(GoogleProvider::class);
             $registry->registerProvider(OpenAiProvider::class);
--- a/src/Providers/ProviderRegistry.php
+++ b/src/Providers/ProviderRegistry.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\Providers;
 
+use Http\Discovery\Exception\NotFoundException as DiscoveryNotFoundException;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
 use WordPress\AiClient\Providers\Contracts\ProviderInterface;
@@ -14,6 +15,7 @@ use WordPress\AiClient\Providers\Http\Contracts\HttpTransporterInterface;
 use WordPress\AiClient\Providers\Http\Contracts\RequestAuthenticationInterface;
 use WordPress\AiClient\Providers\Http\Contracts\WithHttpTransporterInterface;
 use WordPress\AiClient\Providers\Http\Contracts\WithRequestAuthenticationInterface;
+use WordPress\AiClient\Providers\Http\HttpTransporterFactory;
 use WordPress\AiClient\Providers\Http\Traits\WithHttpTransporterTrait;
 use WordPress\AiClient\Providers\Models\Contracts\ModelInterface;
 use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
@@ -85,14 +87,26 @@ class ProviderRegistry implements WithHttpTransporterInterface
         // If there is already a HTTP transporter instance set, hook it up to the provider as needed.
         try {
             $httpTransporter = $this->getHttpTransporter();
-            $this->setHttpTransporterForProvider($className, $httpTransporter);
         } catch (RuntimeException $e) {
             /*
              * If this fails, it's okay. There is no defined sequence between setting the HTTP transporter in the
              * registry and registering providers in it, so it might be that the transporter is set later. It will be
              * hooked up then.
-             * Therefore we can simply ignore this exception.
+             * But for now we can ignore this exception and attempt to set the default HTTP transporter, if possible.
              */
+            try {
+                $this->setHttpTransporter(HttpTransporterFactory::createTransporter());
+            } catch (DiscoveryNotFoundException $e) {
+                /*
+                 * If no HTTP client implementation can be discovered yet, we can ignore this for now.
+                 * It might be set later, so it's not a hard error at this point.
+                 * We'll try again the next time a provider is registered, or maybe by that time an explicit
+                 * HTTP transporter will have been set.
+                 */
+            }
+        }
+        if (isset($httpTransporter)) {
+            $this->setHttpTransporterForProvider($className, $httpTransporter);
         }
 
         // Hook up the request authentication instance, using a default if not set.
--- a/tests/unit/Providers/ProviderRegistryTest.php
+++ b/tests/unit/Providers/ProviderRegistryTest.php
@@ -571,29 +571,6 @@ class ProviderRegistryTest extends TestCase
     }
 
     /**
-     * Tests bindModelDependencies without HTTP transporter when model needs it.
-     *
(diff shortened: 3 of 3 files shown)


JasonTheAdams:
Really interesting, @felixarntz! If I'm following, this comes down to the fact that the WP AI Client doesn't connect the `WP_AI_Client_Discovery_Strategy` until `init` fires? So if someone were to register a provider prior to this point and calls `AIClient::defaultRegistry()`, it will throw a `RealNotFoundException` inside `HttpTransportFactory::createTransport()`? Gotta love a good race condition. 😄 

I mean, really, the discovery system can be hooked up as soon as possible. It doesn't require that anything in WordPress itself is loaded at that time. I could see providers wanting to be registered (and even usable) prior to `init`, so that may be worth considering. I saw your comment in the other Issue that you think there should be a distinct hook for registering providers, which I think I agree with.


As a note, while the WP AI Client does load in `init`, if used as a plugin, it can be used as a package, as in AI Experiments. That plugin, then, is the one deciding when to call `AI_Client::init()`. So if we move things to load earlier, we'll need to do it in both places.

If I'm tracking with all that, the goal of this PR is to at least make it so that when a provider is registered prior to this point it still works. This done is by suppressing the `NotFoundException`, which is necessary because we really don't have any other way of knowing if an HTTP system ought to have been loaded yet. Things would work later because whenever `ProviderRegistery::setHttpTransporter()` is finally called, it sets it for all providers already set.

My only concern with this is the remaining race condition of someone trying to use `AiClient::prompt()` when a provider has been registered by the transporter has not been successfully set. We're sort of shuffling things around. That may be fair, but I wanted to at least raise the point.

Am I tracking with all this?

felixarntz:
@JasonTheAdams I think you're right on all accounts. But I'd like us to look at this from a less WordPress-centric angle too. While we can shuffle things around in the WP AI Client, and maybe set up the HTTP client very early in the WordPress bootstrap cycle, I think there's still no point in throwing an HTTP discovery exception when a provider is registered. So this PR is mostly about being more "forgiving", or less aggressive.

An HTTP client and HTTP transporter only need to be set up once any AI HTTP request is made (e.g. via `AiClient::prompt()`). At that point, it _should_ throw the exception. But before any of that happens, it really doesn't matter, so throwing an exception can unnecessarily throw a wrench in things.

So to your last point, I'm not sure I follow what your concern is with `AiClient::prompt()` - given at that point the lack of an HTTP transporter or client should indeed throw an exception. And I don't think this PR does anything that would jeopardize that?

JasonTheAdams:
Got it. I'm tracking with what you mean. The think the nested `try...catch` is unnecessary, then. The `RuntimeException` catch only seems necessary because we're using the `WithHttpTransporterTrait`, but I don't think that trait really fits here. We're purposely considering the transporter as something nullable in this class, trusting it will eventually come.

I think if we drop the use of that trait and make it a nullable property, then we can clean that up. The only other place the getter is used is in `bindModelDependencies`, and if we consider that a fair point of failure, then we can throw an exception there with a more meaningful exception/description.

felixarntz:
Hmm, but why not use the trait? Can you clarify what you think is the problem with it?

JasonTheAdams:
> Hmm, but why not use the trait? Can you clarify what you think is the problem with it?

The problem is that the trait assumes the transporter is non-nullable, which is why we're having to catch the exception for the getter. In the case of the `ProviderRegistry` we're saying that's not the case — null is completely fine until a certain point. The trait is doing very little for us, by design doesn't introduce a contract, and is making the code less clean, not more. By removing it we can drop the first level `try...catch` and just check if the property is null or not. We can do the same later in `bindModelDependencies` and throw a more meaningful exception if it is null.

Make sense?

felixarntz:
@JasonTheAdams 
> The trait is doing very little for us, by design doesn't introduce a contract, and is making the code less clean, not more.

I'm not sure I agree. Why is catching an exception a bad thing to do here? The trait still takes away responsibility of having the property, setter, and getter. It's all super basic functionality, but I'd argue the trait simply reduces the amount of code we have to (re-)write.

JasonTheAdams (approved):
@felixarntz I think we just disagree, then. Hahah! But it's not a hill I'd die on.

Personally, I think the trait is not matching the actual intentions of the class — namely, that a nullable property is valid. I don't even think we need the getter, either, because it's only used internally. So I think we're applying something to reduce code that isn't actually a proper fit, it just feels close enough. The exception catching becomes a code smell, not that catching exceptions is bad in any way.

I'm approving because I've made my point and am fine either way. 😄

PR #170, opened 2026-01-16
Remove unimplemented streaming methods since there is no point in having them for now

### LINKED ISSUE #166: Removing streaming related scaffolding for now since it doesn't do anything
We don't support streaming yet. So as long as that's the case, there's no point of having interface methods that always throw an exception.

This also means we can later think about a cleaner interface structure once we _actually_ implement streaming. For example, it might be a better solution to have a standalone interface for a model that can stream text - not every provider necessarily supports both regular text generation _and_ text streaming.

felixarntz:
Fixes #166


JasonTheAdams (approved):
Get em out of here! 😁

PR #175, opened 2026-01-20
Fix: Corrects handling of empty function call args

JasonTheAdams:
This attempts to address the bug found in this trac ticket: https://core.trac.wordpress.org/ticket/64483

Basically, when models send a function call they provide the arguments for the function every time, even when there are no arguments (e.g. `get_current_time()`). It's passed as an empty JSON object: `args: {}`. This is converted into an array, because empty lists and associative arrays can't be differentiated in PHP. As a result, an empty array is passed as the arguments of a new `FunctionCall`. This is wrong. An empty JSON object means "there are no arguments" which, in PHP, would be `null`.

This attempts to correct this in two places:
1. By normalizing an empty array in `FunctionCall` to `null`. This is because this is likely a common scenario Providers will run into, and passing an empty array is otherwise not meaningful.
2. When including the `FunctionCall` as part of a message history, it handles converting `null` into an empty object correctly, as the model would expect.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,8 +12,10 @@ The following commands are available for development and can be run using Compos
 *   `composer phpcs`: Runs PHP_CodeSniffer to check for coding standards violations.
 *   `composer phpcbf`: Automatically fixes coding standards violations that can be fixed automatically.
 *   `composer phpstan`: Runs PHPStan for static analysis.
-*   `composer test`: Runs the PHPUnit test suite (this is an alias for `composer phpunit`).
-*   `composer phpunit`: Runs the PHPUnit test suite.
+*   `composer test`: Runs the unit test suite (this is an alias for `composer test:unit`).
+*   `composer test:unit`: Runs the unit test suite.
+*   `composer test:integration`: Runs the integration test suite (requires API keys).
+*   `composer phpunit`: Runs all tests (unit and integration).
 
 ## Coding Standards & Compatibility Constraints
 
@@ -42,7 +44,17 @@ The project uses custom exceptions for error handling. When appropriate, throw a
 
 ### Testing
 
-The project uses PHPUnit for unit testing. Test files are located in the `tests/` directory and mirror the structure of the `src/` directory. The test suite is executed by running `composer phpunit`. The `tests/mocks` directory contains mock implementations for testing purposes, and `tests/traits` contains reusable testing traits. All new code requires corresponding unit tests.
+The project uses PHPUnit for testing. Tests are organized into two suites:
+
+*   **Unit tests** (`tests/unit/`): Fast, isolated tests that mirror the structure of the `src/` directory. Run with `composer test:unit`.
+*   **Integration tests** (`tests/integration/`): Tests that make real API calls to AI providers (OpenAI, Google, Anthropic). Run with `composer test:integration`. These require API keys configured in a `.env` file.
+
+Supporting test infrastructure:
+
+*   `tests/mocks/`: Mock implementations for testing purposes.
+*   `tests/traits/`: Reusable testing traits.
+
+All new code requires corresponding unit tests.
 
 ## Project Architecture Overview
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ The project uses custom exceptions for error handling. When appropriate, throw a
 The project uses PHPUnit for testing. Tests are organized into two suites:
 
 *   **Unit tests** (`tests/unit/`): Fast, isolated tests that mirror the structure of the `src/` directory. Run with `composer test:unit`.
-*   **Integration tests** (`tests/integration/`): Tests that make real API calls to AI providers (OpenAI, Google, Anthropic). Run with `composer test:integration`. These require API keys configured in a `.env` file.
+*   **Integration tests** (`tests/integration/`): Tests that make real API calls to AI providers (OpenAI, Google, Anthropic). Run with `composer test:integration`. These require API keys configured in a `.env` file, and running them will incur cost with these providers.
 
 Supporting test infrastructure:
 
--- a/src/ProviderImplementations/Google/GoogleTextGenerationModel.php
+++ b/src/ProviderImplementations/Google/GoogleTextGenerationModel.php
@@ -120,7 +120,7 @@ class GoogleTextGenerationModel extends AbstractApiBasedModel implements TextGen
 
         $systemInstruction = $config->getSystemInstruction();
         if ($systemInstruction) {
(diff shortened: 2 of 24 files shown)


JasonTheAdams:
Having spoken with @felixarntz, I don't think this will work as it is. We can't generically solve this with normalization because we can't make the assumption that every function declaration's arguments is an object — it may be an array, in which case an empty array is valid. Alas, we'll need to carefully solve this on a per Provider basis. We'll iterate this PR to get there.

justlevine:
> We can't generically solve this with normalization because we can't make the assumption that every function declaration's arguments is an object — it may be an array, in which case an empty array is valid. Alas, we'll need to carefully solve this on a per Provider basis.

@JasonTheAdams @felixarntz since this sounds like a lot of work, I just want to double-check: is the need to coerce `{}` to `null` an issue for _all_ FunctionCalls / declarations over PHP, or something that should really be handled on the _consumer_ level if they need to differentiate?

If it's the latter (and therefore not a php client issue but a WP client issue ) we could probably handle the normalization in https://github.com/WordPress/wp-ai-client/blob/dacb6e8b73ff3097c41a256dfe26fb3d0cea9f48/includes/Builders/Helpers/Ability_Function_Resolver.php#L91 and rely on the fact that Abilities API has verbose schema checking. E.g (psuedocode from my phone):

```php
# wp-ai-client/includes/Builders/Helpers/Ability_Function_Resolver.php

// Execute the ability.
$args   = $call->getArgs();

if ( empty( $args ) ) {
  // Coerce the falsey shape if it's not supported.
  $args = true === rest_validate_value_from_schema( $args, $ability->get_input_schema() ) ? $args : null;
}

$result = $ability->execute( $args );
```

It's an empty var so the double validation is negligible.

felixarntz:
@justlevine I think there are multiple problems. Some of it may be related to logic from WP AI Client or Abilities API, but after looking at just the PHP AI Client SDK yesterday in relation to this, I noticed at least a few problems that come from this repo alone.

In a nutshell: The way providers handle "no args" differs, and we don't handle this scenario properly for some of them. And those differences cover both the `FunctionDeclaration` (how the tool shape itself is passed) and the `FunctionCall` (how the model returns a tool call).

* Some providers require the relevant parameters/input field to be omitted.
* Other providers require the relevant parameters/input field to be an empty object.

#176 was started as a result of this research. Unfortunately, the providers' documentation is not great on details like this in most cases. For all 3 providers, after careful review of the relevant API reference bits, we still have questions. So the only way to know for sure is to try different variants and see what the respective provider does. Obviously #176 will have far more benefits than just for this concrete problem, but it showed that only by testing the actual provider connection and behavior we'll have proper coverage - and we can even use it to learn about subtle nuances.

JasonTheAdams:
Ok, @felixarntz! I've switched this around! Here's what I've done:

1. I adjusted it so 0-argument function calling is handled on a per-provider basis. It's nuanced, but it's really not that bad.
2. I added a bunch more integration tests that really gets into function calling, touching on `FunctionDeclaration`, `FunctionCall`, and `FunctionResponse`.

All the tests are passing for all the providers and I'm feeling pretty darn good about these new integration tests. I just wish we'd added them months ago. 😎

AGENTS.md
```diff
@@ -42,7 +44,17 @@ The project uses custom exceptions for error handling. When appropriate, throw a
 
 ### Testing
 
-The project uses PHPUnit for unit testing. Test files are located in the `tests/` directory and mirror the structure of the `src/` directory. The test suite is executed by running `composer phpunit`. The `tests/mocks` directory contains mock implementations for testing purposes, and `tests/traits` contains reusable testing traits. All new code requires corresponding unit tests.
+The project uses PHPUnit for testing. Tests are organized into two suites:
+
+*   **Unit tests** (`tests/unit/`): Fast, isolated tests that mirror the structure of the `src/` directory. Run with `composer test:unit`.
+*   **Integration tests** (`tests/integration/`): Tests that make real API calls to AI providers (OpenAI, Google, Anthropic). Run with `composer test:integration`. These require API keys configured in a `.env` file.
```
felixarntz:
```suggestion
*   **Integration tests** (`tests/integration/`): Tests that make real API calls to AI providers (OpenAI, Google, Anthropic). Run with `composer test:integration`. These require API keys configured in a `.env` file, and running them will incur cost with these providers.
```

src/ProviderImplementations/Google/GoogleTextGenerationModel.php
```diff
@@ -370,10 +370,15 @@ protected function getMessagePartData(MessagePart $part): ?array
                     'The function_call typed message part must contain a function call.'
                 );
             }
+            // Ensure null becomes empty object for Google's API which expects an object.
+            $args = $functionCall->getArgs();
+            if ($args === null) {
+                $args = new \stdClass();
+            }
```
felixarntz: Are you sure? According to https://ai.google.dev/api/caching#FunctionCall, `args` is optional. So better to omit the entire field in the payload in that case.

src/ProviderImplementations/Google/GoogleTextGenerationModel.php
```diff
@@ -742,11 +747,21 @@ protected function parseResponseCandidateMessagePart(array $partData): MessagePa
...
+            /*
+             * Normalize empty object/array to null.
+             * Google returns `args: {}` for functions with no arguments,
+             * which becomes an empty array after json_decode. Semantically,
+             * an empty object means "no arguments".
+             */
+            $args = $partData['functionCall']['args'];
+            if (is_array($args) && count($args) === 0) {
+                $args = null;
+            }
```
felixarntz, line 756: Per my comment above, I'm a bit baffled by this. Have you tested this? The `parameters` field for the function declaration is not present based on the current logic when we pass function declarations to the Google API, so based on that and the other documentation referenced above, I would expect the `args` field to not be present at all.

felixarntz (commented):
@JasonTheAdams Looks great! Only one last question.

src/ProviderImplementations/Google/GoogleTextGenerationModel.php
JasonTheAdams: Good catch! Resolved in 35ed290

src/ProviderImplementations/Google/GoogleTextGenerationModel.php
JasonTheAdams: I adjusted this to handle either an empty object or omission. I feel like this is the safest way to approach this since either is completely fair.

Resolved in 712a46a

felixarntz (approved):

PR #176, opened 2026-01-21
Adds integration test scaffolding

JasonTheAdams:
This introduces integration tests for running actual prompts against the models. As this has grown it's become more obvious that the only way to really catch Provider issues is by running prompts. While we would do some basic testing locally, having a proper suite of tests is really useful for catching things and avoiding regressions.

These tests can't be run automatically as they require credentials. As such, we'll want to run this manually periodically when developing, and every time before publishing a new release.

API keys can be loaded as part of the environment or put in a `.env` file.

NOTE:
There is a purposely failing test for OpenAI function calling. This will be resolved in #175 as a broader effort to correct issues with function calling across the provider APIs.

### DIFF
--- a/tests/integration/Anthropic/FunctionCallingIntegrationTest.php
+++ b/tests/integration/Anthropic/FunctionCallingIntegrationTest.php
@@ -1,0 +1,172 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Tests\integration\Anthropic;
+
+use PHPUnit\Framework\TestCase;
+use WordPress\AiClient\AiClient;
+use WordPress\AiClient\Results\DTO\GenerativeAiResult;
+use WordPress\AiClient\Tests\integration\traits\IntegrationTestTrait;
+use WordPress\AiClient\Tools\DTO\FunctionCall;
+use WordPress\AiClient\Tools\DTO\FunctionDeclaration;
+
+/**
+ * Integration tests for Anthropic function calling.
+ *
+ * These tests make real API calls to Anthropic and require the ANTHROPIC_API_KEY
+ * environment variable to be set.
+ *
+ * @group integration
+ * @group anthropic
+ * @group function-calling
+ *
+ * @coversNothing
+ */
+class FunctionCallingIntegrationTest extends TestCase
+{
+    use IntegrationTestTrait;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->requireApiKey('ANTHROPIC_API_KEY');
+    }
+
+    /**
+     * Tests function calling with multiple arguments.
+     */
+    public function testFunctionCallingWithMultipleArguments(): void
+    {
+        $getWeather = new FunctionDeclaration(
+            'get_weather',
+            'Get the current weather for a location',
+            [
+                'type' => 'object',
+                'properties' => [
+                    'location' => [
+                        'type' => 'string',
+                        'description' => 'The city and state, e.g. San Francisco, CA',
+                    ],
+                    'unit' => [
+                        'type' => 'string',
+                        'enum' => ['celsius', 'fahrenheit'],
+                        'description' => 'The temperature unit',
+                    ],
+                ],
+                'required' => ['location', 'unit'],
+            ]
+        );
+
+        $result = AiClient::prompt('What is the weather in Paris, France? Use celsius.')
+            ->usingProvider('anthropic')
+            ->usingFunctionDeclarations($getWeather)
+            ->generateTextResult();
+
+        $this->assertInstanceOf(GenerativeAiResult::class, $result);
+
+        $functionCall = $this->extractFunctionCall($result);
+        $this->assertNotNull($functionCall, 'Expected a function call in the response');
+        $this->assertEquals('get_weather', $functionCall->getName());
+
+        $args = $functionCall->getArgs();
+        $this->assertIsArray($args);
+        $this->assertArrayHasKey('location', $args);
+        $this->assertArrayHasKey('unit', $args);
+        $this->assertStringContainsStringIgnoringCase('paris', $args['location']);
+        $this->assertEquals('celsius', $args['unit']);
+    }
+
+    /**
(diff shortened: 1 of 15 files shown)


tests/integration/bootstrap.php
```diff
@@ -0,0 +1,21 @@
...
+require_once dirname(__DIR__, 2) . '/vendor/autoload.php';
+
+use Symfony\Component\Dotenv\Dotenv;
+
+$envFile = dirname(__DIR__, 2) . '/.env';
+if (file_exists($envFile)) {
+    $dotenv = new Dotenv();
+    // Enable putenv() so getenv() works (used by ProviderRegistry for API keys)
+    $dotenv->usePutenv(true);
+    $dotenv->load($envFile);
```
felixarntz, line 20: Do we need to add something here that sets up an HTTP client? There isn't any by default I believe?

Because of this, currently a bunch of tests don't work for me, particularly for Anthropic.

JasonTheAdams: Good catch! Fixed in #179 

We're otherwise using the cURL http package which is auto-discovered, so we don't need anything here.

tests/integration/Anthropic/FunctionCallingIntegrationTest.php
```diff
@@ -0,0 +1,172 @@
...
+     * Note: This tests a function where the parameter schema is a simple string,
+     * not an object with properties. This is a less common but valid use case.
+     */
+    public function testFunctionCallingWithStringParameter(): void
+    {
+        $sayHiTo = new FunctionDeclaration(
+            'say_hi_to',
+            'Says hi to a specific person by name',
+            [
+                'type' => 'string',
```
felixarntz, line 122: This test fails for me, and I think it should? According to what I could gather from the docs, Anthropic requires `object` as type. So this shouldn't be expected to pass?

tests/integration/OpenAi/FunctionCallingIntegrationTest.php
```diff
@@ -0,0 +1,172 @@
...
+     * Note: This tests a function where the parameter schema is a simple string,
+     * not an object with properties. This is a less common but valid use case.
+     */
+    public function testFunctionCallingWithStringParameter(): void
+    {
+        $sayHiTo = new FunctionDeclaration(
+            'say_hi_to',
+            'Says hi to a specific person by name',
+            [
+                'type' => 'string',
```
felixarntz, line 122: Same here, for OpenAI I don't think this is allowed. So maybe we'll want to expect an exception here? Just to make sure we still cover it for these providers that don't support it.

felixarntz (approved):
@JasonTheAdams Almost LGTM, just two things.

Other than that, for the Anthropic `testFunctionCallingWithNoArguments`, I know that'll be fixed in the other PR.

Preemptively approving, pending my feedback below.

tests/integration/Anthropic/FunctionCallingIntegrationTest.php
```diff
@@ -0,0 +1,172 @@
...
+        $this->assertInstanceOf(GenerativeAiResult::class, $result);
+
+        $functionCall = $this->extractFunctionCall($result);
+        $this->assertNotNull($functionCall, 'Expected a function call in the response');
+        $this->assertEquals('say_hi', $functionCall->getName());
+
+        // Args should be null or empty for a no-argument function
+        $args = $functionCall->getArgs();
+        $this->assertTrue(
+            $args === null || $args === [] || $args === (object) [],
```
felixarntz, line 105: I think we should make this more concrete, i.e. let's see what the API actually returns in this case. Unless it varies between distinct requests, we should expect that concrete result. While this is extremely nit-picky, we'll want to know when something related to this changes in the API, so being as precise as possible helps here, I'd argue.

tests/integration/Google/FunctionCallingIntegrationTest.php
```diff
@@ -0,0 +1,172 @@
...
+        $this->assertInstanceOf(GenerativeAiResult::class, $result);
+
+        $functionCall = $this->extractFunctionCall($result);
+        $this->assertNotNull($functionCall, 'Expected a function call in the response');
+        $this->assertEquals('say_hi', $functionCall->getName());
+
+        // Args should be null or empty for a no-argument function
+        $args = $functionCall->getArgs();
+        $this->assertTrue(
+            $args === null || $args === [] || $args === (object) [],
```
felixarntz, line 105: See above.

tests/integration/OpenAi/FunctionCallingIntegrationTest.php
```diff
@@ -0,0 +1,172 @@
...
+        $this->assertInstanceOf(GenerativeAiResult::class, $result);
+
+        $functionCall = $this->extractFunctionCall($result);
+        $this->assertNotNull($functionCall, 'Expected a function call in the response');
+        $this->assertEquals('say_hi', $functionCall->getName());
+
+        // Args should be null or empty for a no-argument function
+        $args = $functionCall->getArgs();
+        $this->assertTrue(
+            $args === null || $args === [] || $args === (object) [],
```
felixarntz, line 105: See above.

tests/integration/Anthropic/FunctionCallingIntegrationTest.php
JasonTheAdams: Good note! I'm going to address and get all function-related tests passing in a subsequent PR. 👍

tests/integration/OpenAi/FunctionCallingIntegrationTest.php
JasonTheAdams: I don't think it is. Hahah! But I'm starting with stupid tests and will fix/add/address them in the next PR.

tests/integration/Anthropic/FunctionCallingIntegrationTest.php
JasonTheAdams: Noted! I'll address this subsequently.

PR #179, opened 2026-01-22
Fix: correctly sets the transporter on subsequent try

JasonTheAdams:
Turns out when we caught and retried setting the transporter we weren't setting the variable. This would cause the first alphabetical Provider (Anthropic) to not have access to the transporter. Oops!

This was caught by #176 — proof that it's valuable already! 😆


felixarntz (approved):
@JasonTheAdams OMG, great catch! 😂

PR #183, opened 2026-01-23
Fix: Send text-only messages as plain string in Responses API Fixes #182

### LINKED ISSUE #182: Bug: OpenAI Responses API rejects text-only messages with 'input_text' type
## Description

When sending text-only messages to the OpenAI Responses API (`/v1/responses`), the API returns a 400 error:

```
Bad Request (400) - Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
```

Before fix: Image

After Fix:
Image

## Root Cause

The `getMessageInputItem()` method in `OpenAiTextGenerationModel.php` sends text-only messages as objects with a `type` field:

```php
// Current (broken) behavior
'content' => [
    'type' => 'input_text',  // ❌ API rejects this
    'text' => $text
]
```

However, the OpenAI Responses API expects text-only messages to have the text sent **directly as a string** in the `content` field, not as an object.

## Expected Behavior

For text-only messages, the `content` field should be a plain string:

```php
// Correct format
'content' => $text  // ✅ Plain string, not an object
```

## Steps to Reproduce

1. Create a text-only `UserMessage` with a single text part
2. Send it via `OpenAiTextGenerationModel::generateTextResult()`
3. API returns 400 error: "Invalid value: 'input_text'"

## Proposed Fix

In `src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php`, modify `getMessageInputItem()` to handle text-only messages specially:

```php
protected function getMessageInputItem(Message $message): ?array
{
    $parts = $message->getParts();

    if (empty($parts)) {
        return null;
    }

    // Special handling for text-only messages: send text directly as a string
    if (count($parts) === 1 && $parts[0]->getType()->isText()) {
        $text = $parts[0]->getText();
        return [
            'role' => $this->getMessageRoleString($message->getRole()),
            'content' => $text, // Send text directly as a string
        ];
    }

    // ... rest of the method for messages with files, function calls, etc.
}
```

## Environment

- **php-ai-client version:** 0.4.0
- **OpenAI API:** Responses API (`/v1/responses`)
- **PHP version:** 7.4+

## Additional Notes

- Messages with files or multiple parts should continue to use the object format with `type` fields
- Only single-part text messages need this special handling
- This is specific to the Responses API - Chat Completions API may have different requirements

## Related

This issue was discovered when upgrading from 0.3.x to 0.4.0 (which switched from Chat Completions to Responses API).

jmarx:
Fix bug that appears when trying to send a text-only message through the OpenAI Responses API. The API rejects messages with `type: 'input_text'` and expects plain strings for text-only content. This fix sends text-only messages as strings while maintaining object format for messages with files.

Fixes #182

### DIFF
--- a/src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
+++ b/src/ProviderImplementations/OpenAi/OpenAiTextGenerationModel.php
@@ -247,6 +247,15 @@ class OpenAiTextGenerationModel extends AbstractApiBasedModel implements TextGen
             return null;
         }
 
+        // Special handling for text-only messages: send text directly as a string
+        if (count($parts) === 1 && $parts[0]->getType()->isText()) {
+            $text = $parts[0]->getText();
+            return [
+                'role' => $this->getMessageRoleString($message->getRole()),
+                'content' => $text, // Send text directly as a string
+            ];
+        }
+
         $content = [];
         foreach ($parts as $part) {
             $partData = $this->getMessagePartData($part);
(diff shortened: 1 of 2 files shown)


felixarntz:
Per https://github.com/WordPress/php-ai-client/issues/182#issuecomment-3796609610, this doesn't truly address the problem. I've opened #184 to fix it.

PR #187, opened 2026-01-29
Fix object cloning to also clone nested objects (deep cloning)

JasonTheAdams:
In https://github.com/WordPress/wp-ai-client/pull/49 we realized that DTOs in this don't clone deeply. It's certainly useful to be able to clone DTOs as they're meant to represent a single, serializable group of data. One could be cloned via `toArray()` and `fromArray()` but that's awkward and doesn't work when cloning via PHP.

This adds support for deep cloning as well as tests to check that cloned DTOs don't reference old objects.


felixarntz (approved):
@JasonTheAdams Generally looks good, but check the test failures.

PR #189, opened 2026-02-03
Fix: corrects more DTO cloning

JasonTheAdams:
As a follow up to #187, there were more DTOs that needed further deep cloning support. Ironically, I originally set out to do this due to wanting to clone `PromptBuilder` and missed that. 🤦‍♂️ 

I _think_ this should be it for DTOs that needed deep cloning support.


felixarntz (approved):
great catch!

PR #191, opened 2026-02-04
Removes specific provider implementations

JasonTheAdams:
This removes the specific provider implementations we've so far had for testing. The providers will be moved to their own repositories which can be used as standalone packages or installed as WordPress plugins. The plugins will be submitted to the WP plugin repository as well.

The packages have been added to this repository as dev dependencies, this is to use them for integration tests.

Here are the new repositories:

* https://github.com/WordPress/google-ai-provider
* https://github.com/WordPress/openai-ai-provider
* https://github.com/WordPress/anthropic-ai-provider


felixarntz (changes requested):
I don't think we should merge this until we've ported all three of them elsewhere? There needs to be a smooth transition path, and as long as people can't get the providers elsewhere, their experience will just break without any reasonable path forward.

Marking as requested changes just until we're ready to merge this.

JasonTheAdams:
Agreed, @felixarntz! I just pushed the code to each repo along with a PR to review for each one. I'm going to update this composer to look at the implementing branch for each one for now so I can at least get the tests passing.

justlevine:
(not a decision that needs to be made now or even prior to 7.0 but want to confirm y'all considered releasing from a monorepo a la WordPress/performance , especially as the number of supported providers will hopefully be more than 3 and repo overhead adds up. If not, might be a topic to bring up once the near-term deadlines have passed)

JasonTheAdams:
All three providers have had their initial PRs merged and 1.0.0 is released to Packagist.

felixarntz (approved):

PR #197, opened 2026-02-12
Pin github actions to specific versions

### LINKED ISSUE #196: Pin Github Actions to a full hash
The github actions currently in `php-lint` and `php-test` are version numbers for third party actions. Specifying a full length commit SHA will use the 3rd party action as an immutable release, ensuring the workflows within the repository are not affected by upstream security problems should they occur.

aaronjorbin:
This pins each of the 3rd party actions to a specific hash to ensure it is immutable.

Also updates a few actions to the most recent version.

Fixes #196.

### DIFF
--- a/.github/workflows/php-lint.yml
+++ b/.github/workflows/php-lint.yml
@@ -38,9 +38,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: shivammathur/setup-php@v2
+      - uses: shivammathur/setup-php@44454db4f0199b8b9685a5d763dc37cbf79108e1 # 2.36.0
         with:
           php-version: 'latest'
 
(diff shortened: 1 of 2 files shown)


JasonTheAdams:
Hmm, I've never seen this before. This is a bit of a two-sided sword. If the locked version has a security issue, then we'll need to be made aware and manually update the version. Whereas if we stay on the major version then we risk new security issues, but also benefit from subsequent patches.

My inclination would be to keep tied to a major version. Curious what others think!

desrosj (approved):
This is something that's recommended in [GitHub's Secure Use documentation](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) and many of the widely used security scanning tools for GitHub Actions (such as [Zizmor](https://docs.zizmor.sh/audits/#unpinned-uses)) flag unpinned references. This has been widely implemented across repositories under the WordPress organization. 

While it is possible a pinned version could have a security vulnerability, it's unfortunately quite common for third-party actions to be compromised. For a few recent examples, see the [changed-files](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/) and [reviewdog's action library](https://www.stepsecurity.io/blog/reviewdog-github-actions-are-compromised).

My understanding is that GitHub Action repositories have a branch corresponding to each version branch, and using `@v3` will always pull the latest commit in that branch, which is not necessarily a tagged version.

If the concern is an inability to keep track of available updates, then a `dependabot.yml` file can be added to automatically open pull requests when updates are available. The bot is even smart enough to update the inline comment as well.

felixarntz (approved):

PR #200, opened 2026-02-12
add php 8.5 to test matrix

### LINKED ISSUE #198: Add PHP 8.5 to testing
[WordPress core supports PHP 8.5](https://make.wordpress.org/core/2025/11/21/php-8-5-support-in-wordpress-6-9/) and since there is a goal of including this in WordPress, it's valuable to have tests running on that version of PHP

aaronjorbin:
Fixes #198

### DIFF
--- a/.github/workflows/php-test.yml
+++ b/.github/workflows/php-test.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ['7.4', '8.0', '8.4']
+        php-version: ['7.4', '8.0', '8.4', '8.5']
 
     steps:
     - uses: actions/checkout@v4


JasonTheAdams (commented):
@aaronjorbin Thanks! Glad it passes! Do you think there's a reason to keep 8.4 if 8.5 is passing? Is there a scenario where 8.5 would pass and 8.4 would not?

justlevine:
> Do you think there's a reason to keep 8.4 if 8.5 is passing? Is there a scenario where 8.5 would pass and 8.4 would not?

It's a FOSS repo, and the CI minutes are free, so I'd suggest we test against _every_ PHP version we support. 

Any new 8.5 feature has the possibility to cause 8.4 to fail if not properly gated, for example if we use [property overrides to gate](https://wiki.php.net/rfc/override_properties)

aaronjorbin:
While I think it's unlikely that something would pass on 8.5 and fail on 8.4, testing helps ensure that if there is an issue, it gets caught early. I think in most cases if it passes on 8.5 and fails on 8.4, it will also fail on 8.0. The most likely scenarios I can think of are either miscoded polyfills or bugs in version detection/gating. 

While there is no monetary cost, enterprises are limited to 500 concurent actions. I don't think that is something we need to worry about here since these are so quick right now, but is something to keep in mind.

JasonTheAdams (approved):
Sounds good! Certainly not a big deal. Thanks for adding this!

PR #194, opened 2026-02-09
Feature/token limit reached exception

saarnilauri:
This PR adds explicit token-limit handling for Anthropic so truncated outputs are surfaced as exceptions instead of normal results.

### What changed
- Added `TokenLimitReachedException` (`src/Common/Exception/TokenLimitReachedException.php`) with structured metadata:
  - `maxTokens`
  - `providerStopReason`
  - optional `functionName`
  - optional `missingRequiredParameters`
- Updated Anthropic response parsing (`src/ProviderImplementations/Anthropic/AnthropicTextGenerationModel.php`) to throw `TokenLimitReachedException` when `stop_reason` is:
  - `max_tokens`
  - `model_context_window_exceeded`
- Added tool-call diagnostics for partial argument payloads by comparing returned args with required schema fields.
- Added Anthropic unit test scaffolding and initial coverage:
  - `tests/unit/ProviderImplementations/Anthropic/MockAnthropicTextGenerationModel.php`
  - `tests/unit/ProviderImplementations/Anthropic/AnthropicTextGenerationModelTest.php`
- Added Anthropic integration tests for max-token scenarios (plain text + tool-calling):
  - `tests/integration/Anthropic/TextGenerationIntegrationTest.php`
  - `tests/integration/Anthropic/FunctionCallingIntegrationTest.php`

  This relates to https://github.com/WordPress/php-ai-client/issues/193
  
I know that the providers will be moved to their own repositories. But as this change is still in very early stage. I decided to do this PR to the client repository, so that discussion on how this should be implemented can start.

The PR is co-authored with OpenAIs Codex 5.3 model.

PR #203, opened 2026-02-13
Adds TokenLimitReachedException for providers

JasonTheAdams:
## Summary
- Adds TokenLimitReachedException for providers to throw when a prompt exceeds the model's context window
   or a configured token maximum
  - Includes an optional $maxTokens parameter so callers catching the exception can programmatically       
  inspect the limit that was hit (e.g., to retry with a shorter prompt or surface a meaningful error to end
   users)

This is a smaller follow up to #194, as the providers are no longer bundled. I didn't bring over all the parameters for the exception in that PR because I wasn't sure what they were all for. Happy to add them back if there's strong reasoning!

### DIFF
--- a/src/Common/Exception/TokenLimitReachedException.php
+++ b/src/Common/Exception/TokenLimitReachedException.php
@@ -11,14 +11,14 @@ namespace WordPress\AiClient\Common\Exception;
  * exceeds the allowed limit, whether that is the model's context window
  * or a configured maximum.
  *
- * @since 0.4.0
+ * @since n.e.x.t
  */
 class TokenLimitReachedException extends RuntimeException
 {
     /**
      * The token limit that was reached, if known.
      *
-     * @since 0.4.0
+     * @since n.e.x.t
      *
      * @var int|null
      */
@@ -27,7 +27,7 @@ class TokenLimitReachedException extends RuntimeException
     /**
      * Creates a new TokenLimitReachedException.
      *
-     * @since 0.4.0
+     * @since n.e.x.t
      *
      * @param string         $message   The exception message.
      * @param int|null       $maxTokens The token limit that was reached, if known.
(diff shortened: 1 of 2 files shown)


src/Common/Exception/TokenLimitReachedException.php
```diff
@@ -0,0 +1,54 @@
...
+namespace WordPress\AiClient\Common\Exception;
+
+/**
+ * Exception thrown when a token limit is reached during prompt fulfillment.
+ *
+ * Providers should throw this exception when the token usage for a request
+ * exceeds the allowed limit, whether that is the model's context window
+ * or a configured maximum.
+ *
+ * @since 0.4.0
```
felixarntz: Use n.e.x.t throughout the PR

felixarntz (approved):
LGTM! Just one thing to fix pre merge.

saarnilauri:
@JasonTheAdams here is explanation for the parameters that I added in #194 

I added `$providerStopReason` as reaching max tokens is not the only case when AI cuts of output. There is also the case for the when the Anthropic API return `model_context_window_exceeded` as a stop reason. I thought this param can help the developers to identify different cases.

The idea for `$functionName` was to highlight if the max tokens was the stop reason when the AI was doing a tool call and the output was cut out. I thought that this could help developers help to identify a tool call that makes AI output tokens over the limit.

The idea for `$missingRequiredParameters` is more detailed addition to `$functionName`. It would list all the required parameters for the tool call that was interupted due reaching the token limit. When Anthropic API is forming the parameters for the tool to be used and the token limit runs full the AI return the tool call but with only the parameters that is was able to fully generate. So, `$missingRequiredParameters` could help the developer to understand what part of the tool call is consuming the tokens and are missing.

JasonTheAdams:
Thanks for the context, @saarnilauri! Those do make sense, but feel a bit nice-to-have at the moment. Let's start simple and gather more feedback over time.

PR #208, opened 2026-02-20
Adds built-in support for basic PSR7 and 17 support

JasonTheAdams:
This introduces basic support for PSR7 and PSR17 along with an abstract discovery class. This way implementing systems can use these built-in classes so they don't have to introduce their own and can simply create their own HTTP Client.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,12 @@
     "require": {
         "php": ">=7.4",
         "ext-json": "*",
+        "nyholm/psr7": "^1.8",
         "php-http/discovery": "^1.0",
         "php-http/httplug": "^2.0",
+        "psr/event-dispatcher": "^1.0",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -22,11 +22,12 @@
     "require": {
         "php": ">=7.4",
         "ext-json": "*",
+        "nyholm/psr7": "^1.8",
         "php-http/discovery": "^1.0",
         "php-http/httplug": "^2.0",
+        "psr/event-dispatcher": "^1.0",
```
felixarntz, line 28: Super nit-pick, but since you're sorting this differently now, can you make it alphabetically?

JasonTheAdams: They are. 😄

felixarntz (approved):

PR #210, opened 2026-02-25
Adds optional description to providers

JasonTheAdams:
This adds "description" as an optional field for providers. This is useful in contexts where providers are being listed out for users.

### DIFF
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -19,6 +19,7 @@ use WordPress\AiClient\Providers\Http\Enums\RequestAuthenticationMethod;
  * @phpstan-type ProviderMetadataArrayShape array{
  *     id: string,
  *     name: string,
+ *     description?: ?string,
  *     type: string,
  *     credentialsUrl?: ?string,
  *     authenticationMethod?: ?string
@@ -30,6 +31,7 @@ class ProviderMetadata extends AbstractDataTransferObject
 {
     public const KEY_ID = 'id';
     public const KEY_NAME = 'name';
+    public const KEY_DESCRIPTION = 'description';
     public const KEY_TYPE = 'type';
     public const KEY_CREDENTIALS_URL = 'credentialsUrl';
     public const KEY_AUTHENTICATION_METHOD = 'authenticationMethod';
@@ -43,6 +45,11 @@ class ProviderMetadata extends AbstractDataTransferObject
      * @var string The provider's display name.
      */
     protected string $name;
+
+    /**
+     * @var string|null The provider's description.
+     */
+    protected ?string $description;
 
     /**
      * @var ProviderTypeEnum The provider type.
@@ -69,16 +76,19 @@ class ProviderMetadata extends AbstractDataTransferObject
      * @param ProviderTypeEnum $type The provider type.
      * @param string|null $credentialsUrl The URL where users can get credentials.
      * @param RequestAuthenticationMethod|null $authenticationMethod The authentication method.
+     * @param string|null $description The provider's description.
      */
     public function __construct(
         string $id,
         string $name,
         ProviderTypeEnum $type,
         ?string $credentialsUrl = null,
-        ?RequestAuthenticationMethod $authenticationMethod = null
+        ?RequestAuthenticationMethod $authenticationMethod = null,
+        ?string $description = null
     ) {
         $this->id = $id;
         $this->name = $name;
+        $this->description = $description;
         $this->type = $type;
         $this->credentialsUrl = $credentialsUrl;
         $this->authenticationMethod = $authenticationMethod;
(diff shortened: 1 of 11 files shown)


src/Providers/DTO/ProviderMetadata.php
```diff
@@ -69,16 +76,19 @@ class ProviderMetadata extends AbstractDataTransferObject
...
+     * @param string|null $description The provider's description.
      */
     public function __construct(
         string $id,
         string $name,
         ProviderTypeEnum $type,
         ?string $credentialsUrl = null,
-        ?RequestAuthenticationMethod $authenticationMethod = null
+        ?RequestAuthenticationMethod $authenticationMethod = null,
+        ?string $description = null
```
felixarntz, line 89: This is getting out of hand in terms of parameters :)

Can we change this to a single param? Either some value object with these, or just the basic `$args` pattern. Don't love the latter, but would be fine - another class for this seems overkill. I think the first 3 parameters should remain as-is, but the optional ones should be combined.

We can make it backward compatible by supporting _either_ an array for the 4th parameter, or what we currently have for the 4th and 5th parameter. And use a `trigger_error` for deprecation warning if the old way is used.

Then 2-3 releases later we can remove the old way.

Please make sure to keep tests around for both behaviors to have BC covered but also the new correct way.

felixarntz (commented):
Looks great! Though this prompts for a slight refactoring of `ProviderMetadata` constructor.

src/Providers/DTO/ProviderMetadata.php
JasonTheAdams: Hahah! Yeah... what we really want here is PHP 8.0 named parameters. In the meantime, I think I'd prefer `$args`. I don't love it, but I find VO instantiation to be even more awkward, especially for a DTO like this. It makes the parameters feel weirdly complicated.

So how about this as the new signature: `__construct(string $id, string $name, array $args = [])`

felixarntz: Works for me, I think. So that means we'll have a default type of Cloud provider then, right?

JasonTheAdams: Oops! Missed that... I mean, yeah, default cloud provider. That's why I left that out... 😁

JasonTheAdams: @felixarntz Tentatively resolved in b2228bd

I'm honestly not convinced that I like this better. It's awkward having a growing list of parameters, for sure, but if we eventually bump to 8.0 I'd actually want to go back to pure parameters as it is now. Types and stuff get really goofy. It feels like we're temporarily weakening types to make the DX _slightly_ better. I'm tempted to revert this and just live with it until named parameters are an option.

src/Providers/DTO/ProviderMetadata.php
```diff
@@ -62,26 +75,51 @@ class ProviderMetadata extends AbstractDataTransferObject
...
+            ];
+            if (isset($allArgs[3])) {
+                $args[self::KEY_CREDENTIALS_URL] = $allArgs[3];
+            }
+            if (isset($allArgs[4])) {
+                $args[self::KEY_AUTHENTICATION_METHOD] = $allArgs[4];
+            }
+            if (isset($allArgs[5])) {
+                $args[self::KEY_DESCRIPTION] = $allArgs[5];
+            }
```
felixarntz: This last one is unnecessary. It was never possible in the old way to pass a description anyway. We should not add support for it in the wrong way that we want to discourage anyway.

src/Providers/DTO/ProviderMetadata.php
```diff
@@ -108,6 +146,18 @@ public function getName(): string
         return $this->name;
     }
 
+    /**
+     * Gets the provider's description.
+     *
+     * @since 0.5.0
```
felixarntz: n.e.x.t - please double check that throughout the PR

felixarntz (approved):
Almost good to go! Preemptively approving, but please address the remaining feedback before merge.

src/Providers/DTO/ProviderMetadata.php
felixarntz: Hmm, fair point. Current DX is IMO worse without the array (functions with more than 2-3 params are generally a problematic pattern for DX). Though you can also argue against the array of `$args` because a lack of auto-completion, and less idiomatic for future PHP compat indeed.

Let's revert - you have me convinced :)

src/Providers/DTO/ProviderMetadata.php
JasonTheAdams: Reverted in 3154324

src/Providers/DTO/ProviderMetadata.php
JasonTheAdams: Reverted in 3154324

src/Providers/DTO/ProviderMetadata.php
JasonTheAdams: Resolved in 4b49f2b

PR #213, opened 2026-02-27
Fix: SupportedOption::isSupportedValue() fails after Redis deserialization

### LINKED ISSUE #212: SupportedOption::isSupportedValue() fails when model metadata is retrieved from a persistent object cache (Redis)
## Summary

`SupportedOption::isSupportedValue()` uses strict (`===`) array comparison which breaks when `AbstractEnum` singleton instances are deserialized from a persistent object cache (Redis, Memcached). This causes model discovery to fail with "No models found for provider X that support Y for this prompt" even though the provider is correctly configured and has matching models.

## Environment

- WordPress with a persistent object cache (Redis via `wp-redis` in my case)
- `wp-ai-client` using `WordPress_Cache` (PSR-16 adapter over `wp_cache_*`)
- Anthropic provider with valid API key

## Steps to reproduce

1. Configure a WordPress site with Redis object cache
2. Register the Anthropic provider and set valid API credentials
3. Use `AiClient::prompt("Say hello")->usingProvider("anthropic")->generateTextResult()`

**Expected:** Text generation succeeds.

**Actual:** Throws `InvalidArgumentException: No models found for provider "anthropic" that support text_generation for this prompt.`

Flushing the object cache (`wp cache flush`) fixes it for a single request, but the issue recurs on the next request once model metadata is cached again.

## Root cause

The issue is in `SupportedOption::isSupportedValue()` (`src/Providers/Models/DTO/SupportedOption.php`):

```php
public function isSupportedValue($value): bool
{
    if ($this->supportedValues === null) {
        return true;
    }

    if (is_array($value)) {
        sort($value);
        foreach ($this->supportedValues as $supportedValue) {
            if (!is_array($supportedValue)) {
                continue;
            }
            sort($supportedValue);
            if ($value === $supportedValue) { // <-- strict comparison
                return true;
            }
        }
        return false;
    }

    return in_array($value, $this->supportedValues, true); // <-- strict comparison
}
```

The `===` comparison requires identical object instances. This works when all `AbstractEnum` values are singletons within the same request, but breaks when model metadata round-trips through a serializing cache:

1. `WithDataCachingTrait::cached()` stores model metadata via the PSR-16 cache (`WordPress_Cache` -> `wp_cache_set` -> Redis)
2. Redis serializes the data with PHP `serialize()`
3. On the next request (or even within the same request on cache hit), `unserialize()` creates **new** `AbstractEnum` instances — they have the same `$value` and `$name` properties but are different objects
4. `ModalityEnum::text()` (fresh singleton) `!==` unserialized `ModalityEnum` (new instance)
5. `[ModalityEnum::text()] === [unserializedModalityEnum]` evaluates to `false`
6. `areMetBy()` returns `false` for every model, so `findProviderModelsMetadataForSupport()` returns an empty list

`AbstractEnum` has a `final private` constructor and no `__wakeup()` method, so `unserialize()` bypasses the singleton registry entirely.

## Suggested fix

Replace strict object identity comparison with value-based comparison in `isSupportedValue()`:

```php
public function isSupportedValue($value): bool
{
    if ($this->supportedValues === null) {
        return true;
    }

    if (is_array($value)) {
        $normalizedValue = self::normalizeForComparison($value);
        foreach ($this->supportedValues as $supportedValue) {
            if (!is_array($supportedValue)) {
                continue;
            }
            if ($normalizedValue === self::normalizeForComparison($supportedValue)) {
                return true;
            }
        }
        return false;
    }

    $normalizedValue = self::normalizeScalar($value);
    foreach ($this->supportedValues as $supported) {
        if ($normalizedValue === self::normalizeScalar($supported)) {
            return true;
        }
    }
    return false;
}

private static function normalizeForComparison(array $items): array
{
    $normalized = array_map([self::class, 'normalizeScalar'], $items);
    sort($normalized);
    return $normalized;
}

private static function normalizeScalar($value)
{
    if ($value instanceof \WordPress\AiClient\Common\AbstractEnum) {
        return $value->value;
    }
    return $value;
}
```

Alternatively, `AbstractEnum` could implement `__wakeup()` to re-register unserialized instances into the singleton cache, preserving identity after deserialization.

## Workaround

Prevent the AI client from using the persistent cache so it falls back to the in-memory local cache in `WithDataCachingTrait`:

```php
// Do not pass WordPress_Cache to AiClient so it uses in-memory local cache
// AiClient::setCache( new WordPress_Cache() );
```

Or flush the `wp_ai_client` cache group, accepting that the bug will recur once the cache is repopulated.

chloe-pomegranate:
## Summary

- Fixes `SupportedOption::isSupportedValue()` failing when `AbstractEnum` singletons are deserialized from a persistent object cache (Redis/Memcached)
- Normalizes `AbstractEnum` instances to their string `->value` before comparison, so object identity (`===`) no longer breaks after `unserialize()`
- Adds tests verifying both scalar and array enum values survive serialize/unserialize round-trips

Closes #212

## Test plan

- [ ] `vendor/bin/phpunit --filter SupportedOptionTest` — passing
- [ ] `vendor/bin/phpcs --standard=phpcs.xml.dist src/Providers/Models/DTO/SupportedOption.php` — clean
- [ ] `vendor/bin/phpstan analyse src/Providers/Models/DTO/SupportedOption.php` — no errors
- [ ] Verify on a WordPress site with Redis object cache that model discovery no longer fails after cache is populated

### DIFF
--- a/src/Providers/Models/DTO/SupportedOption.php
+++ b/src/Providers/Models/DTO/SupportedOption.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace WordPress\AiClient\Providers\Models\DTO;
 
 use WordPress\AiClient\Common\AbstractDataTransferObject;
+use WordPress\AiClient\Common\AbstractEnum;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Providers\Models\Enums\OptionEnum;
 
@@ -87,20 +88,65 @@ class SupportedOption extends AbstractDataTransferObject
 
         // If the value is an array, consider it a set (i.e. order doesn't matter).
         if (is_array($value)) {
-            sort($value);
+            $normalizedValue = self::normalizeArrayForComparison($value);
             foreach ($this->supportedValues as $supportedValue) {
                 if (!is_array($supportedValue)) {
                     continue;
                 }
-                sort($supportedValue);
-                if ($value === $supportedValue) {
+                $normalizedSupported = self::normalizeArrayForComparison($supportedValue);
+                if ($normalizedValue === $normalizedSupported) {
                     return true;
                 }
             }
             return false;
         }
 
-        return in_array($value, $this->supportedValues, true);
+        $normalizedValue = self::normalizeValue($value);
+        foreach ($this->supportedValues as $supportedValue) {
+            if (self::normalizeValue($supportedValue) === $normalizedValue) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Normalizes an AbstractEnum instance to its string value.
+     *
+     * This ensures comparisons work correctly even after deserialization
+     * (e.g. Redis/Memcached object cache), where AbstractEnum singletons
+     * are reconstructed as separate instances.
+     *
+     * @since 1.2.1
+     *
+     * @param mixed $value The value to normalize.
+     * @return mixed The normalized value.
+     */
+    private static function normalizeValue($value)
+    {
+        if ($value instanceof AbstractEnum) {
+            return $value->value;
+        }
+        return $value;
+    }
+
+    /**
+     * Normalizes and sorts an array for comparison.
+     *
+     * Maps each element through normalizeValue() and sorts the result,
+     * ensuring consistent comparison regardless of element order or
+     * AbstractEnum instance identity.
+     *
+     * @since 1.2.1
+     *
+     * @param array<mixed> $items The array to normalize.
+     * @return array<mixed> The normalized, sorted array.
+     */
+    private static function normalizeArrayForComparison(array $items): array
+    {
+        $normalized = array_map([self::class, 'normalizeValue'], $items);
+        sort($normalized);
+        return $normalized;
     }
 
     /**
--- a/tests/unit/Providers/Models/DTO/SupportedOptionTest.php
+++ b/tests/unit/Providers/Models/DTO/SupportedOptionTest.php
@@ -598,7 +598,8 @@ class SupportedOptionTest extends TestCase
         // Fresh singleton enums in arrays should still match
         $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::text()]));
         $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::text(), ModalityEnum::image()]));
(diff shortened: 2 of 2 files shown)


src/Providers/Models/DTO/SupportedOption.php
```diff
@@ -87,20 +88,65 @@ public function isSupportedValue($value): bool
...
+    }
+
+    /**
+     * Normalizes an AbstractEnum instance to its string value.
+     *
+     * This ensures comparisons work correctly even after deserialization
+     * (e.g. Redis/Memcached object cache), where AbstractEnum singletons
+     * are reconstructed as separate instances.
+     *
+     * @since 1.2.1
```
felixarntz, line 120: Just an FYI for any future contributions: We normally use `@since n.e.x.t` everywhere, so that we can dynamically inject the actual version prior to a release.

But since this coming release is indeed going to be 1.2.1, this is good as is 👍

tests/unit/Providers/Models/DTO/SupportedOptionTest.php
```diff
@@ -546,4 +547,59 @@ public function testJsonSchemaReflectsOptionalSupportedValues(): void
...
+            ]
+        );
+
+        // Simulate Redis/Memcached round-trip
+        $deserialized = unserialize(serialize($option));
+
+        // Fresh singleton enums in arrays should still match
+        $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::text()]));
+        $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::text(), ModalityEnum::image()]));
+        $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::image(), ModalityEnum::text()])); // Order shouldn't matter
```
felixarntz: Lines must be under 120 characters, per PHPCS config.
```suggestion
        // Order shouldn't matter
        $this->assertTrue($deserialized->isSupportedValue([ModalityEnum::image(), ModalityEnum::text()]));
```

felixarntz (approved):
@chloe-pomegranate Thank you for the PR, this looks excellent and implements the most suitable solution suggested in #212.

Appreciate you jumping in on this!

@saarnilauri Would be great if you could double check that this addresses the bug you reported in #212. Thank you!

saarnilauri:
> 
> @saarnilauri Would be great if you could double check that this addresses the bug you reported in #212. Thank you!

@felixarntz I can confirm that the fix addresses the bug. Thank you so much both of you!

PR #214, opened 2026-02-27
fix: Throw TokenLimitReachedException on truncated tool calls

### LINKED ISSUE #193: Missing Required Parameters in Tool Calls When Token Limit Reached
## Summary
When using `wp-ai-client`, and it uses `php-ai-client` if the AI reaches the token limit while generating a tool call with input parameters, it returns the tool call without all required parameters. This can cause multiple retries in the agent loop instead of properly handling the token limit condition.

## Current Behavior
- AI generates a tool call that requires input parameters
- Token limit is reached during parameter generation (e.g., while generating long content)
- Tool call is returned with incomplete parameters (missing required fields)
- Agent loop retries multiple times attempting to complete the request ( the agent loop is NOT part of the package)

## Expected Behavior
- Providers should throw a `TokenLimitReachedException` when the token limit is reached
- Exception should contain information about the current max tokens set for the provider
- This allows the plugin developpers agent to handle the exception gracefully instead of entering a retry loop

## Steps to Reproduce
1. Configure `wp-ai-client` with a token limit
2. Use a tool/ability that requires multiple parameters (e.g., `title`, `post_type`, `content`)
3. Generate content that approaches or exceeds the token limit (e.g., long markup for `content`)
4. Observe that the AI calls the ability with only partial parameters (e.g., only `title` and `post_type`, missing `content`)

## Example Case
**Tool:** WordPress ability for creating posts

**Required Parameters:**
- `title` (string)
- `post_type` (string)
- `content` (string)

**Scenario:**
When AI generates long markup for the `content` parameter and reaches the token limit, the tool call is made with only `title` and `post_type`, omitting the required `content` parameter.

## Proposed Solution
Implement proper exception handling at the provider level.

The provider shoulds detect when the token limit is reached during tool call generation and throw this exception instead of returning incomplete tool calls.

## Consideration of the client and provider separation

As the code for the providers will be moved to seperate repositories, there should be guideline to implement this error handling for the providers.

## Impact
- **Severity:** Medium-High
- **Affected Component:** Provider layer, tool/ability system
- **User Impact:** Poor user experience due to multiple retries, wasted API calls, and potentially incomplete operations

## Environment
- Packages: `wp-ai-client` and `php-ai-client`

chloe-pomegranate:
## Summary
Fixes #193.

- When an API response hits the max token limit mid-way through generating a tool call (`finish_reason=length` with `tool_calls` present), throw `TokenLimitReachedException` instead of silently returning a `FunctionCall` with null/incomplete arguments.
- Add a secondary `json_last_error()` check in `parseResponseChoiceMessageToolCallPart()` to catch malformed tool call JSON regardless of finish reason.
- Exception includes the configured `maxTokens` value so callers can handle gracefully (e.g. increase the limit and retry).

## Changes

**`src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php`**
- In `parseResponseChoiceToCandidate()`: detect `finish_reason=length` + `tool_calls` present in raw message data and throw `TokenLimitReachedException` before parsing the message.
- In `parseResponseChoiceMessageToolCallPart()`: replace silent `json_decode` with an explicit `json_last_error()` check that throws `ResponseException` on invalid JSON.

**`tests/unit/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModelTest.php`**
- `testParseResponseChoiceToCandidateThrowsOnLengthWithToolCalls` — truncated tool call throws exception.
- `testParseResponseChoiceToCandidateTokenLimitIncludesMaxTokens` — exception carries configured maxTokens (500).
- `testParseResponseChoiceToCandidateLengthWithoutToolCallsDoesNotThrow` — text-only + length still works.
- `testParseResponseChoiceToCandidateToolCallsFinishReasonDoesNotThrow` — normal tool calls unaffected.

## Test plan
- [x] `vendor/bin/phpunit --filter "testParseResponseChoiceToCandidate"` — all 12 tests pass.
- [x] `vendor/bin/phpunit` — full suite passes (1059 tests, 3921 assertions).
- [x] `vendor/bin/phpstan analyse` — no errors on source file.

> **Note:** The Anthropic and Google providers would benefit from similar checks for their truncation signals (`stop_reason=max_tokens` and `finishReason=MAX_TOKENS` respectively).

### DIFF
--- a/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
+++ b/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
@@ -6,6 +6,7 @@ namespace WordPress\AiClient\Providers\OpenAiCompatibleImplementation;
 
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
+use WordPress\AiClient\Common\Exception\TokenLimitReachedException;
 use WordPress\AiClient\Messages\DTO\Message;
 use WordPress\AiClient\Messages\DTO\MessagePart;
 use WordPress\AiClient\Messages\Enums\MessagePartChannelEnum;
@@ -658,6 +659,23 @@ abstract class AbstractOpenAiCompatibleTextGenerationModel extends AbstractApiBa
         }
 
         $messageData = $choiceData['message'];
+
+        if (
+            'length' === $choiceData['finish_reason']
+            && isset($messageData['tool_calls'])
+            && is_array($messageData['tool_calls'])
+            && count($messageData['tool_calls']) > 0
+        ) {
+            throw new TokenLimitReachedException(
+                sprintf(
+                    '%s response was truncated while generating a tool call.'
+                    . ' Increase the max tokens or simplify the prompt.',
+                    $this->providerMetadata()->getName()
+                ),
+                $this->getConfig()->getMaxTokens()
+            );
+        }
+
         $message = $this->parseResponseChoiceMessage($messageData, $index);
 
         switch ($choiceData['finish_reason']) {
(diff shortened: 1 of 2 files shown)


chloe-pomegranate:
Pushed a new commit for coverage :)

PR #216, opened 2026-03-08
Add support for thought signatures and thought tokens

felixarntz:
This PR does two decoupled but related things:

1. Add support for thought signatures in message parts
    - This is technically more of a bug fix than an enhancement - it's a critical must-have. All of Anthropic, Google, and OpenAI rely heavily in thought signatures nowadays to validate that thought information passed back in a multi-turn chat was actually produced by their own APIs (rather than "made up" by the caller).
    - Contrary to what one might think, this is not only important for `thought` channel message parts. Several models will include thought signatures on actual content parts that (I suppose) involved thinking behind the scenes. For example, function calls commonly include them, but even images in multimodal output models can include them.
    - The rule is basically: If the API returns a thought signature with a message part, it _must_ be passed when using the same content part in a response. Otherwise, the API call may result in an error.
3. Add support for thought tokens in token usage data

Both changes are fully optional and therefore backward compatible. However, once this is launched, we should update the provider implementations to support both where applicable - especially critical for point 1.

### DIFF
--- a/tests/unit/Results/DTO/TokenUsageTest.php
+++ b/tests/unit/Results/DTO/TokenUsageTest.php
@@ -317,12 +317,12 @@ class TokenUsageTest extends TestCase
      */
     public function testCreateWithThoughtTokens(): void
     {
-        $tokenUsage = new TokenUsage(100, 50, 250, 100);
+        $tokenUsage = new TokenUsage(100, 50, 150, 20);
 
         $this->assertEquals(100, $tokenUsage->getPromptTokens());
         $this->assertEquals(50, $tokenUsage->getCompletionTokens());
-        $this->assertEquals(250, $tokenUsage->getTotalTokens());
-        $this->assertEquals(100, $tokenUsage->getThoughtTokens());
+        $this->assertEquals(150, $tokenUsage->getTotalTokens());
+        $this->assertEquals(20, $tokenUsage->getThoughtTokens());
     }
 
     /**
@@ -344,11 +344,11 @@ class TokenUsageTest extends TestCase
      */
     public function testToArrayIncludesThoughtTokens(): void
     {
-        $tokenUsage = new TokenUsage(100, 50, 250, 100);
+        $tokenUsage = new TokenUsage(100, 50, 150, 30);
         $array = $tokenUsage->toArray();
 
         $this->assertArrayHasKey(TokenUsage::KEY_THOUGHT_TOKENS, $array);
-        $this->assertEquals(100, $array[TokenUsage::KEY_THOUGHT_TOKENS]);
+        $this->assertEquals(30, $array[TokenUsage::KEY_THOUGHT_TOKENS]);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


gziolo (approved):
Looks good — clean, backward compatible, well-tested. All checks pass locally (tests, PHPCS clean, PHPStan clean). Left a minor nit on test values.

gziolo:
Minor nit on `tests/unit/Results/DTO/TokenUsageTest.php` line 320 and a few other places:

```php
$tokenUsage = new TokenUsage(100, 50, 250, 100);
```

`thoughtTokens` (100) exceeds `completionTokens` (50), which doesn't match the class docs stating thought tokens are "a subset of completion tokens." Might be worth adjusting for clarity — e.g. `new TokenUsage(100, 150, 250, 100)` — so the test values reinforce the documented invariant. Not blocking.

tests/unit/Results/DTO/TokenUsageTest.php
```diff
@@ -309,4 +309,127 @@ public function testStreamingResponseUsage(): void
...
+     * @return void
+     */
+    public function testCreateWithThoughtTokens(): void
+    {
+        $tokenUsage = new TokenUsage(100, 50, 250, 100);
+
+        $this->assertEquals(100, $tokenUsage->getPromptTokens());
+        $this->assertEquals(50, $tokenUsage->getCompletionTokens());
+        $this->assertEquals(250, $tokenUsage->getTotalTokens());
+        $this->assertEquals(100, $tokenUsage->getThoughtTokens());
```
felixarntz:
```suggestion
        $tokenUsage = new TokenUsage(100, 50, 150, 20);

        $this->assertEquals(100, $tokenUsage->getPromptTokens());
        $this->assertEquals(50, $tokenUsage->getCompletionTokens());
        $this->assertEquals(150, $tokenUsage->getTotalTokens());
        $this->assertEquals(20, $tokenUsage->getThoughtTokens());
```

tests/unit/Results/DTO/TokenUsageTest.php
```diff
@@ -309,4 +309,127 @@ public function testStreamingResponseUsage(): void
...
+     *
+     * @return void
+     */
+    public function testToArrayIncludesThoughtTokens(): void
+    {
+        $tokenUsage = new TokenUsage(100, 50, 250, 100);
+        $array = $tokenUsage->toArray();
+
+        $this->assertArrayHasKey(TokenUsage::KEY_THOUGHT_TOKENS, $array);
+        $this->assertEquals(100, $array[TokenUsage::KEY_THOUGHT_TOKENS]);
```
felixarntz:
```suggestion
        $tokenUsage = new TokenUsage(100, 50, 150, 30);
        $array = $tokenUsage->toArray();

        $this->assertArrayHasKey(TokenUsage::KEY_THOUGHT_TOKENS, $array);
        $this->assertEquals(30, $array[TokenUsage::KEY_THOUGHT_TOKENS]);
```

PR #217, opened 2026-03-08
Complete partial video generation foundation

felixarntz:
While we partially had the required pieces for video generation already present, for some reason we were missing part of the foundation. This PR fixes that.

This does not actually implement any video generation features - it just completes the API. It will allow providers to implement video support in the future.

Nothing special here, this entirely follows the existing patterns we already use for e.g. text generation, image generation, speech generation, ...

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -350,6 +350,7 @@ direction LR
             +generateImageResult() GenerativeAiResult
             +generateSpeechResult() GenerativeAiResult
             +convertTextToSpeechResult() GenerativeAiResult
+            +generateVideoResult() GenerativeAiResult
             +generateText() string
             +generateTexts(?int $candidateCount) string[]
             +generateImage() File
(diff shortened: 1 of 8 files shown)


gziolo (approved):
All checks pass locally (tests, PHPCS clean, PHPStan clean).

Verified the implementation follows the established patterns for speech/image generation:
- `VideoGenerationModelInterface` and `VideoGenerationOperationModelInterface` match the existing interface pairs (sync + async)
- `PromptBuilder` additions (`generateVideoResult`, `generateVideo`, `generateVideos`) follow the same flow: set output modality → delegate to `generateResult` with the correct capability
- Capability inference works from both the model interface and explicit output modality, consistent with other modalities
- The old "video not yet supported" throw is replaced with proper delegation, and the unsupported modality test now uses `document` instead
- Test coverage is thorough — mirrors the existing speech/image test structure across all code paths (result generation, error cases, shorthand helpers, capability inference, `isSupported` check)

Based on my research, the two-interface split (sync `ModelInterface` + async `OperationModelInterface`) is a good fit for providers like OpenAI/Sora where video generation is inherently asynchronous (job submission + polling), while the sync interface can wrap `create_and_poll` + download for simpler usage.

No issues found.

PR #218, opened 2026-03-08
Add missing `PromptBuilder` convenience wrappers to control media output

felixarntz:
The `PromptBuilder` was missing methods to control 3 `ModelConfig` properties:
- `outputMediaOrientation`
- `outputMediaAspectRatio`
- `outputSpeechVoice`

Additionally, this PR fixes an odd but somewhat important bug with the `usingStopSequences` method not working at all (it needs to set the top-level key, not a key in `customOptions`).

### DIFF
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -11,6 +11,7 @@ use WordPress\AiClient\Events\AfterGenerateResultEvent;
 use WordPress\AiClient\Events\BeforeGenerateResultEvent;
 use WordPress\AiClient\Files\DTO\File;
 use WordPress\AiClient\Files\Enums\FileTypeEnum;
+use WordPress\AiClient\Files\Enums\MediaOrientationEnum;
 use WordPress\AiClient\Messages\DTO\Message;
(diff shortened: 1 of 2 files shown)


felixarntz:
@JasonTheAdams One other thing to consider here is that we also don't support setting custom options via `PromptBuilder` convenience methods - they can only be set by using the slightly more complex `usingModelConfig` method.

Maybe that's still a reasonable path forward, because using those is more advanced and assumes extra responsibility on the caller - _they_ will need to know whether the API supports a specific custom option or not - but I wanted to bring it up so we can consider whether or not to add a convenience method for them.

If we want to include that in the `PromptBuilder`, it could be a simple `PromptBuilder::usingCustomOption(string $option, $value)`.

gziolo (approved):
Code review looks good. All checks pass locally (tests, PHPCS clean, PHPStan clean). The stop sequences bug fix is important and the new convenience methods follow established patterns consistently. The changes are straighforwad and covered properly with tests.

Re: custom options convenience method — agree it's worth considering as a follow-up, but shouldn't block this PR.

PR #219, opened 2026-03-09
Add support for setting logo path in `ProviderMetadata`

felixarntz:
See https://github.com/WordPress/gutenberg/pull/76190 (which will make it into WordPress 7.0): We need to support setting an optional logo path per provider.


gziolo (approved):
Perfect, thank you!

PR #220, opened 2026-03-18
Fix incorrect JSON schema for `FunctionCall` and `FunctionResponse`

felixarntz:
* `FunctionCall` and `FunctionResponse` are currently using `oneOf`, which is not correct. It is totally valid for a function call or a function response to have _both_ `id` and `name` set. So this needs to be `anyOf` instead.
    * It doesn't show in any PHP-only usage, but as soon as you do something where the schema matters (e.g. REST API), this schema will lead to errors.
    * Regardless of whether or not it errors though, it is simply semantically incorrect.
* Somewhat related, the `FunctionResponse` was not handling its `id` and `name` the same way as `FunctionCall`. The properties weren't nullable, even though they should be - only one is required, though both are allowed. This was already reflected in parts of the class, but not consistently.

This PR fixes both of these bugs.

### DIFF
--- a/src/Tools/DTO/FunctionResponse.php
+++ b/src/Tools/DTO/FunctionResponse.php
@@ -163,14 +163,9 @@ class FunctionResponse extends AbstractDataTransferObject
     {
         static::validateFromArrayData($array, [self::KEY_RESPONSE]);
 
-        // Validate that at least one of id or name is provided
-        if (!array_key_exists(self::KEY_ID, $array) && !array_key_exists(self::KEY_NAME, $array)) {
-            throw new InvalidArgumentException('At least one of id or name must be provided.');
-        }
-
         return new self(
-            array_key_exists(self::KEY_ID, $array) ? (string) $array[self::KEY_ID] : null,
-            array_key_exists(self::KEY_NAME, $array) ? (string) $array[self::KEY_NAME] : null,
+            $array[self::KEY_ID] ?? null,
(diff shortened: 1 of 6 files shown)


src/Tools/DTO/FunctionResponse.php
```diff
@@ -156,8 +169,8 @@ public static function fromArray(array $array): self
         }
 
         return new self(
```
gziolo, line 166: Non-blocking nitpick (caught by agent): this `fromArray()` now uses `array_key_exists` with a defensive `(string)` cast, while `FunctionCall::fromArray()` still uses `$array[self::KEY_ID] ?? null`. The difference is subtle — `array_key_exists` distinguishes between a missing key and a key explicitly set to `null`. Neither causes a real-world issue, but it's a minor inconsistency between two sibling classes doing the same thing.

gziolo (approved):
Looks good to me\! The `oneOf` → `anyOf` fix is semantically correct and the `FunctionResponse` alignment with `FunctionCall` resolves a real inconsistency. Well tested.

src/Tools/DTO/FunctionResponse.php
felixarntz: Great point, updated in https://github.com/WordPress/php-ai-client/pull/220/changes/22d2d166786e1ddf901d43273c8117518da08a3f.

PR #221, opened 2026-03-18
Validate provider IDs

felixarntz:
Only allow lower-case alphanumeric characters, hyphens, and underscores.

### DIFF
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -104,6 +104,7 @@ class ProviderMetadata extends AbstractDataTransferObject
         if (!preg_match('/^[a-z0-9\-_]+$/', $id)) {
             throw new InvalidArgumentException(
                 sprintf(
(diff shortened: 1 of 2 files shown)


gziolo (approved):
Thank you for a quick follow up 👍🏻

src/Providers/DTO/ProviderMetadata.php
```diff
@@ -99,6 +101,15 @@ public function __construct(
         ?string $description = null,
         ?string $logoPath = null
     ) {
+        if (!preg_match('/^[a-z0-9\-_]+$/', $id)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Invalid provider ID "%s". Only lowercase alphanumeric characters, hyphens, and underscores are allowed.',
```
felixarntz, line 108:
```suggestion
                    // phpcs:ignore Generic.Files.LineLength.TooLong
                    'Invalid provider ID "%s". Only lowercase alphanumeric characters, hyphens, and underscores are allowed.',
```

PR #222, opened 2026-03-20
Add sound generation support

saarnilauri:
## Summary                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  - Introduce SoundGenerationModelInterface and SoundGenerationOperationModelInterface contracts for sound generation models                                                                                                       
  - Add SOUND_GENERATION capability to CapabilityEnum
  - Wire sound generation into PromptBuilder and AiClient with full test coverage                                                                                                                                                  
 
 ## Why this is needed?

Why a separate SOUND_GENERATION capability instead of reusing SPEECH_GENERATION?                                                                                                                                                 
                                                                                                                                                                                                                                   
  Speech and sound are fundamentally different audio domains with different provider landscapes, parameters, and use cases:                                                                                                        
                                                                                                                                                                                                                                   
  1. Different output types — Speech models produce spoken language (voice, narration, dialogue). Sound models produce non-verbal audio (sound effects, ambient noise, foley). A consumer asking for "rain on a tin roof" or "sword
   clash" should not be routed to a speech model.                                                                                                                                                                                  
  2. Different provider models — AI providers expose these as distinct model types (e.g. ElevenLabs sound effects vs. their TTS voices, or dedicated SFX models). Collapsing them into one capability would force providers to     
  either advertise a capability they don't fully support or implement awkward routing logic internally.                                                                                                                            
  3. Consistent with the existing pattern — The client already distinguishes between TEXT_TO_SPEECH_CONVERSION (deterministic TTS), SPEECH_GENERATION (generative voice/dialogue), and MUSIC_GENERATION. Sound generation is a
  natural sibling — it completes the audio capability taxonomy without overlap. Merging sound into speech would break the 1:1 mapping between capability enum values and provider model interfaces that the rest of the            
  architecture relies on.
  4. Clean capability discovery — Consumers can query CapabilityEnum::soundGeneration() to find providers that actually support sound effect generation, rather than getting speech models that can't fulfill the request.
                                                                                                                                                                                                                                   
  ## Test plan
                                                                                                                                                                                                                                   
  - Unit tests added for PromptBuilder sound generation methods                                                                                                                                                                    
  - AiClient test updated to cover the new capability
  - CapabilityEnum test updated for SOUND_GENERATION

### DIFF
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -363,6 +363,30 @@ class AiClient
     }
 
     /**
+     * Generates sound using the traditional API approach.
+     *
+     * @since 1.4.0
+     *
+     * @param Prompt $prompt The prompt content.
+     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
+     *                                                        or model configuration for auto-discovery,
+     *                                                        or null for defaults.
+     * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
+     * @return GenerativeAiResult The generation result.
+     *
+     * @throws \InvalidArgumentException If the prompt format is invalid.
+     * @throws \RuntimeException If no suitable model is found.
+     */
+    public static function generateSoundResult(
+        $prompt,
+        $modelOrConfig = null,
+        ?ProviderRegistry $registry = null
+    ): GenerativeAiResult {
+        self::validateModelOrConfigParameter($modelOrConfig);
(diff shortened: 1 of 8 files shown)


saarnilauri:
@JasonTheAdams @felixarntz @Jameswlepage 

I made this PR some time ago. Would like to hear your thoughts.

PR #224, opened 2026-03-24
Fix: register providers and load .env in cli.php

### LINKED ISSUE #223: cli.php: missing provider registration and .env loading causes "Provider not registered" error
### Description:
Bug Report

### Summary

On running `cli.php` with any provider (e.g. `--providerId=google`) always fails with:

> [ERROR] Invalid arguments while trying to set up prompt builder: Provider not registered: google

### Cause

This happens even when the correct API key is set in `.env` and the provider packages (`wordpress/google-ai-provider`, `wordpress/openai-ai-provider`, `wordpress/anthropic-ai-provider`) are installed as dev dependencies.

`cli.php` is missing two bootstrapping steps that the integration test suite handles in [`tests/integration/bootstrap.php`](https://github.com/WordPress/php-ai-client/blob/trunk/tests/integration/bootstrap.php):

1. Loading `.env` API keys in `.env` are never read into the environment, so `getenv()` calls inside providers return nothing.
2. Registering providers — Provider classes (e.g. `GoogleProvider`) are never registered with the `ProviderRegistry`, so the registry has no providers to resolve by ID.

sarthak-19:
### What

`cli.php` was missing two bootstrapping steps, causing every run to fail with:

```
[ERROR] Invalid arguments while trying to set up prompt builder: Provider not registered: google
```

### Why

The `ProviderRegistry` has no built-in knowledge of provider packages they must be explicitly registered. Additionally, API keys stored in `.env` were never loaded into the environment, so providers could not authenticate even if registered.

### Changes

Added to `cli.php` (mirroring `tests/integration/bootstrap.php`):

1. **`.env` loading** via `symfony/dotenv` (already a dev dependency).
2. **Provider registration** for `AnthropicProvider`, `GoogleProvider`, and `OpenAiProvider`, guarded with `class_exists()` so the script works even if only a subset of provider packages are installed.

### Testing

```bash
# With a key in .env, e.g. GOOGLE_API_KEY=...
php cli.php 'Write a 2-verse poem about PHP.' --providerId=google --modelId=gemini-2.5-flash
```

### References

Fixes  : https://github.com/WordPress/php-ai-client/issues/223

### DIFF
--- a/cli.php
+++ b/cli.php
@@ -23,6 +23,31 @@ use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
 
 require_once __DIR__ . '/vendor/autoload.php';
 
+use Symfony\Component\Dotenv\Dotenv;
+use WordPress\AnthropicAiProvider\Provider\AnthropicProvider;
+use WordPress\GoogleAiProvider\Provider\GoogleProvider;
+use WordPress\OpenAiAiProvider\Provider\OpenAiProvider;
+
+// Load .env file if it exists (same approach as integration test bootstrap).
+$envFile = __DIR__ . '/.env';
+if (file_exists($envFile)) {
+    $dotenv = new Dotenv();
+    $dotenv->usePutenv(true);
+    $dotenv->load($envFile);
+}
+
+// Register provider packages so the registry can discover them.
+$registry = AiClient::defaultRegistry();
+if (class_exists(AnthropicProvider::class)) {
+    $registry->registerProvider(AnthropicProvider::class);
+}
+if (class_exists(GoogleProvider::class)) {
+    $registry->registerProvider(GoogleProvider::class);
+}
+if (class_exists(OpenAiProvider::class)) {
+    $registry->registerProvider(OpenAiProvider::class);
+}
+
 /**
  * Prints the output to stdout.
  *
--- a/cli.php
+++ b/cli.php
@@ -1,14 +1,20 @@
 <?php
+
 /**
  * CLI script for interacting with the AI client.
  *
  * This script allows users to send prompts to the AI and receive responses.
  * It supports named arguments for provider and model selection.
  *
+ * Setup:
+ *   1. Copy cli-providers.php.example to cli-providers.php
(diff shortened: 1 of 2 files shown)


cli.php
```diff
@@ -23,6 +23,31 @@
...
+$registry = AiClient::defaultRegistry();
+if (class_exists(AnthropicProvider::class)) {
+    $registry->registerProvider(AnthropicProvider::class);
+}
+if (class_exists(GoogleProvider::class)) {
+    $registry->registerProvider(GoogleProvider::class);
+}
+if (class_exists(OpenAiProvider::class)) {
+    $registry->registerProvider(OpenAiProvider::class);
+}
```
felixarntz, line 50: It's not ideal to hard-code specific providers here. What if you want to test the CLI with other ones?

I think this should be removed. It would be better to instead support a simple system where another user-controlled file handles the registration of providers. Could look like this:

1. check if the file (e.g. `__DIR__ . '/cli-providers.php`) exists
2. if so, implement a closure here in which that file is loaded, with _only_ access to the `$registry` instance so that it cannot do anything else

The `cli-providers.php` file would be git-ignored. It would need to look something like this (with `$registry` being in the local scope of where it is required):
```php
use WordPress\AnthropicAiProvider\Provider\AnthropicProvider;
use WordPress\GoogleAiProvider\Provider\GoogleProvider;
use WordPress\OpenAiAiProvider\Provider\OpenAiProvider;

$registry->registerProvider(AnthropicProvider::class);
$registry->registerProvider(GoogleProvider::class);
$registry->registerProvider(OpenAiProvider::class);
```

felixarntz (changes requested):
@sarthak-19 Thank you for the PR!

The `.env` integration looks good, but the provider registration needs a cleaner approach.

cli.php
sarthak-19: Hi @felixarntz, thanks for the feedback, yup make sense, I've added few changes, also I believe we should add a template.example file so it'll be easier for new contributors to start as well.

Issue #215, opened 2026-02-27
Add utility to create lists of modality combinations


felixarntz:
See https://github.com/WordPress/ai-provider-for-google/pull/11#issuecomment-3971975538 (cc @gziolo) - this is going to be helpful in reducing tedious code bits.

It's crucial to support `$required` and `$optional` modalities in the utility. Modality lists are considered unordered (as in "order doesn't matter"), so we only need to make sure to cover the combinations like that.

sarthak-19:
Since PR is merged we can close the issue.

cc : @felixarntz

PR #225, opened 2026-03-24
Add `ModalityCombinationsUtil` for building modality combination sets

sarthak-19:
### Summary

Adds a new static utility class `ModalityCombinationsUtil` under `src/Messages/Util/` that generates all valid modality combinations from a set of required modalities and a set of optional modalities.

This reduces repetitive, error-prone manual array declarations that provider implementations currently need when registering which input/output modality combinations their models support.

### How it works

Uses binary bitmask enumeration to produce all `2^n` subsets of the optional modalities, each merged with the fixed required modalities. Because modality lists are unordered, subsets (not permutations) are the correct unit — each combination appears exactly once.

### Usage

```php
// All combinations that include text, plus any subset of image/audio/document
$combos = ModalityCombinationsUtil::buildCombinations(
    [ModalityEnum::text()],
    [ModalityEnum::image(), ModalityEnum::audio(), ModalityEnum::document()]
);
// → 8 combinations

// All subsets of text + audio (including the "neither" case)
$combos = ModalityCombinationsUtil::buildCombinations(
    [],
    [ModalityEnum::text(), ModalityEnum::audio()]
);
// → 4 combinations
```

### Issue : 

- https://github.com/WordPress/php-ai-client/issues/215

### DIFF
--- a/src/Messages/Util/ModalityCombinationsUtil.php
+++ b/src/Messages/Util/ModalityCombinationsUtil.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Messages\Util;
+
+use WordPress\AiClient\Messages\Enums\ModalityEnum;
+
+/**
+ * Utility class for building modality combinations.
+ *
+ * @since n.e.x.t
+ */
+class ModalityCombinationsUtil
+{
+    /**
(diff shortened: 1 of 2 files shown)


felixarntz (approved):
@sarthak-19 This looks great - thank you for the PR!

PR #227, opened 2026-04-12
Expose model capabilities

### LINKED ISSUE #226: Expose model capabilities (input/output types)
Currently, there is no way to check what input/output types a model supports (e.g. text, audio, image).

---

### Problem

When building features like model selection UIs or validation, developers need to know if a model supports:

* text → text
* audio → text
* text → audio

Right now, this requires guessing based on model names (e.g. `transcribe`, `audio`, etc.), which is not reliable.

---

### Expected Solution

Provide a way to access model capabilities, for example:

```php
$model = AiClient::model('gpt-4o');

$model->getCapabilities();

/*
[
  'input' => ['text', 'image'],
  'output' => ['text'],
]
*/
```

Or simple helpers:

```php
$model->supportsInput('text');
$model->supportsOutput('text');
```

---

### Why this helps

* Makes model selection safer
* Removes fragile string-based checks
* Helps build better UIs and integrations

---

sarthak-19:
### Summary

Closes https://github.com/WordPress/php-ai-client/issues/226 **Expose model input/output modality capabilities at runtime.**

Currently there is no ergonomic way to ask a model instance which input and output modalities it accepts. Developers had to guess from model names (`"transcribe"`, `"audio"`, etc.) or dig through raw `SupportedOption` objects — both fragile approaches.

This PR adds three first-class methods to `ModelInterface` and implements them in `AbstractApiBasedModel`:

```php
// Returns all supported input/output modality enums
$model->getCapabilities();
/*
[
  'input'  => [ModalityEnum::text(), ModalityEnum::image()],
  'output' => [ModalityEnum::text()],
]
*/

// Simple boolean helpers
$model->supportsInput(ModalityEnum::text());   // true
$model->supportsInput(ModalityEnum::audio());  // false
$model->supportsOutput(ModalityEnum::text());  // true
$model->supportsOutput(ModalityEnum::image()); // false
```

### DIFF
--- a/src/Providers/ApiBasedImplementation/AbstractApiBasedModel.php
+++ b/src/Providers/ApiBasedImplementation/AbstractApiBasedModel.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\Providers\ApiBasedImplementation;
 
+use WordPress\AiClient\Messages\Enums\ModalityEnum;
 use WordPress\AiClient\Providers\ApiBasedImplementation\Contracts\ApiBasedModelInterface;
 use WordPress\AiClient\Providers\DTO\ProviderMetadata;
 use WordPress\AiClient\Providers\Http\Contracts\WithHttpTransporterInterface;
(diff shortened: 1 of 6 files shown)


felixarntz (changes requested):
@sarthak-19 See my comment on the issue https://github.com/WordPress/php-ai-client/issues/226#issuecomment-4655343702, I'm not sure this is a good idea.

It creates an API surface for something that is already possible, adding arbitrary methods to a central infrastructure class.

If we want to make this more convenient, the code should live in a separate helper class that takes a model instance as input, not on the `ModelInterface` itself.

PR #228, opened 2026-04-16
Change docblock to reflect media orientation

westonruter:
I noticed this copy pasta from `FileTypeEnum` as I was browsing the source.

Introduced in a6a83cb391af300328546a384ce865aa3fa740c0.


felixarntz (approved):
@westonruter Thank you!

PR #231, opened 2026-05-02
Avoid listing models for explicit text model IDs

### LINKED ISSUE #230: Explicit model instantiation should not require live model-list metadata fetch
## Summary

When a caller explicitly configures a provider and model ID, `ProviderRegistry::getProviderModel()` still requires a live provider model-list metadata fetch before it can instantiate the model. For OpenAI this means a configured model such as `gpt-5.4` triggers `GET https://api.openai.com/v1/models` before the actual generation request.

This makes model metadata discovery a hard runtime dependency for explicit model usage. If the model-list request times out or is unavailable, generation fails before the configured model can be used, even though the provider's generation endpoint would validate the model ID anyway.

## Observed failure

In a local WordPress Studio site running wp-ai-client through Data Machine pipeline jobs, child jobs failed with:

```text
wp-ai-client request failed: Network error occurred while sending GET request to https://api.openai.com/v1/models: cURL error 28: Connection timed out after 30000 milliseconds
```

The failing generation request was not large: transcript evidence showed provider `openai`, model `gpt-5.4`, and total request size around 31 KB. Later direct checks from shell `curl` and WordPress PHP `wp_remote_get()` to `/v1/models` returned quickly with `401`, so payload size was not the cause.

## Call chain

The configured model ID still triggers `GET /v1/models` through this path:

1. Caller asks the registry for an explicit provider/model pair:
   `ProviderRegistry::getProviderModel( $provider, $modelId, $modelConfig )`

2. `ProviderRegistry::getProviderModel()` resolves the provider class and calls the provider static model factory:
   `src/Providers/ProviderRegistry.php`, `getProviderModel()`, lines 255-260 in the local copy.

3. `AbstractProvider::model()` always resolves remote metadata for the requested model ID before it creates a model instance:
   `src/Providers/AbstractProvider.php`, `model()`, lines 50-58 in the local copy.

   The important line is:

   ```php
   $modelMetadata = static::modelMetadataDirectory()->getModelMetadata($modelId);
   ```

4. `AbstractApiBasedModelMetadataDirectory::getModelMetadata()` loads the full model metadata map:
   `src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php`, lines 59-65.

5. The map is populated via `sendListModelsRequest()` on cache miss:
   `src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php`, lines 74-78.

   ```php
   return $this->cached(self::MODELS_CACHE_KEY, fn() => $this->sendListModelsRequest(), 86400);
   ```

6. OpenAI-compatible metadata directories implement this as a network request to `models`:
   `src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php`, lines 29-35.

   ```php
   $request = $this->createRequest(HttpMethodEnum::GET(), 'models');
   $request = $this->getRequestAuthentication()->authenticateRequest($request);
   $response = $httpTransporter->send($request);
   ```

7. The OpenAI provider resolves that path under `https://api.openai.com/v1`.

## Expected behavior

If the caller explicitly supplies a provider and model ID, wp-ai-client should be able to instantiate a model object without a live model-list API call. Validation can happen at generation time, or the provider can synthesize conservative metadata for known model-ID families.

## Why this matters

- Explicit model selection should not require remote discovery.
- `/models` metadata fetching is redundant for explicit generation requests.
- The current behavior adds latency and a failure point before the actual generation API call.
- The failure is especially painful for short-lived job runners where per-request in-memory metadata cache is cold.

## Possible fix direction

Add an explicit-model path that does not require `modelMetadataDirectory()->getModelMetadata( $modelId )` to hit a live list-models endpoint.

Possible shapes:

- Let providers create model metadata from an explicit model ID without listing all models.
- Add a `createModelFromId()` / `getFallbackModelMetadata()`-style provider hook.
- For OpenAI-compatible providers, derive conservative capabilities/options from known model ID prefixes and let the generation endpoint be authoritative.

The key contract change: `getProviderModel( $provider, $modelId )` should not require network I/O merely to instantiate the explicitly requested model.

## Related mitigation

Persistent cache via `AiClient::setCache()` can reduce repeated `/models` calls, but it does not fix this contract. It still makes explicit model usage depend on successful model-list metadata discovery at least once per cache TTL.

chubes4:
## Summary

- Adds a provider-overridable path for explicit model IDs to provide metadata without listing all provider models first.
- Applies explicit metadata consistently across `getModelMetadata()`, `hasModelMetadata()`, and listed model metadata for matching listed IDs.
- Keeps explicit metadata request-local and memoized per directory instance instead of storing synthetic metadata in the shared models cache.

Fixes #230.

## Why

Explicit provider/model selection currently calls the provider model-list endpoint before the model can be instantiated. For providers that accept explicit/current model IDs, that means `getProviderModel( 'openai', 'gpt-5.4' )` can fail on a transient `GET /models` issue before the actual generation endpoint has a chance to validate the explicitly requested model.

## Testing

- `composer test -- --filter AbstractApiBasedModelMetadataDirectoryTest`
- `composer lint`

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.6 Sol)
- **Used for:** Drafting the patch and tests, including the batched explicit-metadata follow-up; Chris reviewed the behavior on a local Studio site and remains responsible for the change.


src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleModelMetadataDirectory.php
```diff
@@ -23,6 +24,41 @@
...
+        if (str_starts_with($modelId, 'gpt-image-') || str_starts_with($modelId, 'dall-e-')) {
+            return false;
+        }
+
+        if (str_starts_with($modelId, 'gpt-') || str_starts_with($modelId, 'chatgpt-')) {
+            return true;
+        }
+
+        return preg_match('/^o\d(?:-|$)/', $modelId) === 1;
+    }
```
JasonTheAdams: I don't think this should live in this codebase, rather the OpenAI Provider repository. While we do have "OpenAI Compatible" classes, that's about the shape of the API, not coupling to any specific models.

Curious if you agree, @felixarntz.

felixarntz: Yes, model IDs should only be in the provider specific repos.

chubes4: I've made the requested changes and opened a follow-up draft in the OpenAI provider: https://github.com/WordPress/ai-provider-for-openai/pull/23

This PR now just provides the hook and abstract test coverage. Thank you!

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -68,6 +68,18 @@ final public function hasModelMetadata(string $modelId): bool
...
+            $modelsMetadata = $this->getModelMetadataMap();
+            if (isset($modelsMetadata[$modelId])) {
+                return $modelsMetadata[$modelId];
+            }
+        }
+
+        $explicitModelMetadata = $this->createModelMetadataForExplicitModelId($modelId);
+        if ($explicitModelMetadata !== null) {
+            return $explicitModelMetadata;
+        }
```
felixarntz: it's a bit of a strange design that this will only be used if no cached value has been found, especially since this injected model metadata does not get cached itself. what if the data from the API differs from what's locally defined in the code?

Either we need to cache the result of this too (which I'd have other concerns over though because of cache invalidation) or we need to always run this before the cache logic. If the implementer wants to add some kind of caching themselves for it, they still could.

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -68,6 +68,18 @@ final public function hasModelMetadata(string $modelId): bool
...
     final public function getModelMetadata(string $modelId): ModelMetadata
     {
+        if ($this->hasCache(self::MODELS_CACHE_KEY)) {
+            $modelsMetadata = $this->getModelMetadataMap();
+            if (isset($modelsMetadata[$modelId])) {
+                return $modelsMetadata[$modelId];
+            }
+        }
+
+        $explicitModelMetadata = $this->createModelMetadataForExplicitModelId($modelId);
```
felixarntz: This being only in `getModelMetadata()` but not in `listModelMetdata()` will lead to inconsistent data - so the approach doesn't work correctly in terms of the full API.

I understand this is meant to be used to bypass the need for a request in case an explicit model ID is passed, but that's not controlled by this class. Any code could call `getModelMetadata()` here, so it'd be confusing if `getModelMetadata()` returned a different set of metadata than the same model would appear with as part of `listModelMetadata()`.

I think there's a way to implement this in a way that the code here still bypasses the request while ensuring consistent data. E.g. for `listModelMetadata()` we could make the request first (necessary there!), and then for every single model ID call this method. If the method returns something, override what the API returned - this way it's the same result as here.

There's performance and data integrity to consider on that path - we'd probably want to build in memoization and potentially even caching of this so that:
- this isn't called for every model every time that `listModelMetadata()` is called
- this doesn't create multiple instances for the same model ID within the same request

It's not super trivial, but I think it's possible.

felixarntz (changes requested):
@chubes4 While this addresses the reported concern, it creates data inconsistencies when considering the API surface of this class holistically.

I think there's a path forward, but it requires some modifications in the approach.

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
chubes4: > what if the data from the API differs from what's locally defined in the code?

I think we should run it before cache logic. I'm trying to keep this as simple as possible.

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
chubes4: Thanks for the review and thoughtful comments! 

I think this is mainly a problem on slow connections or local environments where the CPU/RAM is constrained. I have not seen it on my live sites but have seen it repeatedly in a local environment inside Studio/WordPress Playground. 

I will revisit this based on your feedback and update the PR to consider the broader implications.

chubes4:
Updated this to address the consistency concern:

- `getModelMetadata()` and `hasModelMetadata()` now check explicit metadata before the list/cache path, so provider-local metadata is authoritative when available.
- `listModelMetadata()` still performs the required list request, then overlays explicit metadata for matching listed model IDs so listed data and direct lookup agree for the same ID.
- Explicit metadata is memoized only for the directory instance, avoiding persistence/cache invalidation concerns for synthetic metadata.

Verified with:

- `composer test -- --filter AbstractApiBasedModelMetadataDirectoryTest`
- `composer lint`

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
```diff
@@ -77,6 +95,43 @@ final public function getModelMetadata(string $modelId): ModelMetadata
...
+     * @since n.e.x.t
+     *
+     * @param string $modelId The explicit model ID.
+     * @return ModelMetadata|null The model metadata, or null to fall back to listing provider models.
+     */
+    private function getExplicitModelMetadata(string $modelId): ?ModelMetadata
+    {
+        if (!array_key_exists($modelId, $this->explicitModelMetadataCache)) {
+            $this->explicitModelMetadataCache[$modelId] = $this->createModelMetadataForExplicitModelId($modelId);
+        }
```
felixarntz: instead of calling `createModelMetadataForExplicitModelId` for every single model ID (including all the ones that inevitably won't have explicit model metadata), why don't we instead make `$this->explicitModelMetadataCache` a null value initially that gets populated upon the first call for a model, but with all relevant models that have explicit metadata?

since explicit model metadata will have to be provided in "manual" code, I doubt you'd have more than a dozen or so entries there. But there may be 100s of model IDs in a provider, and many providers probably wouldn't even implement explicit model ID metadata at all. So I think it would be worth optimizing performance for the basic case that you get out of the box (no explicit model metadata).

felixarntz (approved):
@chubes4 I think this looks good, though one optimization recommendation remains.

src/Providers/ApiBasedImplementation/AbstractApiBasedModelMetadataDirectory.php
chubes4: Updated in 7c6b04b. The core hook now receives all unchecked candidate IDs in one call and returns a keyed override map. The directory memoizes both returned metadata and omitted IDs, so providers using the default implementation incur one O(1) callback for an entire listed set rather than one callback per model. The dependent OpenAI implementation remains dynamic by matching future model IDs within the supplied batch: https://github.com/WordPress/ai-provider-for-openai/pull/23

Verified with the focused 11-test directory suite, full lint, and an integrated OpenAI `gpt-5.4` lookup that completed without a models request.

AI assistance: OpenCode (GPT-5.6 Sol) implemented and verified the batch follow-up; Chris reviewed and remains responsible for the change.

chubes4:
Thank you, @felixarntz!

PR #232, opened 2026-05-06
Use `non-empty-string` instead of `string` for Provider properties 

westonruter:
The `non-empty-string` is a more specific type than `string`. It's clear that this is the intended type. For example, a `ProviderMetadata` constructor throws an exception if an empty string is provided: `!preg_match('/^[a-z0-9\-_]+$/', $id)`.

This addresses some issues uncovered in https://github.com/WordPress/wordpress-develop/pull/11701.

### DIFF
--- a/src/Providers/DTO/ProviderMetadata.php
+++ b/src/Providers/DTO/ProviderMetadata.php
@@ -20,13 +20,13 @@ use WordPress\AiClient\Providers\Http\Enums\RequestAuthenticationMethod;
  * @since 1.3.0 Added optional logoPath property.
  *
  * @phpstan-type ProviderMetadataArrayShape array{
- *     id: string,
- *     name: string,
- *     description?: ?string,
- *     type: string,
- *     credentialsUrl?: ?string,
- *     authenticationMethod?: ?string,
- *     logoPath?: ?string
+ *     id: non-empty-string,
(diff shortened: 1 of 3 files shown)


src/Providers/DTO/ProviderMetadata.php
```diff
@@ -20,13 +20,13 @@
...
- *     name: string,
- *     description?: ?string,
- *     type: string,
- *     credentialsUrl?: ?string,
- *     authenticationMethod?: ?string,
- *     logoPath?: ?string
+ *     id: non-empty-string,
+ *     name: non-empty-string,
+ *     description?: ?non-empty-string,
+ *     type: non-empty-string,
```
felixarntz, line 26: While you're at it - this is technically an enum value. Does PHPStan have something for this?

Not a must-have to address here, but wanted to flag in case it's trivial.

felixarntz (approved):
@westonruter Thank you, looks great! One follow up question.

src/Providers/DTO/ProviderMetadata.php
westonruter: Yes, we can use string literals like:

```suggestion
 *     type: 'type1'|'type2'|'type3',
```

Not sure what the values are specifically in the enum.

PR #234, opened 2026-05-09
Preserve empty JSON schema object maps

### LINKED ISSUE #233: Preserve empty JSON Schema object maps in FunctionDeclaration parameters
## Description

`FunctionDeclaration` accepts JSON Schema-like parameter definitions, but empty object-map fields inside those schemas currently serialize as JSON arrays. In PHP, callers naturally represent an empty object map as `[]`, but JSON Schema fields such as `properties` are object maps and should serialize as `{}` when empty.

This can make an otherwise valid function parameter schema invalid by the time a provider request body is JSON encoded.

## Minimal reproduction

```php
use WordPress\AiClient\Tools\DTO\FunctionDeclaration;

$declaration = new FunctionDeclaration(
    'inspectObject',
    'Inspects an object with optional nested metadata',
    [
        'type' => 'object',
        'properties' => [
            'metadata' => [
                'type' => 'object',
                'properties' => [],
            ],
        ],
    ]
);

echo json_encode($declaration, JSON_THROW_ON_ERROR);
```

## Actual behavior

The nested JSON Schema object map is emitted as a JSON array:

```json
{
  "name": "inspectObject",
  "description": "Inspects an object with optional nested metadata",
  "parameters": {
    "type": "object",
    "properties": {
      "metadata": {
        "type": "object",
        "properties": []
      }
    }
  }
}
```

## Expected behavior

JSON Schema object-map fields should preserve object semantics when empty:

```json
{
  "name": "inspectObject",
  "description": "Inspects an object with optional nested metadata",
  "parameters": {
    "type": "object",
    "properties": {
      "metadata": {
        "type": "object",
        "properties": {}
      }
    }
  }
}
```

## Evidence

I added a focused PHPUnit repro locally on `trunk`:

```php
public function testJsonSerializationPreservesEmptySchemaObjectMaps(): void
{
    $declaration = new FunctionDeclaration(
        'inspectObject',
        'Inspects an object with optional nested metadata',
        [
            'type' => 'object',
            'properties' => [
                'metadata' => [
                    'type' => 'object',
                    'properties' => [],
                ],
            ],
        ]
    );

    $json = json_encode($declaration, JSON_THROW_ON_ERROR);

    $this->assertStringContainsString('"properties":{', $json);
    $this->assertStringNotContainsString('"properties":[]', $json);
}
```

It fails on current `trunk`:

```text
Failed asserting that
'{"name":"inspectObject","description":"Inspects an object with optional nested metadata","parameters":{"type":"object","properties":{"metadata":{"type":"object","properties":[]}}}}'
does not contain ""properties":[]".

Tests: 1, Assertions: 2, Failures: 1.
```

The full `FunctionDeclarationTest` file also only fails this added assertion:

```text
Tests: 15, Assertions: 56, Failures: 1.
```

## Why this likely belongs in the client abstraction

This is provider-agnostic. The issue is not that one provider has a special rule; it is that `FunctionDeclaration` stores JSON Schema parameter definitions in PHP arrays, and PHP arrays do not preserve empty object-map semantics during JSON encoding.

The affected concept is shared by all providers that consume `FunctionDeclaration` parameters. Installed provider code paths also commonly embed raw `getParameters()` or `toArray()` output into request payloads, so individual providers can encounter the same empty-map serialization issue.

## Related

This is adjacent to #229 because both involve JSON Schema payload correctness, but it is not the same bug. #229 is about the OpenAI-compatible structured-output envelope. This issue is about preserving JSON Schema object-map semantics inside `FunctionDeclaration` parameters.

chubes4:
## Summary
- Adds a JSON-serializable parameter accessor for `FunctionDeclaration` that preserves empty JSON Schema object-map fields as `{}`.
- Updates `FunctionDeclaration::jsonSerialize()` to use the safe parameter representation.
- Updates the shared OpenAI-compatible tool payload builder to use the safe parameter representation when embedding function declarations.

Fixes https://github.com/WordPress/php-ai-client/issues/233.

## Testing
- `vendor/bin/phpunit tests/unit/Tools/DTO/FunctionDeclarationTest.php --filter testJsonSerializationPreservesEmptySchemaObjectMaps`
- `vendor/bin/phpunit tests/unit/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModelTest.php --filter testPrepareToolsParamPreservesEmptySchemaObjectMaps`
- `vendor/bin/phpunit tests/unit/Tools/DTO/FunctionDeclarationTest.php`
- `vendor/bin/phpunit tests/unit/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModelTest.php`
- `composer phpcs`
- `composer phpstan`

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Drafted the focused serialization fix and tests from the issue reproduction; Chris remains responsible for review and merge.

### DIFF
--- a/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
+++ b/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
@@ -492,9 +492,15 @@ abstract class AbstractOpenAiCompatibleTextGenerationModel extends AbstractApiBa
     {
         $tools = [];
         foreach ($functionDeclarations as $functionDeclaration) {
+            $function = $functionDeclaration->toArray();
+            $parameters = $functionDeclaration->getJsonSerializableParameters();
+            if ($parameters !== null) {
+                $function[FunctionDeclaration::KEY_PARAMETERS] = $parameters;
+            }
+
             $tools[] = [
                 'type' => 'function',
-                'function' => $functionDeclaration->toArray(),
+                'function' => $function,
             ];
         }
 
--- a/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
+++ b/src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
@@ -492,15 +492,9 @@ abstract class AbstractOpenAiCompatibleTextGenerationModel extends AbstractApiBa
     {
         $tools = [];
         foreach ($functionDeclarations as $functionDeclaration) {
-            $function = $functionDeclaration->toArray();
-            $parameters = $functionDeclaration->getJsonSerializableParameters();
-            if ($parameters !== null) {
-                $function[FunctionDeclaration::KEY_PARAMETERS] = $parameters;
-            }
-
             $tools[] = [
                 'type' => 'function',
-                'function' => $function,
+                'function' => $functionDeclaration->jsonSerialize(),
             ];
         }
 
--- a/src/Tools/DTO/FunctionDeclaration.php
+++ b/src/Tools/DTO/FunctionDeclaration.php
@@ -95,28 +95,6 @@ class FunctionDeclaration extends AbstractDataTransferObject
     }
 
     /**
-     * Gets the function parameters schema in a JSON-serializable form.
-     *
-     * JSON Schema object-map fields such as properties must encode as JSON
-     * objects even when empty. PHP arrays cannot preserve that distinction
-     * without casting the empty map before serialization.
-     *
-     * @since 0.1.0
-     *
-     * @return array<string, mixed>|\stdClass|null The JSON-serializable parameters schema.
-     */
-    public function getJsonSerializableParameters()
-    {
-        if ($this->parameters === null) {
-            return null;
-        }
-
-        /** @var array<string, mixed>|\stdClass $parameters */
-        $parameters = $this->prepareJsonSchemaObjectMaps($this->parameters, self::KEY_PARAMETERS);
-        return $parameters;
-    }
-
-    /**
      * {@inheritDoc}
      *
      * @since 0.1.0
@@ -176,7 +154,10 @@ class FunctionDeclaration extends AbstractDataTransferObject
         $data = $this->toArray();
 
         if ($this->parameters !== null) {
-            $data[self::KEY_PARAMETERS] = $this->getJsonSerializableParameters();
+            $data[self::KEY_PARAMETERS] = $this->prepareJsonSchemaObjectMaps(
+                $this->parameters,
+                self::KEY_PARAMETERS
+            );
         }
 
         return $data;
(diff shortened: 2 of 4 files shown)


src/Providers/OpenAiCompatibleImplementation/AbstractOpenAiCompatibleTextGenerationModel.php
```diff
@@ -492,9 +492,15 @@ protected function prepareToolsParam(array $functionDeclarations): array
...
+            $function = $functionDeclaration->toArray();
+            $parameters = $functionDeclaration->getJsonSerializableParameters();
+            if ($parameters !== null) {
+                $function[FunctionDeclaration::KEY_PARAMETERS] = $parameters;
+            }
+
             $tools[] = [
                 'type' => 'function',
-                'function' => $functionDeclaration->toArray(),
+                'function' => $function,
```
felixarntz: isn't all this code just the same as this simpler version?
```suggestion
            $tools[] = [
                'type' => 'function',
                'function' => $functionDeclaration->jsonSerialize(),
```

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -94,6 +94,28 @@ public function getParameters(): ?array
...
+     *
+     * JSON Schema object-map fields such as properties must encode as JSON
+     * objects even when empty. PHP arrays cannot preserve that distinction
+     * without casting the empty map before serialization.
+     *
+     * @since 0.1.0
+     *
+     * @return array<string, mixed>|\stdClass|null The JSON-serializable parameters schema.
+     */
+    public function getJsonSerializableParameters()
```
felixarntz: see above, this really shouldn't be a public method. it's oddly specific to be part of a public API, and doesn't seem needed as public anyway.

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -94,6 +94,28 @@ public function getParameters(): ?array
...
     }
 
+    /**
+     * Gets the function parameters schema in a JSON-serializable form.
+     *
+     * JSON Schema object-map fields such as properties must encode as JSON
+     * objects even when empty. PHP arrays cannot preserve that distinction
+     * without casting the empty map before serialization.
+     *
+     * @since 0.1.0
```
felixarntz: `@since` annotations for new code must always use `@since n.e.x.t`.

src/Tools/DTO/FunctionDeclaration.php
```diff
@@ -143,6 +165,69 @@ public function toArray(): array
         return $data;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @since 0.1.0
```
felixarntz, line 149: see above. also applies throughout the PR.

felixarntz (changes requested):
@chubes4 Thank you for the PR! Mostly looks good, a few minor things to address.

chubes4:
Thank you! Will fix everything noted.

chubes4:
Thanks for the review, @felixarntz. I addressed the feedback in db2e83e: `prepareToolsParam()` now uses `jsonSerialize()` directly, the public helper was removed, and the new `@since` annotations use `n.e.x.t`.

PR #132, opened 2025-11-26
Add embedding generation support

Jameswlepage:
Fixes #130

## Summary
- add embedding DTOs/results/operations and expose helper transformers
- extend PromptBuilder, AiClient, and the CLI with embedding-specific inputs, result helpers, and output formats
- introduce embedding contracts in the provider layer plus an OpenAI embedding model implementation alongside updated mocks/tests/documentation

## Testing
- composer phpstan
- composer phpunit
- php cli.php "["Embed this doc"]" --capability=embeddings --providerId=openai --modelId=text-embedding-3-large --outputFormat=embeddings-vectors

PR #244, opened 2026-06-03
Add embedding generation APIs

### LINKED ISSUE #242: Implement embedding generation contracts and client APIs
## Problem
The AI Client has an embedding generation capability enum, but it does not currently expose usable embedding generation model contracts, result DTOs, or builder methods. WordPress AI applications need provider-agnostic embeddings for semantic search, recommendations, clustering, and knowledge-base retrieval.

## Proposed shape
Add first-class embedding generation support to the AI Client.

Suggested additions:
- EmbeddingGenerationModelInterface
- embedding result DTO containing vectors, dimensions, model metadata, and token usage
- PromptBuilder methods for generateEmbedding() and generateEmbeddings()
- model requirements support for CapabilityEnum::embeddingGeneration()
- supported options such as dimensions or encoding format where providers support them

## Acceptance criteria
- Consumers can request embeddings through the same provider-agnostic client pattern used for text/image generation.
- Embedding-capable models can be discovered and selected by capability.
- Result objects expose vectors and usage metadata.
- Mock provider tests cover single and batch embedding generation.

chubes4:
## Summary
- Adds provider-agnostic embedding contracts, immutable result DTOs, model options, and embedding-specific lifecycle events.
- Introduces a dedicated `EmbeddingBuilder` and shares model/provider selection with `PromptBuilder` through a composed `ModelResolver`.
- Keeps embedding inputs independent of conversational messages with a direct `list<string>` provider contract.
- Supports single and batch embedding generation through fluent and traditional APIs, with one provider call per batch.

Fixes #242.

## Testing
1. Run `composer phpcs`.
2. Run `composer phpstan`.
3. Run `php -d error_reporting='E_ALL & ~E_DEPRECATED' vendor/bin/phpunit --testsuite unit`.
4. Configure `.env` with `OPENAI_API_KEY` and install an OpenAI provider branch implementing the embedding contract.
5. Run `composer test:integration -- --filter EmbeddingGenerationIntegrationTest`.
6. Confirm single input returns one embedding, batch input preserves result order and cardinality, and requested dimensions are honored.

Local verification against current `trunk`: 1,132 tests and 4,184 assertions passed. A live OpenAI cross-package check also verified one-input and two-input requests at 256 dimensions.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (OpenAI GPT-5.6 Sol and GPT-5.6 Terra)
- **Used for:** Implemented and reviewed the builder/model-resolution architecture, provider contract, tests, documentation, and integration verification; Chris directed the architecture and remains responsible for the change.


dkotter:
I know we already have #132 open to add embedding support and while that PR is fairly old now, it has already gone through an initial review. I've not compared the code there to here but curious if the approach taken here is different than that and if so, what the benefits are here as compared to that? Ideally we proceed with a single approach to solve the same problem to not overtax contributors here.

chubes4:
Thanks for pointing that out. I honestly missed #132 before opening this, and there's definitely overlap. My apologies for the duplicated effort.

The main distinction I see is that this PR is provider-agnostic, while #132 includes OpenAI-specific implementation that I think belongs in the OpenAI provider package, not `php-ai-client`.

Happy to close this if maintainers want #132 to remain canonical, or take ownership here if this narrower direction makes more sense.

dkotter:
> Happy to close this if maintainers want #132 to remain canonical, or take ownership here if this narrower direction makes more sense.

I'd defer to @JasonTheAdams and/or @felixarntz on that, I think ideal to have a single PR we're working from in order to not duplicate effort but I also know that original PR has sat for months and likely needs work to get it across the finish line so may be worth closing that out and proceeding here instead.

JasonTheAdams (changes requested):
This is great! Thanks for the PR, @chubes4! One key suggestion:

Let's introduce an `Embedding` Value Object that's iterable and immutable, takes a `list<int|float>` of values and an integer for dimensions. This will contain all the validations (cleaning up `EmbeddingResult` and make it a first-class type.

chubes4:
@JasonTheAdams Pushed some updates with the requested changes. Let me know if you'd like to see anything else, and thanks again!

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        $normalizedValues = [];
+        foreach ($values as $value) {
+            if (!is_int($value) && !is_float($value)) {
+                throw new InvalidArgumentException('Embedding values must be integers or floats.');
+            }
+
+            $normalizedValues[] = $value;
+        }
```
JasonTheAdams: We're not really normalizing the values, we're just validating them. I don't think the extra array is necessary.

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Gets the number of vector values.
+     *
+     * @since n.e.x.t
+     *
+     * @return int The number of vector values.
+     */
+    public function count(): int
+    {
+        return count($this->values);
```
JasonTheAdams: Given that this can be quite large and is immutable, let's internally cache the result of this so subsequent `count()` calls use the cached value.

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,168 @@
...
+use Traversable;
+use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+
+/**
+ * Represents a single generated embedding vector.
+ *
+ * @since n.e.x.t
+ *
+ * @implements IteratorAggregate<int, float|int>
+ */
```
JasonTheAdams, line 22: Add a `@phpstan-type EmbeddingShape` here, then use that throughout. In other files use `@phpstan-import-type` to pull this in. This will make a proper type for `list<int|float>`

src/Builders/PromptBuilder.php
```diff
@@ -1166,6 +1226,62 @@ public function generateImage(): File
...
+        if (!$model instanceof EmbeddingGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support embedding generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        return $model->generateEmbeddingResult($promptMessages)->getEmbeddings();
```
dkotter: I see when we generate a single embedding result in `generateEmbeddingResult`, we call `$this->dispatchEvent` prior to calling `$model->generateEmbeddingResult`. But we don't call that here. I think that means anyone that runs batch generations won't have that event fire if I'm tracing things correctly here

src/Builders/PromptBuilder.php
```diff
@@ -1120,6 +1150,36 @@ public function generateVideoResult(): GenerativeAiResult
...
+        if (!$model instanceof EmbeddingGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support embedding generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        $this->dispatchEvent(new BeforeGenerateResultEvent($this->messages, $model, $capability));
```
dkotter: We dispatch the before event but doesn't seem we ever dispatch the after event. Is that something that we can add? I think that will break things like logging (for instance, the AI plugin uses the before and after events for logging)

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,168 @@
...
+final class Embedding implements Countable, IteratorAggregate, JsonSerializable
+{
+    /**
+     * @var list<float|int>
+     */
+    private array $values;
+
+    private int $dimensions;
```
dkotter, line 33: Any reason to not have a docblock on this to match the above?

src/Results/DTO/EmbeddingResult.php
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * @var list<Embedding>
+     */
+    private array $embeddings;
+
+    private int $dimensions;
+    private TokenUsage $tokenUsage;
+    private ProviderMetadata $providerMetadata;
+    private ModelMetadata $modelMetadata;
```
dkotter, line 73: Same question on all these, any reason to not have docblocks?

src/Results/DTO/EmbeddingResult.php
```diff
@@ -0,0 +1,249 @@
...
+        $this->id = $id;
+        $this->embeddings = $normalizedEmbeddings;
+        $this->dimensions = $dimensions;
+        $this->tokenUsage = $tokenUsage;
+        $this->providerMetadata = $providerMetadata;
+        $this->modelMetadata = $modelMetadata;
+        $this->additionalData = $additionalData;
+    }
+
+    public function getId(): string
```
dkotter, line 135:
```suggestion
    /**
     * {@inheritDoc}
     *
     * @since n.e.x.t
     */
    public function getId(): string
```

src/Results/DTO/EmbeddingResult.php
```diff
@@ -0,0 +1,249 @@
...
+     * @since n.e.x.t
+     *
+     * @return Embedding The first embedding.
+     */
+    public function getEmbedding(): Embedding
+    {
+        return $this->embeddings[0];
+    }
+
+    public function getDimensions(): int
```
dkotter, line 171: All of these methods should have docblocks

dkotter (changes requested):
Left a few minor comments but overall this is looking good to me.

A couple other things that ideally we see added here:

1. We have a `cli.php` script that is useful for quickly testing things. That currently doesn't support embeddings so would be nice to update that to make testing this PR easier
2. We should ensure our documentation is updated with this new functionality, in particular the `README.md`

chubes4:
Shipped an update to address the latest review! Thank you

src/Events/AfterGenerateResultEvent.php
```diff
@@ -47,13 +47,13 @@ class AfterGenerateResultEvent
...
      * @param CapabilityEnum|null $capability The capability that was used for generation.
-     * @param GenerativeAiResult $result The result from the model.
+     * @param ResultInterface $result The result from the model.
      */
     public function __construct(
         array $messages,
         ModelInterface $model,
         ?CapabilityEnum $capability,
-        GenerativeAiResult $result
+        ResultInterface $result
```
dkotter: While I understand why we're switching from `GenerativeAiResult` to the more generic `ResultInterface` in this file, this does open up backward incompatibilities. I'd be inclined to be okay with that if this was just a stand-alone package but because the goal is to merge this back into WordPress core, not sure we can proceed with this approach.

As an example, if someone is using this after event hook and they run some code like this:

```php
$event->getResult()->getCandidates()
```

This will throw a fatal error if the result is an `EmbeddingResult`. Consumers would instead need to run a check like this first:

```php
$result = $event->getResult();
if ($result instanceof GenerativeAiResult) {
    // generative handling
} elseif ($result instanceof EmbeddingResult) {
    // embedding handling
}
```

I think ideally we need to figure out a way to handle the difference between a `GenerativeAiResult` and `EmbeddingResult` ourselves (if possible) to avoid that breakage.

Could potentially introduce new `getGenerativeResult` and `getEmbeddingResult` methods and keep `getResult` as is. Or maybe there's other, better options here

src/Builders/PromptBuilder.php
```diff
@@ -479,6 +482,34 @@ public function usingCandidateCount(int $candidateCount): self
...
 
+    /**
+     * Sets the embedding dimensions.
+     *
+     * @since n.e.x.t
+     *
+     * @param int $dimensions The embedding dimensions.
+     * @return self
+     */
+    public function usingDimensions(int $dimensions): self
```
dkotter: This isn't for this PR but noting any of these new `using` or `generate` methods will need to be handled upstream in WordPress Core as part of the WP AI Client, so just noting that here so we don't forget

src/Builders/PromptBuilder.php
```diff
@@ -1166,6 +1230,70 @@ public function generateImage(): File
...
+
+        $eventMessages = array_merge(...$promptMessages);
+
+        $this->dispatchEvent(new BeforeGenerateResultEvent($eventMessages, $model, $capability));
+
+        $result = $model->generateEmbeddingResult($promptMessages);
+
+        $this->dispatchEvent(new AfterGenerateResultEvent($eventMessages, $model, $capability, $result));
+
+        return $result->getEmbeddings();
```
dkotter: Just a thought but there's a lot of overlap here between this code and what `generateEmbeddingResult` does. Is there a possibility to refactor things a bit so we can directly call `generateEmbeddingResult` here? Maybe loop through each prompt message and call `generateEmbeddingResult` on each individual message?

src/Builders/PromptBuilder.php
```diff
@@ -1166,6 +1230,70 @@ public function generateImage(): File
...
+            throw new InvalidArgumentException('Cannot generate embeddings from an empty prompt list.');
+        }
+
+        $promptMessages = [];
+        foreach ($prompts as $prompt) {
+            $promptMessages[] = $this->parsePromptToMessages($prompt);
+        }
+
+        $capability = CapabilityEnum::embeddingGeneration();
+        $model = $this->getConfiguredModel($capability);
```
dkotter: This is likely solved if we do some refactoring I mention in the comment below but I think there's a bug here right now where we don't actually consider the input modality when we look for the right model here. `$this->messages` is `null` at this point which `getConfiguredModel` will look at to help determine the right model to use

src/Builders/PromptBuilder.php
```diff
@@ -1166,6 +1230,70 @@ public function generateImage(): File
...
+        if (!$model instanceof EmbeddingGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support embedding generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        $eventMessages = array_merge(...$promptMessages);
```
dkotter: This will also be solved if we do the refactoring below but by merging these all together, the events below fire once instead of once per embedding generation. If someone wanted to hook in and change things before each event, that wouldn't work

dkotter (dismissed):

src/Builders/PromptBuilder.php
```diff
@@ -479,6 +482,34 @@ public function usingCandidateCount(int $candidateCount): self
...
+
+    /**
+     * Sets the embedding encoding format.
+     *
+     * @since n.e.x.t
+     *
+     * @param string $encodingFormat The embedding encoding format.
+     * @return self
+     */
+    public function usingEncodingFormat(string $encodingFormat): self
```
dkotter: Can you explain what this param is for? I see value in setting the dimensions but unsure the use case for setting the encoding format

chubes4:
Thanks Darrin. I pushed updates to address the latest review feedback:

- Restored the existing generative result event contract so `AfterGenerateResultEvent::getResult()` remains `GenerativeAiResult`.
- Added embedding-specific before/after events instead of routing embeddings through the generative result event.
- Fixed batch embedding model selection to derive requirements from the provided batch prompts.
- Kept batch embeddings as a single model/provider call.
- Removed the generic `encodingFormat` API for now.

Also verified the updated client/provider branch pair in WP Codebox with a live OpenAI embeddings call.

src/Builders/PromptBuilder.php
```diff
@@ -1166,6 +1220,68 @@ public function generateImage(): File
...
+        if (!$model instanceof EmbeddingGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support embedding generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        $this->dispatchEvent(new BeforeGenerateEmbeddingEvent($promptMessages, $model, $capability));
```
dkotter: Currently the WP AI Client will fire `do_action` calls for events (see https://github.com/WordPress/wordpress-develop/blob/52e3c3013a9635b45d66c4888f995ade320bc49e/src/wp-includes/ai-client/adapters/class-wp-ai-client-event-dispatcher.php#L56). Right now that means we have two events, `wp_ai_client_before_generate_result` and `wp_ai_client_after_generate_result`.

With these new event classes, I believe that will add two new hooks, `wp_ai_client_before_generate_embedding` and `wp_ai_client_after_generate_embedding`. This does mean anyone that uses those existing hooks will not capture these new events. Is that something to worry about here? Similar to my last comment about backwards incompatibility, technically this works and won't cause errors but also is a new thing that people will need to be aware of.

Is there anyway to continue to use the existing event system and make it work for both generative AI results and embedding results? Or is there a way to use the generative AI results for embeddings instead of introducing a new DTO specific to embedding results? Or some other middle ground approach?

And maybe we're fine with these changes and the implications, just want to ensure we're thinking through the various ways this will impact any consumers of these events

dkotter: And just to be clear, opening this as something to discuss, not necessarily something that needs changed. Just want to ensure we're okay with introducing new events

chubes4: I will have to look into it later this week. I'm about to be off grid camping for a few days. Get back to you asap.

dkotter:
@chubes4 I've pushed a few changes here, namely modifying slightly how the embedding generation methods work to better align with all of our other generation methods (and avoid duplicate code). As an example, we have `generateText` and `generateTexts` methods that both end up calling `generateTextResult`. Same for our image generation methods.

Now we have the same for embeddings, where we have `generateEmbedding` and `generateEmbeddings` that both call `generateEmbeddingResult`. Also cleaned up a few other things and added integration tests so embeddings can actually be tested that way (you'll need to create a `.env` file, add your OpenAI key to that and then update the composer file to pull the OpenAI Provider from your embeddings branch, until that is merged).

If you have any concerns with any of these changes feel free to push back or modify things further. I haven't done anything yet in regards to the event handling, that may be something I ask in our AI call tomorrow to see if anyone else has concerns there.

src/Builders/PromptBuilder.php
dkotter: Thinking about this a bit more, I've landed on the approach in this PR being the right one, where we introduce new embedding only events instead of trying to force embeddings to use the existing event classes (especially since those are built around `GenerativeAiResult`).

Not needed for this PR but I do think there's a likely better future where we introduce a shared event interface that the existing events (and events in the future) can all use. A subscriber could then listen for all events there if desired.

In addition, I think a catch-all hook upstream in WordPress would also be useful, so someone wanting to track all events could use that instead of having to keep track of any new event hooks that may get added in the future.

Neither of those are blockers for this PR and I'd recommend those as a separate PR in this repo and one PR in the WordPress-develop repo.

src/Builders/PromptBuilder.php
```diff
@@ -1324,7 +1421,22 @@ protected function appendPartToMessages(MessagePart $part): void
...
+     * Gets the model to use for generation for the provided messages.
+     *
+     * @since n.e.x.t
+     *
+     * @param CapabilityEnum $capability The capability the model will be using.
+     * @param list<Message>  $messages The messages to derive requirements from.
+     * @return ModelInterface The model to use.
+     * @throws InvalidArgumentException If no suitable model is found or set model doesn't meet requirements.
+     */
+    private function getConfiguredModelForMessages(CapabilityEnum $capability, array $messages): ModelInterface
```
JasonTheAdams: Can you explain why we're adding this? Is `getConfiguredModel()` not taking the messages into account in some way? Do we need both methods?

src/Builders/PromptBuilder.php
```diff
@@ -1120,6 +1138,56 @@ public function generateVideoResult(): GenerativeAiResult
...
+     * @since n.e.x.t
+     *
+     * @param list<Prompt>|null $prompts Optional prompts to embed as a batch. If null, the builder's
+     *                                   own prompt is embedded as a single input.
+     * @return EmbeddingResult The generated embedding result.
+     * @throws InvalidArgumentException If a prompt or model validation fails.
+     * @throws RuntimeException If the model doesn't support embedding generation, or returns an
+     *                          embedding count that does not match the number of batch prompts.
+     */
+    public function generateEmbeddingResult(?array $prompts = null): EmbeddingResult
```
JasonTheAdams: Why are we passing `$prompts` as a parameter here? That's inconsistent with the way the messages are otherwise assembled. The `generateXResult` functions are terminating functions, they don't built the prompt.

src/Events/AfterGenerateEmbeddingEvent.php
```diff
@@ -0,0 +1,127 @@
...
+ * Event dispatched after embeddings have been generated.
+ *
+ * @since n.e.x.t
+ */
+class AfterGenerateEmbeddingEvent
+{
+    /**
+     * @var list<list<Message>> The prompts that were sent to the model.
+     */
+    private array $prompts;
```
JasonTheAdams: Let's call these messages to be consistent with the existing events already in the code.

JasonTheAdams (changes requested):
Some more suggestions, @chubes4! I'll try to be quick on feedback this week for quick turnaround.

src/Results/DTO/Embedding.php
dkotter: @JasonTheAdams I can remove that `$normalizedValues` array but any thoughts on the validation here? I thought about leaving that but not sure it's actually needed and can slightly help with performance to remove that, as we wouldn't have to iterate over each value, especially considering these embeddings are usually multiple thousands of values.

src/Builders/PromptBuilder.php
dkotter: So I didn't write this code but in parsing it to try and understand, I believe this is the thought/reasoning:

`getConfiguredModel` is hardwired to only use `$this->messages`. The change here is to keep `getConfiguredModel` but treat that as a thin wrapper around `getConfiguredModelForMessages`, which allows you to pass in the messages you want to use.

So any existing code that uses `getConfiguredModel` continues to work the same, it will just consider `$this->messages`. But these new embedding methods can directly use `getConfiguredModelForMessages` and pass in their own custom messages as needed.

The question then is why is that needed for embeddings? I believe the answer there is there's a legitimate use case where you pass in an array of messages you want to generate embeddings for, which is different than passing in a multi-turn prompt conversation, which is what the code is currently tuned for. So without this change, the model selection for batch embedding generation doesn't take into account the messages you pass in.

Now there may be other ways to solve this but I believe that gives a high-level explanation of the why

src/Results/DTO/Embedding.php
dkotter: Since we enforce the count of `$values` to equal the `$dimensions`, I've changed this to just return `$dimensions` instead of running `count` again, though let me know if you'd like a different approach: https://github.com/WordPress/php-ai-client/pull/244/changes/65ea6b09faee45947ff085cb0a9f7050c62ccd4a

src/Results/DTO/Embedding.php
dkotter: Added in https://github.com/WordPress/php-ai-client/pull/244/changes/c41a5826d5c265da6a833e2520e7965c7f6f55b0

src/Builders/PromptBuilder.php
dkotter: So I debated on the best path forward here as the existing structure doesn't really match well with how embeddings work.

Right now everything is built around the idea of Messages (which we also call prompts in multiple places), which are meant to be things like user messages, system messages, function responses, etc, to support single and multi-turn conversation.

With embedding generation though, you want to be able to pass in one or more pieces of data that you want to get embeddings for but there is no conversation happening.

After exploring a few options, I've landed on introducing a new `$inputs` param (and methods like `withInputs` and `inputs`) to handle this, similar to how normal generative AI results use `prompt` or `withText`. This means we no longer pass data into the `generate` methods and instead can do things like this:

```php
$embeddings = AiClient::inputs([
        'PHP powers a large part of the web.',
        'WordPress makes publishing accessible.',
    ])
    ->usingProvider('openai')
    ->generateEmbeddings();

$embedding = AiClient::prompt('PHP powers a large part of the web.')
    ->usingProvider('openai')
    ->generateEmbedding();

$embedding = AiClient::inputs([
        'PHP powers a large part of the web.',
    ])
    ->usingProvider('openai')
    ->generateEmbedding();

$embedding = AiClient::prompt()
    ->withInputs([
        'PHP powers a large part of the web.',
    ])
    ->usingProvider('openai')
    ->generateEmbedding();
```

This was updated in https://github.com/WordPress/php-ai-client/pull/244/changes/4c5439f9ea6a24d8133db2e0b187006a907f9a24 and https://github.com/WordPress/php-ai-client/pull/244/changes/1c7f010dad4f8bdafde1f67c50753d1d6da77b0f

Definitely open to suggestions on this approach though

src/Events/AfterGenerateEmbeddingEvent.php
dkotter: With the changes mentioned in a comment thread above, I've renamed this from `$prompts` to `$inputs` as that's what we're using now. But I can easily rename this to `$messages` if you want, though technically we don't pass actual `Messages` in here

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,173 @@
...
+     * @var int The vector dimension count.
+     */
+    private int $dimensions;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param list<mixed> $values The embedding vector values.
```
JasonTheAdams:
```suggestion
     * @param EmbeddingShape $values The embedding vector values.
```

src/Results/DTO/Embedding.php
```diff
@@ -0,0 +1,173 @@
...
+use JsonSerializable;
+use Traversable;
+use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+
+/**
+ * Represents a single generated embedding vector.
+ *
+ * @since n.e.x.t
+ *
+ * @phpstan-type EmbeddingShape list<float|int>
```
JasonTheAdams: I regret calling this `EmbeddingShape`. Let's call it `EmbeddingList` instead.

src/Results/DTO/Embedding.php
JasonTheAdams: I think we should still validate. While there are a lot of values, `is_int` and `is_float` are incredibly fast, C-level functions. These sorts of operations are negligible.

But assigning everything to a new array is unnecessary. We can just validate and use the original `$values` array.

JasonTheAdams (changes requested):
Thanks, @dkotter! I left a couple more comments.

src/Results/DTO/Embedding.php
dkotter: Updated that in https://github.com/WordPress/php-ai-client/pull/244/commits/5d75bc3395833db032519b187d4af6f824d64335

src/Results/DTO/Embedding.php
dkotter: Updated in https://github.com/WordPress/php-ai-client/pull/244/commits/5d75bc3395833db032519b187d4af6f824d64335

src/Results/DTO/Embedding.php
dkotter: So after changing from `list<mixed>` to `EmbeddingList`, PHPStan flagged the validation here as not needed, as `EmbeddingList` only allows integers and floats. So I've removed the validation for now but let me know if you'd like to proceed differently there

dkotter: Sorry, quick follow up after thinking about it more. I've left validation but moved it into it's own method so we can use `EmbeddingList` without PHPStan complaining. Updated in https://github.com/WordPress/php-ai-client/pull/244/commits/384044fd1470e377d341efe02ce4e1ec43dc16f1

JasonTheAdams: @dkotter Check out the `@phpstan-assert-if-true` notation in the [Narrowing Types](https://phpstan.org/writing-php-code/narrowing-types) doc. You can use that and move the whole loop into a private validation function.

dkotter: @JasonTheAdams Thanks for the tip! Updated the approach in https://github.com/WordPress/php-ai-client/pull/244/commits/860ce29bc6af6596a10a824e5670b160e6a1b13f

src/Builders/PromptBuilder.php
JasonTheAdams: This is actually making me think in a pretty different direction: Making a separate `EmbeddingBuilder`.

The `PromptBuilder` is about taking a prompt, a typical set of parameters, and generating something. Embeddings aren't actually about generating anything; they're about transforming something. If we add embeddings to `PromptBuilder` the majority of its parameters become irrelevant. To your point, we're not even talking about things like Messages anymore. This is why Embeddings are a completely separate API for providers.

Let's make a `ModelResolver` to extract all the model selection and such from `PromptBuilder` (which will also clean it up), and make that a property of both builders. We may even use a trait to add similar `withModel()` and such methods along with the property.

Let me know what y'all think!

chubes4: Agreed, and this is now implemented.

- Added a dedicated `EmbeddingBuilder` for direct text inputs.
- Extracted shared model/provider/config/preference/request-option resolution into a composed `ModelResolver`.
- Removed embedding state and generation methods from `PromptBuilder`.
- Changed the provider contract and embedding events from nested conversational messages to direct `list<string>` inputs.
- Preserved one provider call per batch, positional result validation, dimensions, and the existing embedding DTOs/events.

I also aligned WordPress/ai-provider-for-openai#34 with the new contract and verified single and two-input live OpenAI requests at 256 dimensions. The client is rebased on current `trunk`; PHPCS, PHPStan, 1,132 unit tests / 4,184 assertions, all GitHub PHP-version jobs, and Codecov are green.

Thanks for steering this toward the cleaner boundary. Ready for another look.

dkotter: Thanks for picking up these updates @chubes4 and apologies for not verifying with you as looks like we were both working on this at roughly the same time. I've pushed my updates to a new branch to not conflict with your changes here but if interested, can view the approach I took here:

- Full comparison of my new branch against trunk: https://github.com/WordPress/php-ai-client/compare/trunk...dkotter:php-ai-client:issue/242-embedding-generation-v2?expand=1
- And then just a comparison of your branch vs mine if interested to see where they diverge: https://github.com/chubes4/php-ai-client/compare/issue/242-embedding-generation...dkotter:php-ai-client:issue/242-embedding-generation-v2?expand=1

chubes4:
Reconciliation update: this branch now merges Darrin Kotter’s `issue/242-embedding-generation-v2` history into the canonical implementation, preserving all four commits and authorship.

The combined API keeps the dedicated `EmbeddingBuilder` and shared model resolver while adopting modality-aware `MessagePart` inputs, `AiClient::embed()` single/list ergonomics, `withInput()` / `withInputs()`, and `ModelRequirements::fromEmbeddingData()`. Embedding inputs remain independent items and providers receive one ordered batch.

Verification on the reconciled head:
- PHPCS passed
- PHPStan passed
- Full unit suite passed: 1,163 tests / 4,259 assertions
- OpenAI provider integration passed with one and two text inputs at 256 dimensions

The dependent provider update is in WordPress/ai-provider-for-openai#34.

src/AiClient.php
```diff
@@ -239,6 +243,28 @@ public static function prompt($prompt = null, ?ProviderRegistry $registry = null
...
+     * embedding generation method such as generateEmbedding() or generateEmbeddings() to produce
+     * one vector per input.
+     *
+     * @since n.e.x.t
+     *
+     * @param EmbeddingInput|list<EmbeddingInput>|null $input Optional initial input(s) to embed.
+     * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
+     * @return EmbeddingBuilder The embedding builder instance.
+     */
+    public static function embed($input = null, ?ProviderRegistry $registry = null): EmbeddingBuilder
```
JasonTheAdams: Let's go with `input()` as it reads better:

```php
// Reads weird
AiClient::embed('foo')->generateEmbedding();

// Reads nicely
AiClient::input('foo')->generateEmbedding();
```

The precedent is `AiClient::prompt()` which is about what you start with — starting with a prompt and then generating something from it. With embeddings the parameters are called "inputs", and from inputs you generate embeddings.

src/Builders/PromptBuilder.php
```diff
@@ -1610,20 +1233,22 @@ private function parseMessage($input, MessageRoleEnum $defaultRole): Message
      */
     private function validateMessages(): void
     {
-        if (empty($this->messages)) {
+        $messages = $this->messages;
```
JasonTheAdams, line 1237: This is a bit tucked in here and doesn't technically have to do with the broader work. I'm assuming we're doing this so the internal pointer of `$this->messages` isn't affected by `reset()` and `end()` being used? If so, let's leave a comment to clarify that.

src/Builders/PromptBuilder.php
JasonTheAdams, line 1: I'm really happy to see this massive class get trimmed down quite a bit. 😄

src/Builders/EmbeddingBuilder.php
```diff
@@ -0,0 +1,333 @@
...
+     * @param EmbeddingInput $input The input to embed.
+     * @return self
+     * @throws InvalidArgumentException If the input is invalid.
+     */
+    public function withInput($input): self
+    {
+        $this->inputs[] = $this->parseInput($input);
+
+        return $this;
+    }
```
JasonTheAdams: Let's combine these into a single, variadic `withInput(...$input)` method.

JasonTheAdams (changes requested):
Alright! We're close! Just a few small suggestions.

jeffpaul:
@JasonTheAdams I don't know that anyone else is still around to resolve those final items and merge, but If you're able to do that then I can try and see about getting a release out later this evening (after which we'd still need someone to bump the library in WP core before too long into tomorrow).  All that to say, maybe things are too late now?

src/AiClient.php
chubes4: Updated in 49452a2. The fluent entry point is now `AiClient::input()` throughout source, CLI, tests, README, integration examples, and architecture docs.

src/Builders/EmbeddingBuilder.php
chubes4: Updated in 49452a2. `withInput(...$input)` is now the single append method, with coverage for multiple inputs and an empty invocation; `withInputs()` was removed from source and documentation.

src/Builders/PromptBuilder.php
chubes4: Exactly. Added a concise comment in 49452a2 explaining that the local copy prevents `reset()` and `end()` from changing the property array internal pointer.

chubes4:
It's not too late for me @jeffpaul @JasonTheAdams let's cook this PR. I just pushed an update.

src/Builders/EmbeddingBuilder.php
dkotter: I had an integration test that was doing this:

```php
$embedding = AiClient::input()
            ->withInput( array( 'PHP powers a large part of the web.' ) )
            ->usingProvider('google')
            ->generateEmbedding();
```

And it no longer passes with the latest changes. Instead you have to pass in a single string into `withInput`. The answer here could just be update the test but wondering if we want to support the above approach still?

chubes4: Good catch. I agree the list form should remain supported, both for consistency with `AiClient::input([…])` and to allow callers to choose list or variadic input syntax.

Updated in 9d623f9. These now both work:

```php
->withInput(['first', 'second'])
->withInput('first', 'second')
```

Associative message-part array shapes remain single inputs, while list arrays are expanded into independent inputs. I added unit coverage for list, variadic, and empty-list handling, changed the OpenAI integration test to exercise the exact array-wrapped single-input form, and ran a fresh mirrored cross-package live OpenAI check: one list input returned one 256-dimensional embedding; two variadic inputs returned two ordered 256-dimensional embeddings. Thanks for catching the gap.

JasonTheAdams: You don't actually need to do that. PHP supports spreading list arrays:
```php
->withInput(...['first', 'second']) // spreads over variadic parameter
```

chubes4: Agreed. Simplified in 66fa4d7: `withInput()` now handles only its variadic `EmbeddingInput` arguments, and callers pass lists with PHP spread syntax (`withInput(...$inputs)`). `AiClient::input([...])` retains its constructor convenience by spreading list input before delegating. Updated the unit/integration coverage and README accordingly. PHPCS and PHPStan pass; the full unit suite passes with 1,164 tests / 4,260 assertions.

JasonTheAdams (approved):
Ran the integration tests locally to confirm this is working against https://github.com/WordPress/ai-provider-for-openai/pull/34, and it worked!

Thanks @chubes4 and @dkotter!

PR #247, opened 2026-06-04
Bump minimum PHP requirement to 7.4.12 to avoid covariant return type…

### LINKED ISSUE #240: PHP 7.4.0-7.4.11 fatal due to covariant return type bug in php-ai-client DTOs
The plugin declare PHP 7.4+ support, but on PHP 7.4.0 the AI Client DTO inheritance pattern can fatal with:

_Declaration of GenerativeAiResult::fromArray(array $array): GenerativeAiResult must be compatible with WithArrayTransformationInterface::fromArray(array $array): WithArrayTransformationInterface_

This appears to be PHP bug (80126), fixed in PHP 7.4.12:

https://bugs.php.net/bug.php?id=80126
https://www.php.net/ChangeLog-7

So the practical minimum for the PHP SDK path may be PHP 7.4.12+, not PHP 7.4.0+.

Could the compatibility docs/plugin headers/runtime checks clarify that PHP 7.4.12+ is required for the PHP AI Client SDK path, or add a guard/admin notice for PHP 7.4.0-7.4.11?

vyskoczilova:
## Summary

Fixes #240 — PHP 7.4.0–7.4.11 have a known PHP engine bug ([bug #80126](https://bugs.php.net/bug.php?id=80126)) that causes a fatal error with the covariant return type pattern used throughout this library's DTOs:

> Declaration of `GenerativeAiResult::fromArray(array $array): GenerativeAiResult` must be compatible with `WithArrayTransformationInterface::fromArray(array $array): WithArrayTransformationInterface`

The bug was fixed in PHP 7.4.12. This PR raises the declared minimum from `>=7.4` to `>=7.4.12` and adds a runtime guard so the problem surfaces as a clear, actionable message rather than a cryptic fatal error.

### `composer.json`
- `require.php`: `>=7.4` → `>=7.4.12`
- `config.platform.php`: `7.4` → `7.4.12`

### `src/polyfills.php`
- Added a `trigger_error(E_USER_WARNING)` guard at load time for PHP 7.4.0–7.4.11. This file is autoloaded via Composer's `files` mechanism, so the warning fires on the very first request — before any DTO class is loaded and before a cryptic fatal can occur.

### `README.md`
- Added a **Requirements** section listing PHP 7.4.12+ with a link to the PHP bug report.

### `CONTRIBUTING.md`
- Updated the PHP Compatibility section to state 7.4.12 as the minimum and explain why.

### `AGENTS.md`
- Updated the key constraints list to reflect 7.4.12.

### `phpcs.xml.dist`
- Updated the inline comment to reference 7.4.12+. The `testVersion` value stays `7.4-` because PHPCompatibility does not support patch-level version filtering — this is correct and intentional.

## Notes

- No logic changes, no API surface changes.
- All 1088 unit tests pass.
- The `trigger_error` approach (rather than throwing an exception) is deliberate: it fires early without hard-aborting, matches the WordPress-agnostic nature of the library, and follows the same conditional-check pattern already used in `polyfills.php`.

Closes WordPress/php-ai-client#240

---

https://profiles.wordpress.org/vyskoczilova/

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,7 @@ All code in this project MUST adhere to the coding standards, naming conventions
 
 Key constraints include:
 
-*   PHP 7.4 as the minimum required version.
+*   PHP 7.4.12 as the minimum required version (PHP 7.4.0–7.4.11 have a known covariant return type bug that causes fatal errors with this codebase).
 *   PER Coding Style (extending PSR-12).
 *   Strict type hinting for all parameters, return values, and properties.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ Note that `list<string>` and `string[]` _are not_ the same. The latter is an ali
 
 ## PHP Compatibility
 
-All code must be backward compatible with PHP 7.4, which is the minimum required PHP version for this project.
+All code must be backward compatible with PHP 7.4.12, which is the minimum required PHP version for this project. PHP 7.4.0–7.4.11 have a known engine bug (https://bugs.php.net/bug.php?id=80126) that causes a fatal error with covariant return types in DTO inheritance; those patch releases are not supported.
 
 ## Running Tests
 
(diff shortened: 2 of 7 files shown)


vyskoczilova:
Done, sorry for the mess with the composer.lock, but fixed now.

felixarntz (commented):
@vyskoczilova Thank you for the PR!

Generally, this seems reasonable to me, but we need to consider how this affects WordPress Core, which relies on this package.

@johnbillion What is the exact minimum PHP version supported by WordPress now? is it 7.4 exactly? or any particular patch version of it?

johnbillion:
The minimum is 7.4.0: https://make.wordpress.org/core/2026/01/09/dropping-support-for-php-7-2-and-7-3/

justlevine:
Per last week's contributor call, since we need to keep our minimum PHP version in sync with core, the path forward on #240 is to move the "problematic" types out of the signatures and into doc-blocks. 

Thanks everyone for your contribution here 🙇

PR #255, opened 2026-06-18
Add support for streaming

### LINKED ISSUE #100: Add support for streaming
See #8: A crucial piece of functionality still missing in our SDK is response streaming - a quite complex piece of functionality to get right, especially with the incomplete documentation of how different providers implement it. But we need it, that's for sure.

What we'll need to do is:
* Implement the primitives needed for streaming (likely some kind of stream response class, potentially a new stream method on `HttpTransporter`).
* Implement actual streaming for the three providers (which can fortunately be centralized in the OpenAI API compatible implementation for now).

Putting here some references for inspiration in how I implemented streaming (still with some limitations) in the AI Services plugin for WordPress:
* https://github.com/felixarntz/ai-services/blob/main/includes/Services/HTTP/Contracts/Stream_Request_Handler.php
* https://github.com/felixarntz/ai-services/blob/main/includes/Services/HTTP/Stream_Response.php
* https://github.com/felixarntz/ai-services/blob/main/includes/Services/Traits/Generative_AI_API_Client_Trait.php#L182
* https://github.com/felixarntz/ai-services/blob/main/includes/Services/Base/OpenAI_Compatible_AI_Text_Generation_Model.php#L118

There's definitely things to improve there which we should do here, but it's a decent starting point to review this and come up with effectively a v2 of that implementation.

thelovekesh:
## Summary
Fixes https://github.com/WordPress/php-ai-client/issues/100

This PR introduces the core changes needed for stable streaming support in text generation.

Here is a script to test streaming with text generation. It supports streaming of reasoning, content and tool calls.

<details><summary>stream-text.php</summary>

```php
<?php
/**
 * Usage:
 *   AI_API_KEY=sk-... php stream-text.php "Tell me a short joke"
 *
 * Optional env:
 *   AI_BASE_URL  Base URL incl. version path. Default: https://api.openai.com/v1/
 *                DeepSeek: https://api.deepseek.com/   Groq: https://api.groq.com/openai/v1/
 *   AI_MODEL     Model id. Default: gpt-4o-mini
 */

declare(strict_types=1);

use GuzzleHttp\Client;
use WordPress\AiClient\Messages\DTO\MessagePart;
use WordPress\AiClient\Messages\DTO\UserMessage;
use WordPress\AiClient\Providers\DTO\ProviderMetadata;
use WordPress\AiClient\Providers\Enums\ProviderTypeEnum;
use WordPress\AiClient\Providers\Http\DTO\ApiKeyRequestAuthentication;
use WordPress\AiClient\Providers\Http\DTO\Request;
use WordPress\AiClient\Providers\Http\Enums\HttpMethodEnum;
use WordPress\AiClient\Providers\Http\HttpTransporter;
use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
use WordPress\AiClient\Providers\Models\DTO\ModelMetadata;
use WordPress\AiClient\Providers\Models\Enums\CapabilityEnum;
use WordPress\AiClient\Providers\OpenAiCompatibleImplementation\AbstractOpenAiCompatibleTextGenerationModel;
use WordPress\AiClient\Tools\DTO\FunctionDeclaration;

require_once __DIR__ . '/vendor/autoload.php';

if (!class_exists(Client::class)) {
    fwrite(STDERR, "Guzzle is required. Install it:\n  composer require --dev guzzlehttp/guzzle\n");
    exit(1);
}

$apiKey = getenv('AI_API_KEY') ?: '';
if ($apiKey === '') {
    fwrite(STDERR, "Set AI_API_KEY, e.g.:\n  AI_API_KEY=sk-... php stream-text.php \"Hello\"\n");
    exit(1);
}

$modelId = getenv('AI_MODEL') ?: 'gpt-4o-mini';
$baseUrl = getenv('AI_BASE_URL') ?: 'https://api.openai.com/v1/';

$prompt = $argv[1] ?? 'Plan a 12-day trip across Japan and South Korea in May 2026 for 2 adults '
    . 'and 1 child, budget about $9000, prefer trains between cities, with a couple of activities per stop.';

$model = new class ($modelId, $baseUrl) extends AbstractOpenAiCompatibleTextGenerationModel {
    private string $baseUrl;

    public function __construct(string $modelId, string $baseUrl)
    {
        parent::__construct(
            new ModelMetadata($modelId, $modelId, [CapabilityEnum::textGeneration()], []),
            new ProviderMetadata('openai-compatible', 'OpenAI-compatible', ProviderTypeEnum::cloud())
        );
        $this->baseUrl = rtrim($baseUrl, '/') . '/';
    }

    protected function createRequest(
        HttpMethodEnum $method,
        string $path,
        array $headers = [],
        $data = null
    ): Request {
        return new Request($method, $this->baseUrl . ltrim($path, '/'), $headers, $data);
    }
};

$planTrip = new FunctionDeclaration(
    'plan_trip',
    'Build a detailed multi-city travel itinerary from the user request.',
    [
        'type' => 'object',
        'properties' => [
            'travelers' => [
                'type' => 'object',
                'properties' => [
                    'adults' => ['type' => 'integer', 'minimum' => 1],
                    'children' => ['type' => 'integer', 'minimum' => 0],
                ],
                'required' => ['adults'],
            ],
            'date_range' => [
                'type' => 'object',
                'properties' => [
                    'start' => ['type' => 'string', 'description' => 'ISO 8601 date, e.g. 2026-05-01'],
                    'end' => ['type' => 'string', 'description' => 'ISO 8601 date'],
                ],
                'required' => ['start', 'end'],
            ],
            'budget_usd' => ['type' => 'number', 'description' => 'Total budget in USD.'],
            'primary_transport' => [
                'type' => 'string',
                'enum' => ['flight', 'train', 'car', 'mixed'],
            ],
            'stops' => [
                'type' => 'array',
                'description' => 'Ordered list of cities to visit.',
                'items' => [
                    'type' => 'object',
                    'properties' => [
                        'city' => ['type' => 'string'],
                        'country' => ['type' => 'string'],
                        'nights' => ['type' => 'integer', 'minimum' => 1],
                        'activities' => [
                            'type' => 'array',
                            'items' => ['type' => 'string'],
                        ],
                        'accommodation' => [
                            'type' => 'string',
                            'enum' => ['hostel', 'hotel', 'ryokan', 'apartment'],
                        ],
                    ],
                    'required' => ['city', 'country', 'nights'],
                ],
            ],
        ],
        'required' => ['travelers', 'date_range', 'stops'],
    ]
);

$config = new ModelConfig();
$config->setFunctionDeclarations([$planTrip]);
$config->setSystemInstruction(
    'You are a meticulous travel planner. First write one short sentence telling the user what '
    . 'you are about to do, then call the plan_trip tool with detailed, structured arguments.'
);
$config->setCustomOption('tool_choice', 'required');
$model->setConfig($config);

$model->setHttpTransporter(new HttpTransporter(new Client()));
$model->setRequestAuthentication(new ApiKeyRequestAuthentication($apiKey));

fwrite(STDERR, "Streaming from {$baseUrl} ({$modelId})\n\n");

try {
    $stream = $model->streamGenerateTextResult([new UserMessage([new MessagePart($prompt)])]);

    $section = '';
    $toolFragments = 0;
    foreach ($stream as $chunk) {
        $reasoning = $chunk->getReasoningDeltaText();
        if ($reasoning !== '') {
            if ($section !== 'reasoning') {
                echo "\n[reasoning]\n";
                $section = 'reasoning';
            }
            echo $reasoning;
            flush();
        }

        $answer = $chunk->getDeltaText();
        if ($answer !== '') {
            if ($section !== 'answer') {
                echo PHP_EOL . '[answer] ';
                $section = 'answer';
            }
            echo $answer;
            flush();
        }

        foreach ($chunk->getToolCallDeltas() as $toolCall) {
            $name = $toolCall->getFunctionName();
            if ($name !== null) {
                echo PHP_EOL . "[tool-call] {$name} args (each ⟨…⟩ is one streamed chunk):" . PHP_EOL;
                $section = 'tool';
            }
            $fragment = $toolCall->getArgumentsFragment();
            if ($fragment !== '') {
                echo '⟨' . $fragment . '⟩';
                $toolFragments++;
                flush();
            }
        }
    }
    echo "\n";

    $result = $stream->getFinalResult();

    foreach ($result->getCandidates() as $candidate) {
        foreach ($candidate->getMessage()->getParts() as $part) {
            $functionCall = $part->getFunctionCall();
            if ($functionCall !== null) {
                fwrite(STDERR, sprintf(
                    "\n[assembled call] %s(%s)",
                    (string) $functionCall->getName(),
                    (string) json_encode($functionCall->getArgs(), JSON_PRETTY_PRINT)
                ));
            }
        }
    }

    $usage = $result->getTokenUsage();
    fwrite(STDERR, sprintf(
        "\n\n[done] tool-call fragments received: %d | tokens: %d prompt + %d completion = %d total\n",
        $toolFragments,
        $usage->getPromptTokens(),
        $usage->getCompletionTokens(),
        $usage->getTotalTokens()
    ));
} catch (\Throwable $e) {
    fwrite(STDERR, "\n[error] " . $e->getMessage() . "\n");
    exit(1);
}

```

</details> 

### Tasks

- [x] Add stream support in `Request` and `Response`.
- [x] Add `EventStreamParserInterface` and SSE Parser.
- [x] Add streaming support for text generation covers both reasoning and content.
- [x] Update `AbstractOpenAiCompatibleTextGenerationModel` to use streaming in text generation.
- [x] Add initial implementation for chunk aggregation.
- [x] Add event dispatch on text result generation with streaming.
- [x] Update prompt builder and ai client to consume streaming for text generation.
- [x] Add streaming support for function calls.

### DIFF
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -1064,7 +1064,8 @@ class PromptBuilder
         $this->includeOutputModalities(ModalityEnum::text());
         $this->validateMessages();
 
-        $model = $this->getConfiguredModel(CapabilityEnum::textGeneration());
+        $capability = CapabilityEnum::textGeneration();
+        $model = $this->getConfiguredModel($capability);
 
         if (!$model instanceof StreamingTextGenerationModelInterface) {
             throw new RuntimeException(
@@ -1075,7 +1076,15 @@ class PromptBuilder
             );
         }
 
-        return $model->streamGenerateTextResult($this->messages);
+        $messages = $this->messages;
+
+        $this->dispatchEvent(new BeforeGenerateResultEvent($messages, $model, $capability));
+
+        return $model->streamGenerateTextResult($messages)->onComplete(
+            function (GenerativeAiResult $result) use ($messages, $model, $capability): void {
+                $this->dispatchEvent(new AfterGenerateResultEvent($messages, $model, $capability, $result));
+            }
+        );
     }
 
     /**
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -26,11 +26,13 @@ use WordPress\AiClient\Providers\Models\DTO\ModelRequirements;
 use WordPress\AiClient\Providers\Models\Enums\CapabilityEnum;
 use WordPress\AiClient\Providers\Models\ImageGeneration\Contracts\ImageGenerationModelInterface;
 use WordPress\AiClient\Providers\Models\SpeechGeneration\Contracts\SpeechGenerationModelInterface;
+use WordPress\AiClient\Providers\Models\TextGeneration\Contracts\StreamingTextGenerationModelInterface;
 use WordPress\AiClient\Providers\Models\TextGeneration\Contracts\TextGenerationModelInterface;
 use WordPress\AiClient\Providers\Models\TextToSpeechConversion\Contracts\TextToSpeechConversionModelInterface;
 use WordPress\AiClient\Providers\Models\VideoGeneration\Contracts\VideoGenerationModelInterface;
 use WordPress\AiClient\Providers\ProviderRegistry;
 use WordPress\AiClient\Results\DTO\GenerativeAiResult;
+use WordPress\AiClient\Results\StreamedGenerativeAiResult;
 use WordPress\AiClient\Tools\DTO\FunctionDeclaration;
 use WordPress\AiClient\Tools\DTO\FunctionResponse;
 use WordPress\AiClient\Tools\DTO\WebSearch;
@@ -1046,6 +1048,53 @@ class PromptBuilder
 
         // Generate and return the result with text generation capability
         return $this->generateResult(CapabilityEnum::textGeneration());
+    }
+
+    /**
+     * Streams a text result from the prompt.
+     *
+     * @since n.e.x.t
+     *
+     * @return StreamedGenerativeAiResult The streamed result.
+     * @throws InvalidArgumentException If the prompt or model validation fails.
+     * @throws RuntimeException If the model does not support streaming text generation.
+     */
+    public function streamGenerateTextResult(): StreamedGenerativeAiResult
+    {
+        $this->includeOutputModalities(ModalityEnum::text());
+        $this->validateMessages();
+
+        $model = $this->getConfiguredModel(CapabilityEnum::textGeneration());
+
+        if (!$model instanceof StreamingTextGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support streaming text generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        return $model->streamGenerateTextResult($this->messages);
+    }
+
+    /**
+     * Streams generated text from the prompt as it arrives.
+     *
+     * @since n.e.x.t
+     *
+     * @return iterable<string> The text deltas, in order.
+     * @throws InvalidArgumentException If the prompt or model validation fails.
+     * @throws RuntimeException If the model does not support streaming text generation.
+     */
+    public function streamGenerateText(): iterable
+    {
+        foreach ($this->streamGenerateTextResult() as $chunk) {
+            $delta = $chunk->getDeltaText();
+            if ($delta !== '') {
+                yield $delta;
+            }
+        }
     }
 
     /**
--- a/src/Builders/PromptBuilder.php
+++ b/src/Builders/PromptBuilder.php
@@ -9,6 +9,7 @@ use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
 use WordPress\AiClient\Events\AfterGenerateResultEvent;
 use WordPress\AiClient\Events\BeforeGenerateResultEvent;
+use WordPress\AiClient\Events\GenerateResultErrorEvent;
 use WordPress\AiClient\Files\DTO\File;
 use WordPress\AiClient\Files\Enums\FileTypeEnum;
 use WordPress\AiClient\Files\Enums\MediaOrientationEnum;
@@ -1078,13 +1079,16 @@ class PromptBuilder
 
         $messages = $this->messages;
 
-        $this->dispatchEvent(new BeforeGenerateResultEvent($messages, $model, $capability));
-
-        return $model->streamGenerateTextResult($messages)->onComplete(
-            function (GenerativeAiResult $result) use ($messages, $model, $capability): void {
+        return $model->streamGenerateTextResult($messages)
+            ->onStart(function () use ($messages, $model, $capability): void {
+                $this->dispatchEvent(new BeforeGenerateResultEvent($messages, $model, $capability));
+            })
+            ->onComplete(function (GenerativeAiResult $result) use ($messages, $model, $capability): void {
                 $this->dispatchEvent(new AfterGenerateResultEvent($messages, $model, $capability, $result));
-            }
-        );
+            })
+            ->onError(function (\Throwable $error) use ($messages, $model, $capability): void {
+                $this->dispatchEvent(new GenerateResultErrorEvent($messages, $model, $capability, $error));
+            });
     }
 
     /**
--- a/src/Results/StreamedGenerativeAiResult.php
+++ b/src/Results/StreamedGenerativeAiResult.php
@@ -85,9 +85,17 @@ final class StreamedGenerativeAiResult implements IteratorAggregate
      * @since n.e.x.t
      *
      * @return Generator<int, GenerativeAiResultChunk> The chunks, in order.
+     *
+     * @throws RuntimeException If the source stream has already been consumed.
      */
     public function getIterator(): Generator
     {
+        if ($this->started) {
+            throw new RuntimeException(
+                'This streamed result has already been consumed; the stream can be read only once.'
+            );
+        }
+
         while (true) {
             $chunk = $this->pull();
             if ($chunk === null) {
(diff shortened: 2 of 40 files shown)


thelovekesh:
Will add docs once the API is reviewed and approved.

src/Builders/PromptBuilder.php
```diff
@@ -1048,6 +1051,65 @@ public function generateTextResult(): GenerativeAiResult
...
+        $model = $this->getConfiguredModel($capability);
+
+        if (!$model instanceof StreamingTextGenerationModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support streaming text generation.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
```
thelovekesh, line 891: If a model doesn't implement this interface, it will fail even if the model supports text generation. It fails because streaming is gated behind this opt-in interface that no provider actually implements. We can't discover our way around it either, since streaming isn't a `CapabilityEnum`, so discovery can't tell which models stream and may still pick one that doesn't.

The clean solution is to remove `StreamingTextGenerationModelInterface` and rather add `streamGenerateTextResult(array $prompt): StreamedGenerativeAiResult` to `TextGenerationModelInterface`. Streaming is a networking primitive, not a real model capability, so every text model can stream: natively where the provider's API supports it, or by emulating it (one `generateTextResult()` call yielded as a single chunk) where it doesn't. Adding a method to the interface sounds like a breaking change, but if we put the emulation default in the shared base (`AbstractApiBasedModel`) that every provider already extends, they all inherit it and nothing breaks.

Another way is to add a new capability and wire discovery to find a model with it, but that isn't standard given streaming isn't a model-level capability, it's a networking primitive, not something a model generates. We'd also still need a method on an interface plus every provider opting in, so it's more machinery for something that isn't really model-level.

thelovekesh:
Created a tiny server to see streaming in action - https://gist.github.com/thelovekesh/a05a99570989672ad4b493b81a4b6dc3

src/Builders/PromptBuilder.php
```diff
@@ -1048,6 +1051,65 @@ public function generateTextResult(): GenerativeAiResult
...
     }
 
+    /**
+     * Streams a text result from the prompt.
+     *
+     * @since n.e.x.t
+     *
+     * @return StreamedGenerativeAiResult The streamed result.
+     * @throws InvalidArgumentException If the prompt or model validation fails.
+     * @throws RuntimeException If the model does not support streaming text generation.
```
justlevine, line 874: Didn't do a real review yet, just a quick note: can you make sure the PHPDoc types use FQCN? E.g.

```suggestion
     * @return \WordPress\AiClient\Results\StreamedGenerativeAiResult The streamed result.
     * @throws \WordPress\AiClient\Common\Exception\InvalidArgumentException If the prompt or model validation fails.
     * @throws \WordPress\AiClient\Common\Exception\RuntimeException If the model does not support streaming text generation.
```

thelovekesh: wouldn't it be resolved automatically from the namespace aliases?

src/Results/StreamedGenerativeAiResult.php
thelovekesh, line 1: I think we should allow passing some sort of abort controller here, so users have a way to close a stream midway. This is helpful for freeing up resources and avoiding connection leaks when the client connection is aborted (e.g. a browser tab is closed mid-stream).

A few use cases:
- **Browser tab closed mid-stream** - `connection_aborted()` can detect the dropped client connection, and the stream can be closed.
- **CLI abort** - detect signals like `SIGINT` and close the stream.
- **Cron/queue jobs** - close the stream on error or a configured timeout.

src/Builders/PromptBuilder.php
justlevine: You'd think, but nope, PHPDoc and IDEs are weird.

thelovekesh: practically there's no functional difference, both resolve the same way. on my end the IDE and phpstan both pick up the types fine (phpstan parses the docblocks too), so i don't think there's a correctness issue. if someone is seeing different behavior, it could be down to indexing or IDE config.

per the phpdoc spec both are valid, so it's mostly a convention. our existing docblocks already use imported names consistently, so converting just these to FQCNs would make the file inconsistent and wrap the comments. i'd keep them as-is and leave that out of scope for this PR.

justlevine: Hm, that's weird I assumed we were enforcing that rule here the way we do in the other AI Team repos. So yeah, I guess we can punt to a follow-up PRs that addresses the whole repo. 🤔

> practically there's no functional difference, both resolve the same way. 

The DX issue occurs downstream with stub generation. Once the file has more than one namespace both PHPStan and the IDE have trouble.
 E.g. 
image

thelovekesh: can I get some repro steps? maybe something that can be reported upstream.

justlevine:
```bash
git clone https://github.com/WordPress/ai
cd ai
composer require php-stubs/wordpress-stubs:"7.0 as 6.9.4" && composer update -W
composer run phpstan
```

Or if you have any existing setup using `szepeviktor/phpstan-wordpress` or `php-stubs/wordpress-stubs` directly, just try to interact with anything that uses a non FQCN docblock.

justlevine: https://github.com/WordPress/php-ai-client/issues/257

jeffpaul:
@JasonTheAdams slow ping here to get your review so this could potentially get into a future release (and then perhaps into an upcoming WP core release).

justlevine:
> slow ping here
@JasonTheAdams slightly faster ping. 

Anything we can do on our end to help speed up your review on this (besides handling the small merge conflict in cli.php?)

whyisjake:
For anyone reviewing: I ran a multi-reviewer pass over this PR at `4e7197b` and stacked the resulting fixes on the branch as thelovekesh/php-ai-client#1. It addresses seven confirmed findings (three P1: a truncated stream finalized as a clean `stop`, an unbounded empty-read loop in the SSE parser, and `Response::getStream()` returning a drained stream after `getBody()`), each as its own commit with tests, and lists three design questions left for maintainers. Unit suite, PHPCS and PHPStan are clean on that branch.

Issue #260, opened 2026-07-14
ModelMetadata has no context-window concept, and TokenLimitReachedException is never thrown


georgestephanis:
## Summary

`ModelMetadata` has no concept of a model's context window (maximum input/total token count), and `TokenLimitReachedException` is declared in the SDK but never thrown anywhere in it. There's currently no way for a provider, a `PromptBuilder` consumer, or a plugin built on top of the SDK to know how much room a model has for a given prompt, or to be told proactively that a prompt won't fit — the only way to find out is to send the request and let the underlying API reject it (or silently truncate/misbehave, depending on the server).

## Current behavior

- `ModelMetadata` (`src/Providers/Models/DTO/ModelMetadata.php`) exposes `id`, `name`, `supportedCapabilities`, and `supportedOptions` — nothing describing the model's context window or max input tokens.
- `ModelConfig::maxTokens` (→ `OptionEnum::maxTokens()`) only ever maps to the *output*-side `max_tokens` request parameter. It's a per-request cap on generation length, unrelated to how much input the model can accept.
- `TokenLimitReachedException` (`src/Common/Exception/TokenLimitReachedException.php`) exists, with a docblock describing it as being thrown "when token usage exceeds the model's context window or a configured maximum," and carries a `getMaxTokens()` accessor — but grepping the SDK, it is never actually thrown by any provider implementation or by `PromptBuilder`. It's a defined contract with no implementation.
- `PromptBuilder` builds the message list to send without any token estimation or trimming.

## Why this matters

Callers that maintain their own conversation history (chat UIs, agents, plugins) have no way to:
1. Discover a model's context window size from the SDK, so they could truncate/summarize history themselves.
2. Get a clear, typed exception *before* firing a request that's too large, rather than an opaque provider-specific HTTP error.

This came up while working on an OpenAI-compatible-server connector plugin: we let site owners configure a "context length" for their self-hosted model, but the SDK gives us nowhere to plug that in on the input side — it can only be forwarded to the server as a best-effort hint (e.g. Ollama's `num_ctx`), not used by the SDK itself to warn or trim before sending.

## Relationship to #193

#193 ("Missing Required Parameters in Tool Calls When Token Limit Reached") is adjacent but describes a different failure mode: *output*-side truncation, where generation stops mid-tool-call because `finish_reason` hit the length cap, and asks for `TokenLimitReachedException` to be thrown reactively when that happens.

[...]

PR #264, opened 2026-07-24
Add model context window metadata and proactive token-limit check

the-hercules:
Closes - #260 
## Summary

Adds an **input-side** counterpart to the existing output-side `maxTokens` cap. Today the SDK has no concept of a model's context window, and `TokenLimitReachedException` is declared but never thrown. This PR lets callers (chat UIs, agents, plugins) discover a model's context window and be told — with a typed, catchable exception, *before* a request is sent — when an assembled prompt is estimated to be too large, instead of relying on an opaque provider-side HTTP error.

Implements the input-side of #260 (context-window metadata + proactive budgeting). The reactive/output-side (#193) is intentionally left as a follow-up; both can share the same `TokenLimitReachedException`.

## What changed

**Metadata**
- `ModelMetadata` gains an optional `contextWindow` (`?int`, total input + output tokens). Nullable, defaults to `null`, backward compatible: constructor param is trailing/optional, included in `toArray()`/schema only when set, read in `fromArray()` without becoming required. Rejected if `< 1`.

**Token counting (BYOB)**
- New `TokenCounterInterface` (`countTokens(list<Message>, ModelMetadata): int`).
- New default `HeuristicTokenCounter`: a rough, dependency-free `strlen / 4` estimate over text parts, documented as advisory. Non-text parts (files, tool calls/responses) are not counted.
- New `PromptBuilder::usingTokenCounter()` to inject an accurate, model-specific counter per request.

**Proactive check**
- `PromptBuilder::generateResult()` now estimates the prompt (message text + system instruction, plus the configured output `maxTokens` when set) and throws `TokenLimitReachedException` when the total exceeds the resolved model's `contextWindow`. **No-op when the context window is unknown**, so existing behavior is preserved until providers populate the field.

**Docs**
- Clarified `contextWindow` (input budget) vs `ModelConfig::maxTokens` (output cap) in docblocks, `docs/ARCHITECTURE.md`, and `docs/GLOSSARY.md`.

## Design notes / assumptions

- `contextWindow` is the **total** window (input + output), matching published specs — not input-only.
- The check is **always-on when a window is known, but advisory**: a heuristic can be wrong, so it can false-positive; `usingTokenCounter()` is the escape hatch, and the docblock states a passing check is not a guarantee.
- Override is **per-call only**; registry-level/provider propagation (the `HttpTransporter`-style parity) is deliberately out of scope here.
- No real per-model context-window numbers are shipped — those belong in the provider packages; core stays `null`.
- The check is not capability-gated (runs for any model exposing a window; only text is counted). Easy to restrict to text generation if preferred.

## Relationship to #193

#193 covers the reactive/output side (truncation via `finish_reason == length`). This PR is the proactive/input side. The exception and its docblock cover both cases so #193 can build on this; the `length` finish reason already flows through to `Candidate::getFinishReason()->isLength()`.

## Testing

- `composer lint` (PHPCS PSR-12/PER + Slevomat, and PHPStan level `max`): clean.
- `composer test:unit`: full suite passes, including new tests for the `contextWindow` field, `HeuristicTokenCounter`, and the `PromptBuilder` check (throws when over, no-op when the window is unknown, honors reserved output, honors an injected counter).

## AI usage disclosure

This PR was developed collaboratively with Claude, in a Claude Code session, under human direction and review.

- **Model used:** Claude Opus 4.8 (`claude-opus-4-8`), via Claude Code.
- **How it was used:**
  - Investigated the SDK to confirm the gap (`ModelMetadata`, `ModelConfig`, `TokenLimitReachedException`, `PromptBuilder`) and reviewed related issue #193.
  - Studied the repo's conventions (DTO patterns, the `HttpTransporter`-style override mechanism, exception/`finish_reason` handling, test structure, PHP 7.4 / PHPStan-max / PSR-12 constraints) before writing code.
  - Drafted the implementation, unit tests, and the doc updates (`ARCHITECTURE.md`, `GLOSSARY.md`), and wrote this PR description and the issue comment.
  - Ran `composer lint` (PHPCS + PHPStan level `max`) and `composer test:unit`, plus an offline demo script, to verify behavior.
- **Human involvement:** A human maintainer directed the scope, clarified the approach with the issue author, reviewed every change.
- **Caveats:**
  - Verification was done via unit tests and an offline demo; no live provider API calls were made.
  - No real per-model context-window values are included — those are left to the provider packages.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -365,6 +365,7 @@ direction LR
             +usingRequestOptions(RequestOptions $requestOptions) self
             +usingSystemInstruction(string $systemInstruction) self
             +usingMaxTokens(int $maxTokens) self
+            +usingTokenCounter(TokenCounterInterface $tokenCounter) self
             +usingTemperature(float $temperature) self
             +usingTopP(float $topP) self
             +usingTopK(int $topK) self
@@ -556,6 +557,7 @@ direction LR
             +usingProvider(string $providerIdOrClassName) self
             +usingSystemInstruction(string $systemInstruction) self
             +usingMaxTokens(int $maxTokens) self
+            +usingTokenCounter(TokenCounterInterface $tokenCounter) self
             +usingTemperature(float $temperature) self
             +usingTopP(float $topP) self
             +usingTopK(int $topK) self
@@ -1103,6 +1105,7 @@ direction LR
             +getName() string
             +getSupportedCapabilities() CapabilityEnum[]
             +getSupportedOptions() SupportedOption[]
+            +getContextWindow() ?int
             +getJsonSchema() array< string, mixed >$
         }
         class ModelRequirements {
--- a/docs/GLOSSARY.md
+++ b/docs/GLOSSARY.md
@@ -5,6 +5,7 @@ This glossary defines common terms relevant for the PHP AI Client and related pr
 * **Agent**: An autonomous system that can perceive its environment, make decisions, and take actions to achieve specific goals, often leveraging AI models.
 * **Candidate Count**: The number of different response options an LLM generates internally.
 * **Capability**: A specific skill, function, or type of task that an AI model can perform, e.g. text generation or image generation.
+* **Context Window**: The total number of tokens (input plus output) a _Model_ can process in a single request. Exposed as model metadata (`ModelMetadata::getContextWindow()`) and used for an advisory pre-flight check. Distinct from the _Max Tokens_ option, which caps only the generated output.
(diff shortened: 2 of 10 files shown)


georgestephanis:
@the-hercules Is this missing an AI Use disclosure in the PR?  (What models were used, how, caveats, etc)

the-hercules:
@georgestephanis Yes, it was missing AI disclosure, it was my first time raising PR on this repo so didn't know that we write a disclosure at the end. Will keep in mind from next time.

georgestephanis:
Thanks!

PR #267, opened 2026-07-29
Fix avoid file_exists() on oversized strings in File detection to prevent PHP warnings

### LINKED ISSUE #258: File DTO: file_exists() called on oversized strings causes PHP warning spam when handling base64 image data
## Description

`File::detectAndProcessFile()` in `src/Files/DTO/File.php` calls `file_exists()` before checking whether the input string is plain base64 data. When a caller passes a large base64-encoded image (e.g. from Google Imagen / Gemini via `bytesBase64Encoded`), PHP emits:

```
PHP Warning: file_exists(): File name is longer than the maximum allowed path length on this platform (4096): /9j/4AAQSkZJRgABAQEBLAEsAAD/...
```

Because the base64 data is embedded verbatim in the warning message, each occurrence adds **~1 MB** to the PHP error log. In production this caused a 5.3 MB error log from only 6–8 image generation calls.

## Root Cause

The detection order in `detectAndProcessFile()` is:

1. URL check → skip
2. Data URI check → skip
3. **`file_exists($file)` called** → PHP warning (string > 4096 chars)
4. Plain base64 regex check → matches correctly

Base64-encoded JPEG data starts with `/9j/` (the base64 representation of the JPEG magic bytes `FF D8 FF`), which looks like an absolute file path on Linux. PHP calls the filesystem with a ~750 KB–1 MB "path" and emits the warning before the code ever reaches the base64 branch.

## Reproduction

```php
// Simulate what ai-provider-for-google does when Google AI returns bytesBase64Encoded
$base64Jpeg = base64_encode(file_get_contents('/path/to/any/image.jpg')); // > 4 KB
$file = new \WordPress\AiClient\Files\DTO\File($base64Jpeg, 'image/jpeg');
// → PHP Warning: file_exists(): File name is longer than the maximum allowed path length on this platform (4096)
```

## Affected code

**File:** `src/Files/DTO/File.php`, line 94 (as of the version bundled with WordPress 6.9)

```php
// Before (bug):
if (file_exists($file) && is_file($file)) {

// After (fix):
if (strlen($file) <= PHP_MAXPATHLEN && file_exists($file) && is_file($file)) {
```

`PHP_MAXPATHLEN` is the PHP constant that reflects the OS limit (typically 4096 on Linux). A string longer than that can never be a valid file path, so short-circuiting the check is semantically correct and has no side effects on legitimate use cases.

## Context

- Observed on a WordPress 6.9 site using the `ai-provider-for-google` plugin v1.1.0
- The plugin passes `$predictionData['bytesBase64Encoded']` directly to `new File(...)`, which is the documented usage (plain base64 with MIME type)
- The `File` class is supposed to handle this case, but the `file_exists()` call causes a noisy warning before it gets there
- Images returned by Google AI contain C2PA metadata, making them even larger and amplifying the log impact

the-hercules:
Closes #258 
## Summary

`File::detectAndProcessFile()` calls `file_exists()` while determining whether an
input string is a local path, before the plain-base64 branch is reached. When the
input is a large base64 payload (e.g. an image returned by a provider via
`bytesBase64Encoded`), the string exceeds the platform's maximum path length and
PHP emits:

> `file_exists(): File name is longer than the maximum allowed path length on this platform (4096): /9j/4AAQSkZJRg...`

The warning message embeds the entire input string, so each occurrence writes
~1 MB to the error log. In practice this produced multi-megabyte error logs from
only a handful of image-generation calls. Base64-encoded JPEG data begins with
`/9j/`, which resembles an absolute path, so the string reaches `file_exists()`
before detection falls through to the base64 handling that processes it correctly.

Functionally the input was always handled correctly — this is a log-noise issue,
not a data-correctness one.

## Change

Guard the filesystem check with a length comparison:

```php
if (strlen($file) <= PHP_MAXPATHLEN && file_exists($file) && is_file($file)) {
```

### AI Disclosure

Claude Opus 4.8 was used for identification and then verification of correctness of the solution.

### DIFF
--- a/src/Files/DTO/File.php
+++ b/src/Files/DTO/File.php
@@ -100,8 +100,10 @@ class File extends AbstractDataTransferObject
             return;
         }
 
-        // Check if it's a local file path (before base64 check)
-        if (file_exists($file) && is_file($file)) {
+        // Check if it's a local file path (before base64 check).
+        // The length guard avoids calling file_exists() on over-length strings (e.g. base64 data),
+        // which would emit a warning containing the entire string.
+        if (strlen($file) <= PHP_MAXPATHLEN && file_exists($file) && is_file($file)) {
             $this->fileType = FileTypeEnum::inline();
             $this->base64Data = $this->convertFileToBase64($file);
             $this->mimeType = $this->determineMimeType($providedMimeType, null, $file);
--- a/tests/unit/Files/DTO/FileTest.php
+++ b/tests/unit/Files/DTO/FileTest.php
@@ -112,6 +112,43 @@ class FileTest extends TestCase
     }
 
     /**
+     * Tests creating a File from base64 data longer than the maximum path length without a PHP warning.
+     *
+     * @return void
+     */
+    public function testCreateFromLargePlainBase64DoesNotEmitWarning(): void
+    {
+        // Base64-encoded JPEG data, longer than PHP_MAXPATHLEN and starting with the "/9j/" prefix.
+        $rawImage = "\xFF\xD8\xFF" . str_repeat("\x00", PHP_MAXPATHLEN);
+        $base64Data = base64_encode($rawImage);
+        $mimeType = 'image/jpeg';
+
+        $this->assertStringStartsWith('/9j/', $base64Data);
+        $this->assertGreaterThan(PHP_MAXPATHLEN, strlen($base64Data));
+
+        $capturedWarning = null;
+        set_error_handler(
+            static function (int $errno, string $errstr) use (&$capturedWarning): bool {
+                $capturedWarning = $errstr;
+
+                return true;
+            },
+            E_WARNING
+        );
+
+        try {
+            $file = new File($base64Data, $mimeType);
+        } finally {
+            restore_error_handler();
+        }
+
+        $this->assertNull($capturedWarning, 'Constructing a File from large base64 data must not emit a warning.');
+        $this->assertEquals(FileTypeEnum::inline(), $file->getFileType());
+        $this->assertEquals($base64Data, $file->getBase64Data());
+        $this->assertEquals($mimeType, $file->getMimeType());
+    }
+
+    /**
      * Tests that plain base64 without MIME type throws exception.
      *
      * @return void


src/Files/DTO/File.php
```diff
@@ -100,8 +100,10 @@ private function detectAndProcessFile(string $file, ?string $providedMimeType):
             return;
         }
 
-        // Check if it's a local file path (before base64 check)
-        if (file_exists($file) && is_file($file)) {
+        // Check if it's a local file path (before base64 check).
+        // The length guard avoids calling file_exists() on over-length strings (e.g. base64 data),
+        // which would emit a warning containing the entire string.
+        if (strlen($file) <= PHP_MAXPATHLEN && file_exists($file) && is_file($file)) {
```
felixarntz, line 106: not sure this is the adequate check - can't we instead perform a check for whether the string is possibly a file path? maybe check if it starts with `/` (since it always needs to be an absolute path in practice)

felixarntz (commented):
@the-hercules Great catch! I think we may be able to find a cleaner solution though.

tyrann0us:
>[…] this is a log-noise issue […]

Not entirely. I came across this because on nginx (Apache untested), with Query Monitor active, image generation fails with:
```
The response is not a valid JSON response.
```

According to Claude Code, this is because Query Monitor takes the full warning text and puts it in an `X-QM-php-errors-error-1` response header, causing a 502 "too big header" error, and thus failing the image generation/insertion. I confirmed it by deactivating Query Monitor. Alternatively, a mu-plugin like this would also work around the issue:
```php
add_filter( 'qm/dispatch/rest', '__return_false' )
// or more specifically
add_filter( 'qm/outputter/headers', fn( $o ) => array_diff_key( $o, [ 'php_errors' => 1 ] ), 999 ); // untested
````

**Comment by the-hercules** [comment]
Hi @felixarntz 
Thanks for the review! I looked at this closely, and a leading `/` check wouldn't actually prevent the reported warning on its own. Base64-encoded JPEGs start with `/9j/`, which already passes a "starts with slash" test.

The warning is triggered purely by the length, the length guard is the part that actually prevents the warning here. I feel this is the standard fix for this specific PHP behavior.

Adding a leading-slash requirement might also unnecessarily drop support for relative local paths.

@tyrann0us Yes it might be more than a log issue at this point.

**Review by Infinite-Null — COMMENTED** [review]
Hi @the-hercules, thank you for the PR 🙌

I was wondering if it would make sense to suppress the warning from the filesystem probe as well?

```php
if (strlen($file) < PHP_MAXPATHLEN && @file_exists($file) && @is_file($file)) {
```

I also think the test should cover the exact `PHP_MAXPATHLEN` boundary and a smaller Base64 payload, so we know both cases are covered.

One small thing to keep in mind: the current test's custom error handler will still capture warnings even when `@file_exists()` is used, so the test may need to account for the suppressed warning as well.

What do you think?

the-hercules:
Hey @Infinite-Null, thanks for the review! 🙌

These are good suggestions, I agree with the points:

- < instead of <= — You're right, PHP_MAXPATHLEN includes the null terminator so strict less-than is the safer boundary. Will fix.
- `@ suppression` — Makes sense as a defensive way, masks error, but can it make debugging harder?
- Boundary + smaller payload tests — Yes, can add cases for the exact PHP_MAXPATHLEN boundary and a shorter base64 string.

the-hercules:
Hey @Infinite-Null, all three are in 🙌

* < → <= was a real off-by-one: PHP warns at >= MAXPATHLEN, so exactly 4096 chars could still be logged.
* @ suppression is necessary, not just defensive. With open_basedir enabled, PHP can log the full input even when it’s below the length limit. The length guard alone therefore didn’t fully address the issue.
* Updated the error-handler check to (error_reporting() & $errno) !== 0, correctly handling @ suppression across PHP 7.4, 8.0, 8.4, and 8.5.

Added boundary and smaller-payload tests. All tests pass across the supported PHP versions.

One caveat: the warning tests only reproduce with open_basedir enabled, so they don’t fail against the old code in the default CI environment. Making them enforce the behavior would require running them in a separate process with open_basedir configured.

PR #269, opened 2026-07-31
Add text extraction (OCR / document parsing) capability

### LINKED ISSUE #268: Add provider-agnostic text extraction (OCR / document parsing) capability
# Proposal: Text Extraction Capability (OCR / Document Parsing)

## Summary

Add a first-class, provider-agnostic **text extraction** capability to the PHP AI Client, so that provider packages can register document-parsing/OCR models the same way they register text generation and image generation models today.

The design is validated against five concrete provider APIs with very different shapes:

| Provider | API style | Sync/Async | Native output | Validation |
|---|---|---|---|---|
| Mistral OCR (`mistral-ocr-*`) | dedicated `POST /v1/ocr` | sync | markdown per page + image bboxes | **Tested** (working PoC connector, live API) |
| LlamaParse (LlamaCloud) | upload → job → poll (`/api/v1/parsing/*`; v2 exists) | async only | markdown/text/structured items per page | **Tested** (working PoC connector, live API) |
| Google Cloud Document AI | processor `:process` / `:batchProcess` | both | `Document` JSON (blocks, anchors, bboxes) | Researched with AI (official docs) |
| AWS Textract | `DetectDocumentText` / `AnalyzeDocument` + Start/Get jobs | both | flat `Blocks[]` graph (bboxes, tables, KV) | Researched with AI (official docs) |
| Qwen2.5-VL / Qwen3-VL / `qwen-vl-ocr` | chat completions with image input | sync | prompt-shaped text (md/HTML/LaTeX/JSON) | Researched with AI (official docs) |

The goal is that each of these can ship as an independent connector package implementing one shared interface, exactly as `ai-provider-for-mistral` does for text generation today.

## Motivation

Document text extraction is one of the most common AI tasks in CMS contexts (ingesting PDFs into posts, indexing uploads for search/RAG, accessibility). Today the SDK cannot model it:

1. `CapabilityEnum` has no extraction-shaped value; `ProviderRegistry::findModelsMetadataForSupport()` therefore can never discover such models.
2. Extraction results (pages, markdown, bounding boxes, extracted images, page counts) do not fit the `GenerativeAiResult` envelope (candidates + finish reasons + messages). Flattening a multi-page structured response into a single model-message candidate loses the structure that is the entire point of dedicated OCR endpoints.
3. The prompt paradigm doesn't fit: the primary input is a *document*, not a message list, and options like page ranges have no home in `PromptBuilder`.

Downstream provider plugins are blocked on this (e.g. `ai-provider-for-mistral` cannot expose `mistral-ocr-latest`); the alternative, provider-specific side APIs outside the SDK, defeats the SDK's core value of capability-agnostic model discovery.

### Why not just send the file to a text generation model?

The SDK can already attach a document to a chat prompt (`inputModalities: [document]`) and ask an LLM to "transcribe this". That path is a complement, not a substitute, because dedicated extraction endpoints are fundamentally more reliable for getting a document's actual text out:

- **Hallucination.** A generative model *produces* text conditioned on the document; nothing constrains its output to be a faithful transcription. On long PDFs especially, LLMs are known to silently skip pages, paraphrase, "fix" numbers and names, or fabricate plausible-looking content for low-quality scans. Extraction pipelines transcribe what is on the page, and several (Google, AWS, optionally Mistral) attach per-word/per-block confidence scores so uncertainty is *reported* instead of papered over.
- **Length limits.** A long PDF quickly exhausts a chat model's context window, and output token limits cap how much text can come back in one completion. Extraction endpoints process documents page by page (Mistral OCR: up to ~1,000 pages per request; Textract/Document AI batch: thousands of pages) with no prompt-window coupling.
- **Determinism and structure.** Extraction returns machine-verifiable structure, page indices, bounding boxes, tables, embedded images, that a free-text completion cannot guarantee and that flattening into a chat message destroys.
- **Cost.** Per-page extraction pricing (e.g. $4/1,000 pages for Mistral OCR) is orders of magnitude cheaper at document scale than paying per input+output token to have an LLM re-emit an entire document.

For ingestion pipelines (search indexing, RAG, archiving, accessibility), where fidelity to the source is the requirement, extraction is the correct tool, and the SDK currently cannot express it. Document QnA via chat remains the right tool for *reasoning about* a document, and the two compose: extract first, then feed clean markdown to a text generation model.

### Completing the RAG pipeline with embedding generation

The SDK gained embedding generation in 1.4.0, but embedding models take text (`list<MessagePart>`), they cannot read a PDF. Text extraction is the missing first stage of that pipeline: without it, a consumer who wants to embed their document library must leave the SDK to get the text out, which reintroduces exactly the per-provider integration work the SDK exists to remove.

With this capability the full document-to-vector flow becomes SDK-native and provider-agnostic:

```php
$result = AiClient::document($pdfUrl)->extractTextResult();

foreach ($result->getPages() as $page) {
    $embedding = AiClient::input($page->getMarkdown())->generateEmbedding();
    // store vector with page number for citation-accurate retrieval
}
```

The extraction result is also a better embedding input than any ad-hoc alternative: per-page (and optionally per-block) segmentation gives natural chunk boundaries that respect document structure instead of arbitrary character offsets; markdown preserves headings and tables that improve embedding quality; and page numbers/bounding boxes carried alongside each chunk enable citation-accurate retrieval, the retrieved passage can point back to the exact page and region of the source document. Layout-aware chunking is precisely what Google's Layout Parser markets itself for in RAG contexts; this proposal makes that pattern available uniformly across providers.

**Precedent:** embedding generation (1.4.0) faced the same mismatch and established the pattern this proposal follows: a new capability value, a standalone model interface, a dedicated result DTO implementing `ResultInterface`, a separate builder (`AiClient::input()` → `EmbeddingBuilder`), a dedicated `ModelRequirements` factory, and its own events. Text extraction is the same kind of "not prompt-shaped" capability.

## Naming

Recommend **`TEXT_EXTRACTION = 'text_extraction'`** over `OCR`:

- "OCR" implies raster input; LlamaParse and Document AI Layout Parser parse born-digital DOCX/PPTX/HTML with no optical step.
- It parallels existing verb-object names (`text_generation`, `image_generation`, `embedding_generation`).
- It leaves room for the same capability to cover future "document parsing" providers without a misleading name.

(Naming is genuinely bikesheddable; `DOCUMENT_TEXT_EXTRACTION` is the verbose alternative. The rest of this proposal uses `TEXT_EXTRACTION`.)

## Design principles (derived from the API survey)

1. **Input is a `File`.** The existing `File` DTO already models the two transport forms every provider accepts in its sync path: remote URL and inline base64 (Mistral `document_url`/data-URI, LlamaParse `source_url`/upload, Google `rawDocument.content`, Textract `Document.Bytes`, Qwen `image_url`/data-URI). Provider-side file references (S3/GCS/pre-uploaded file IDs) are provider implementation details, reachable via `customOptions` in v1.
2. **Normalize to pages of markdown; keep structure optional.** Every provider has a per-page notion and can produce text per page. Markdown is the richest common text format (Mistral and LlamaParse emit it natively; Google/Textract adapters synthesize it from blocks; Qwen is prompted into it). Bounding boxes, tables, and extracted images exist only in some providers, so they are optional fields, and model metadata declares what a model can do via supported options.
3. **Never discard provider fidelity.** The raw decoded provider payload rides along in `additionalData`, so consumers who need Textract `Blocks` relationships or the full Document AI JSON aren't blocked by the normalization.
4. **Sync interface first; async as a follow-up that reuses the existing operations track.** Mistral and Qwen are sync; Google and Textract have sync paths with limits; LlamaParse is async-only but short-lived (an adapter can poll internally). A `TextExtractionOperationModelInterface` mirroring the existing `generateXOperation()` pattern is specified but explicitly deferred (the operations track has no concrete implementation in-tree yet).
5. **Auth stays out of the capability.** Google needs OAuth service accounts and AWS needs SigV4, neither fits `RequestAuthenticationMethod::apiKey()`. That is an orthogonal, pre-existing gap tracked in #237 (provider-owned custom request authentication); this proposal depends on it only for those two connectors, not for the capability itself.

### 1. Capability enum value

`src/Providers/Models/Enums/CapabilityEnum.php`, add one constant plus the magic-method docblock lines (no factory body needed; `AbstractEnum::__callStatic` provides `textExtraction()` and `isTextExtraction()` automatically):

```php
/**
 * Text extraction (OCR / document parsing) capability.
 *
 * @since n.e.x.t
 */
public const TEXT_EXTRACTION = 'text_extraction';
```

```php
 * @method static self textExtraction() Creates an instance for TEXT_EXTRACTION capability.
 * @method bool isTextExtraction() Checks if the capability is TEXT_EXTRACTION.
```

`ProviderRegistry` discovery works with zero further changes, `ModelRequirements::areMetBy()` is capability-generic.

### 2. Model interface

`src/Providers/Models/TextExtraction/Contracts/TextExtractionModelInterface.php` (standalone interface, like all capability contracts; concrete models also implement `ModelInterface` via `AbstractApiBasedModel`):

```php
interface TextExtractionModelInterface
{
    /**
     * Extracts text and structure from a document.
     *
     * Extraction options (page selection, image/bounding-box inclusion,
     * output format) are provided via the model's ModelConfig, consistent
     * with how generation options are handled.
     *
     * @since n.e.x.t
     *
     * @param File $document The document to process (remote URL or inline data).
     * @return TextExtractionResult The structured extraction result.
     */
    public function extractTextResult(File $document): TextExtractionResult;
}
```

Options ride on `ModelConfig` (house style, capability interfaces take only the payload; `setConfig()` carries everything else), which also makes them discoverable/matchable through `SupportedOption` metadata.

### 3. Result DTOs

Following the `EmbeddingResult` precedent: a dedicated result implementing `ResultInterface` (NOT candidate-based), living in `src/Results/DTO/`. All DTOs extend `AbstractDataTransferObject` with `KEY_*` constants, `toArray()`/`fromArray()`/`getJsonSchema()`, and deep `__clone`.

```php
class TextExtractionResult implements ResultInterface
{
    public function getId(): string;
    /** @return list<ExtractedPage> */
    public function getPages(): array;
    public function getPageCount(): int;                 // pages processed (billing-relevant for per-page providers)
    public function getTokenUsage(): TokenUsage;         // zeros for page-priced providers; real for VLM-based extraction
    public function getProviderMetadata(): ProviderMetadata;
    public function getModelMetadata(): ModelMetadata;
    public function getAdditionalData(): array;          // MUST include the raw provider payload under 'raw'
    public function toText(): string;                    // convenience: all pages' markdown joined
    public function toMarkdown(): string;                // alias emphasizing format
}

class ExtractedPage extends AbstractDataTransferObject
{
    public function getPageNumber(): int;                // 1-based, normalized across providers
    public function getMarkdown(): string;               // markdown (may be plain text for text-only providers)
    /** @return list<ExtractedBlock> */
    public function getBlocks(): array;                  // optional; empty when unsupported/not requested
    /** @return list<ExtractedImage> */
    public function getImages(): array;                  // optional; empty when unsupported/not requested
    public function getDimensions(): ?PageDimensions;    // null when the provider doesn't report them
}

class ExtractedBlock extends AbstractDataTransferObject
{
    public function getType(): TextExtractionBlockTypeEnum; // PARAGRAPH | HEADING | TABLE | LIST | IMAGE | OTHER
    public function getText(): string;
    public function getBoundingBox(): ?BoundingBox;
    public function getConfidence(): ?float;             // 0–1; null when the provider doesn't score
}

class ExtractedImage extends AbstractDataTransferObject
{
    public function getId(): string;
    public function getFile(): ?File;                    // inline base64 File when returned, else null
    public function getBoundingBox(): ?BoundingBox;
}

class BoundingBox extends AbstractDataTransferObject
{
    // Normalized coordinates in the 0–1 range, origin top-left.
    // Rationale: Textract is natively normalized; Google provides both;
    // pixel-native providers (Mistral) divide by page dimensions, which
    // they always return. Pixel values are recoverable via PageDimensions.
    public function getLeft(): float;
    public function getTop(): float;
    public function getWidth(): float;
    public function getHeight(): float;
}

class PageDimensions extends AbstractDataTransferObject
{
    public function getWidth(): int;    // pixels
    public function getHeight(): int;   // pixels
    public function getDpi(): ?int;
}
```

`TokenUsage` note: `ResultInterface` requires it; page-priced providers return `new TokenUsage(0, 0, 0)` and the meaningful unit is `getPageCount()`. VLM-based connectors (Qwen) populate real token counts. Both are honest; neither overloads the other's field.

### 4. ModelConfig / OptionEnum additions

`OptionEnum` reflects `ModelConfig::KEY_*` constants automatically, so adding config keys is the entire change. Proposed minimal set:

```php
ModelConfig::KEY_EXTRACTION_PAGES          = 'extractionPages';         // list<int>, 1-based page selection
ModelConfig::KEY_EXTRACTION_INCLUDE_IMAGES = 'extractionIncludeImages'; // bool
ModelConfig::KEY_EXTRACTION_INCLUDE_BLOCKS = 'extractionIncludeBlocks'; // bool (bounding boxes / layout)
```

Deliberately reused rather than duplicated:

- **`outputMimeType`**, `text/markdown` (default) vs `text/plain`; a connector for a provider with HTML table output could advertise `text/html`.
- **`outputSchema`**, schema-driven structured extraction (Mistral `document_annotation_format` with JSON schema, Textract QUERIES/FORMS mapping, Document AI custom extractors, `qwen-vl-ocr` `key_information_extraction.result_schema`). A model advertising `SupportedOption(OptionEnum::outputSchema())` under this capability means "can return schema-shaped extraction", the result's `additionalData['structuredData']` carries it in v1.
- **`inputModalities`**, declares what documents a model accepts: `[document]`, `[image]`, or both. This is how the Qwen connector honestly declares images-only (no native PDF input; rasterization is out of scope), while Mistral/LlamaParse/Google declare `[document, image]`.
- **`customOptions`**, escape hatch for provider-specific knobs (Mistral `image_min_size`, LlamaParse `tier`, Textract `FeatureTypes`, Google processor selection, DashScope `ocr_options.task`).

### 5. Fluent API

Text extraction gets its own builder, exactly as embeddings did (per ARCHITECTURE.md's rationale: prompt-oriented parameters don't apply).

`AiClient` additions:

```php
public static function document($document = null, ?ProviderRegistry $registry = null): TextExtractionBuilder;

public static function extractTextResult($document, $modelOrConfig = null, ?ProviderRegistry $registry = null): TextExtractionResult;
public static function extractText($document, $modelOrConfig = null, ?ProviderRegistry $registry = null): string;
```

`src/Builders/TextExtractionBuilder.php`:

```php
$result = AiClient::document('https://example.com/report.pdf')
    ->fromPages([1, 2, 3])
    ->includingImages()
    ->includingBlocks()
    ->usingProvider('mistral')            // via the shared ModelResolutionTrait
    ->extractTextResult();                // TextExtractionResult

$markdown = AiClient::document($file)->extractText(); // shorthand → string
```

The builder:

- accepts `File|string` (URL, data URI, or local path, `File` already normalizes all three);
- uses `ModelResolutionTrait` (`usingModel()`, `usingProvider()`, `usingModelPreference()`, …) unchanged;
- builds requirements via a new `ModelRequirements::fromExtractionData(File $document, ModelConfig $config)`: capability `textExtraction()`, `RequiredOption(inputModalities, [document|image])` chosen from the file's MIME type, plus required options for any set config keys;
- validates the resolved model `instanceof TextExtractionModelInterface` and throws the standard `'Model "%s" does not support text extraction.'` otherwise;
- dispatches `BeforeExtractTextEvent` / `AfterExtractTextEvent` (mirroring the embedding events);
- exposes `isSupported(): bool` for feature detection.

### 6. Abstract base class

`src/Providers/ApiBasedImplementation/AbstractApiBasedTextExtractionModel.php`, mirrors the existing OpenAI-compatible generation bases: holds metadata/config, template method flow `prepareRequest(File): Request` → authenticate → send → `throwIfNotSuccessful()` → `parseResponseToTextExtractionResult(Response): TextExtractionResult`. Concrete connectors override the two abstract ends.

Unlike text generation there is no dominant wire format to ship a shared "compatible" implementation for, each connector implements its own request/parse pair. That is fine; the shared value is in the result normalization and discovery, not the HTTP shape.

### 7. Async operations (specified, deferred)

`TextExtractionOperationModelInterface::extractTextOperation(File $document): TextExtractionOperation` plus a `TextExtractionOperation` DTO (`id`, `OperationStateEnum`, `?TextExtractionResult`), mirroring the existing `generateXOperation()` contracts and `GenerativeAiOperation`. Deferred because the operations track has no concrete provider-side implementation or builder exposure anywhere in the SDK yet; text extraction shouldn't be the pioneer.

Until then, async-only providers poll internally behind the sync interface with a configurable timeout (LlamaParse jobs on typical documents complete in seconds; this is what its own SDKs do). Connectors SHOULD expose the timeout via `customOptions` and throw a clear `RuntimeException` on expiry.

## Connector mapping

How each target provider implements `TextExtractionModelInterface`:

| Concern | Mistral OCR | LlamaParse | Google Document AI | AWS Textract | Qwen-VL / qwen-vl-ocr |
|---|---|---|---|---|---|
| Request | `POST /v1/ocr` with `document_url`/data-URI | upload or `source_url` → `POST /api/v2/parse` → poll `GET /api/v2/parse/{id}` | `processors/{id}:process` with base64 `rawDocument` | `AnalyzeDocument`/`DetectDocumentText` with `Document.Bytes` | chat completion, image content + extraction prompt / `ocr_options` |
| Auth | Bearer key (supported today) | Bearer key (supported today) | OAuth service account, **needs #237** | SigV4, **needs #237** | Bearer key (DashScope) or none (self-hosted vLLM) |
| Pages → `ExtractedPage` | native `pages[].markdown` | v2 `expand=markdown,items` per page | synthesize markdown from `pages[].blocks/paragraphs` + `textAnchor` | synthesize from `LINE`/`LAYOUT_*` blocks grouped by `Page` | one page per input image; text as returned |
| Blocks/bboxes | `include_blocks` paragraph bboxes (pixel → normalize by `dimensions`) | layout `items` | `boundingPoly` (already normalized) | `Geometry.BoundingBox` (already normalized) | only `qwen-vl-ocr` `advanced_recognition` (rotated rects → axis-aligned approximation, or omit) |
| Images | `include_image_base64` → `ExtractedImage` | image download URLs → fetch → inline `File` | none (empty) | none (empty) | none (empty) |
| Model discovery | `GET /v1/models` (`mistral-ocr-*`) | fixed tier list, hardcoded metadata | list processors as "models" | single hardcoded model entry (+ adapters later) | hardcoded / `GET /v1/models` on self-hosted |
| Usage | `usage_info.pages_processed` → `getPageCount()` | page count from result | `pages` length | `DocumentMetadata.Pages` | real `TokenUsage` from completion |

The Qwen connector is the important stress test: it proves the capability is about the *contract*, not the transport, a chat-completions-backed extractor and a dedicated-endpoint extractor are interchangeable to consumers, which is precisely the provider-agnostic philosophy. It also demonstrates why `inputModalities` must be declarable per model (`[image]` only).

## Proof of concept

The proposed API has been implemented end to end and validated against two live providers with intentionally different API shapes:

- **SDK capability**, [`saarnilauri/php-ai-client` @ `feature/text-extraction-poc`](https://github.com/saarnilauri/php-ai-client/tree/feature/text-extraction-poc) (fork of this repo): `CapabilityEnum::TEXT_EXTRACTION`, `TextExtractionModelInterface`, the result DTOs (`TextExtractionResult`, `ExtractedPage`, `ExtractedImage`, `BoundingBox`, `PageDimensions`), `ModelRequirements::fromExtractionData()`, `TextExtractionBuilder`, and the `AiClient::document()` / `extractTextResult()` / `extractText()` entry points, with unit tests.
- **Sync connector**, [`saarnilauri/ai-provider-for-mistral` @ `feature/text-extraction-poc`](https://github.com/saarnilauri/ai-provider-for-mistral/tree/feature/text-extraction-poc): Mistral's dedicated `POST /v1/ocr` endpoint. Covers URL and inline (data URI) input, page selection and other options via custom-options passthrough, pixel→normalized bounding box conversion, and embedded image decoding.
- **Async connector**, [`saarnilauri/ai-provider-for-llamaparse`](https://github.com/saarnilauri/ai-provider-for-llamaparse) (`main`): LlamaParse's job-based API (upload → poll → fetch result) hidden behind the synchronous interface with internal polling, multipart file upload, and hardcoded model metadata (LlamaParse has no model-list endpoint, parsing modes are exposed as model entries).

Both connectors pass the same integration test suite against the real APIs, capability discovery, multi-page remote PDF extraction, local PDF upload, and local image extraction, with only the provider ID differing between the suites, which demonstrates the provider-agnostic claim in practice.

### Running the PoC locally

The connector packages consume the SDK branch via a Composer **path repository** with a temporary version pin:

```json
"require-dev": { "wordpress/php-ai-client": "dev-feature/text-extraction-poc" },
"repositories": [ { "type": "path", "url": "../php-ai-client", "options": { "symlink": true } } ]
```

This means the three checkouts must be **sibling directories** (e.g. `dev/php-ai-client`, `dev/ai-provider-for-mistral`, `dev/ai-provider-for-llamaparse`), with the SDK checkout on the `feature/text-extraction-poc` branch. The pin is PoC-only and reverts to a released version constraint once the capability ships in the SDK.

Integration tests require `MISTRAL_API_KEY` / `LLAMAPARSE_API_KEY` in each connector's `.env` and are run with `composer test:integration`. Extracted markdown and embedded images are written to `tests/integration/extractions/` for inspection.

### Findings from the PoC

- The embedding-generation pattern (separate builder + dedicated result type) transferred cleanly; no changes to `ProviderRegistry`, `ModelResolver`, or option matching were needed beyond the new requirements factory.
- MIME inference needs an explicit escape hatch: extensionless document URLs (e.g. `https://arxiv.org/pdf/1805.04770`) cannot be typed automatically, which is why `withDocument()` accepts an optional `$mimeType`.
- Providers differ in how strictly they validate declared vs. actual file content: Mistral silently accepted an AVIF image mislabeled as PNG, while LlamaParse hard-failed the parsing job. Content sniffing (magic bytes) in the `File` DTO would harden this for all capabilities.
- Async-behind-sync polling is workable for LlamaParse-scale jobs (seconds to ~1 minute for a 15-page PDF), supporting the decision to defer the operation interface.

## Out of scope (v1)

- **PDF rasterization** for image-only models (Qwen), consumer or connector concern; the model honestly declares `inputModalities: [image]`.
- **First-class table / key-value DTOs**, tables arrive as markdown/HTML inside page text (all providers can deliver that); a structured `ExtractedTable` normalization across Textract CELL graphs, Document AI `pages.tables`, and LlamaParse items is a follow-up once two connectors ship.
- **First-class structured annotation results**, reuse `outputSchema` + `additionalData['structuredData']` in v1; promote to a typed result field later.
- **Async operation wiring** (specified above, deferred).
- **Streaming**, no target provider streams extraction results.
- **Provider-side storage inputs** (S3/GCS URIs, pre-uploaded file IDs), via `customOptions` in v1.
- **Cost estimation**, pricing units diverge irreconcilably (per page vs per credit vs per token).

## Rollout plan

1. **SDK core** (this issue): `CapabilityEnum::TEXT_EXTRACTION`, `TextExtractionModelInterface`, result DTOs, `ModelConfig`/`OptionEnum` keys, `ModelRequirements::fromExtractionData()`, `TextExtractionBuilder`, `AiClient::document()/extractTextResult()/extractText()`, events, `AbstractApiBasedTextExtractionModel`, unit tests (DTO `fromArray(toArray($x))` roundtrips, builder validation, requirements matching, JSON schemas).
2. **Reference connector**: `ai-provider-for-mistral` registers `mistral-ocr-latest` (simplest API: sync, bearer auth, markdown-native, list-models discovery), proves the contract end to end.
3. **Second connector, different shape**: LlamaParse (async-only, no model listing), validates internal polling and hardcoded metadata; then Qwen (chat-transport), validates the VLM path.
4. **Auth-dependent connectors**: Google Document AI and AWS Textract, gated on #237.
5. **Follow-ups**: async operations, table normalization, typed structured-annotation results.

All additions use `@since n.e.x.t`, PHP 7.4-compatible code, and are purely additive (no BC breaks).

## Open questions

1. **Capability name**: `text_extraction` (recommended) vs `ocr` vs `document_text_extraction`?
2. **Builder entry point name**: `AiClient::document()` (recommended, parallels `AiClient::input()`) vs `AiClient::extract()`?
3. Should `ExtractedPage::getMarkdown()` be named `getText()` with format governed by `outputMimeType`, to avoid baking a format into the API name?
4. Should `pageCount` live on `TextExtractionResult` (proposed) or inside `TokenUsage` as a new nullable field usable by other per-unit-priced capabilities?
5. Multi-document batching in one call (Document AI batch, Textract async), out of scope, or should the interface accept `list<File>` from day one? (Proposed: single `File`; batching via multiple calls.)

## References

- Mistral OCR: https://docs.mistral.ai/capabilities/document_ai/basic_ocr and https://docs.mistral.ai/api/endpoint/ocr, OCR 4 model card: https://docs.mistral.ai/models/model-cards/ocr-4-0
- LlamaParse v2 API: https://docs.cloud.llamaindex.ai/ (parse endpoints `/api/v2/parse`)
- Google Cloud Document AI: https://cloud.google.com/document-ai/docs (processors, `Document` format)
- AWS Textract: https://docs.aws.amazon.com/textract/ (`AnalyzeDocument`, Blocks)
- Qwen VL / DashScope: https://www.alibabacloud.com/help/en/model-studio/ (`qwen-vl-ocr`, compatible-mode chat completions)
- Related issues: #150 (speech-to-text, the same "not GenerativeAiResult-shaped" family), #90 (TTS implementation), #165 (document chunks in chat, complementary: chat QnA loses page structure), #160 (multimodal output models), #237 (custom request authentication, prerequisite for Google/AWS connectors), #226 (exposing model capabilities)
- Embedding capability (1.4.0) as the structural template: `EmbeddingResult`, `EmbeddingBuilder`, `ModelRequirements::fromEmbeddingData()`
- Downstream: https://github.com/saarnilauri/ai-provider-for-mistral (blocked on this to expose `mistral-ocr-latest`)

## Use of AI Tools

This proposal and the proof-of-concept implementations were drafted with the assistance of Claude Code (Anthropic), used for researching the provider APIs, analyzing the SDK architecture, and writing code and prose. All work was done with a human in the loop: the design direction, scope decisions, and API trade-offs were made or reviewed by the author, and the proof of concept was verified by the author against the live Mistral and LlamaParse APIs (including real integration test runs and inspection of the extracted output).

saarnilauri:
> **Description updated after review.** Prose whose meaning is now superseded is ~~struck through~~ and followed by its replacement, so the changes are easy to spot; the Scope and model-resolution sections are additions. The point-by-point response to review feedback is in a separate comment below. This will be tidied into final form before merge.

~~Adds a provider-agnostic text extraction capability, following the structural pattern established by embedding generation in 1.4.0 (dedicated builder, result type, and requirements factory):~~

Adds a provider-agnostic text extraction capability, following the structural pattern established by embedding generation in 1.4.0 (dedicated builder, result type, and requirements factory). Note that #274 has since moved embeddings off resolver-based model selection; this PR deliberately does not follow it there — see [Model resolution vs. #274](#model-resolution-vs-274) below.

- `CapabilityEnum::TEXT_EXTRACTION` with magic accessors.
- `TextExtractionModelInterface::extractTextResult(File): TextExtractionResult`.
- Result DTOs: `TextExtractionResult` (`ResultInterface`, non-candidate-based), `ExtractedPage` (1-based page numbers, markdown content), `ExtractedImage`, `BoundingBox` (normalized 0-1 coordinates), `PageDimensions`.
- ~~`ModelRequirements::fromExtractionData()` mapping the document's MIME type to a document or image input-modality requirement.~~<br>`ModelRequirements::fromExtractionData()` deriving the input-modality requirement from the document's MIME type, and rejecting types that carry no readable text (audio, video, unrecognized) instead of mapping them to `document`.
- `TextExtractionBuilder` with `withDocument($document, $mimeType)` plus `AiClient::document()` / `extractTextResult()` / `extractText()` entry points.
- `BeforeExtractTextEvent` / `AfterExtractTextEvent`, dispatched through the shared event dispatcher forwarded by `AiClient::document()`.

Validated end to end by two downstream provider PoCs with intentionally different API shapes: `ai-provider-for-mistral` (synchronous dedicated OCR endpoint) and `ai-provider-for-llamaparse` (async job-based parsing with internal polling).

For more details see the issue that this PR closes #268

## Scope

This lands step 1 of the rollout plan in #268, minus the pieces that are better designed against shipped connectors. Explicitly **deferred**, not overlooked:

| Deferred | Why |
|---|---|
| `ExtractedBlock` / `TextExtractionBlockTypeEnum` | Block-level normalization across Mistral paragraph boxes, Document AI `blocks`, and Textract `Blocks` graphs needs two shipped connectors to design against. Bounding boxes and page dimensions are in, so nothing here is blocked. |
| `ModelConfig::KEY_EXTRACTION_PAGES` / `_INCLUDE_IMAGES` / `_INCLUDE_BLOCKS`, and the `fromPages()` / `includingImages()` / `includingBlocks()` fluent methods | Page selection and image inclusion travel via `customOptions` for now. Promoting them to first-class config keys means committing to their semantics across providers whose page-range and image flags differ; worth doing once more than one connector is real. |
| `AbstractApiBasedTextExtractionModel` | There is no dominant wire format to share, so the base class would only hold the metadata/config plumbing `AbstractApiBasedModel` already provides. |
| Async `TextExtractionOperationModelInterface` | As stated in #268: the operations track has no concrete in-tree implementation, and text extraction should not pioneer it. Async-only providers poll internally behind the sync interface. |

Consequence worth calling out: the `->fromPages([1, 2, 3])->includingImages()` example in #268 is not available yet. The equivalent today is `usingModelConfig()` with custom options, which is what both PoC connectors use.

## Model resolution vs. #274

#274 landed after this branch was opened and made an explicit model **required** for embeddings, dropping resolver-based discovery. This PR deliberately keeps the resolver for extraction, matching text and image generation rather than embeddings.

The reason #274 exists is comparability: embedding vectors from different models occupy different spaces, so a silently chosen model yields results that are wrong in a way the caller cannot detect. Text extraction has no equivalent property — markdown from Mistral OCR and markdown from LlamaParse are interchangeable to the consumer, which is the whole point of normalizing to pages of markdown. Callers who care which provider runs the job still say so with `usingProvider()` or `usingModel()`.

Happy to align with #274 instead if maintainers prefer uniformity across non-prompt capabilities, but it reads as a rationale specific to embeddings rather than a house rule.

## Use of AI Tools

~~This implementations was drafted with the assistance of Claude Code (Anthropic), used for researching the provider APIs, analyzing the SDK architecture, and writing code.~~

This implementation was drafted with the assistance of Claude Code (Anthropic), used for researching the provider APIs, analyzing the SDK architecture, writing code, and applying the fixes from code review. All work was done with a human in the loop: the design direction, scope decisions, and API trade-offs were made or reviewed by the author, and the proof of concept was verified by the author against the live Mistral and LlamaParse APIs (including real integration test runs and inspection of the extracted output).

### DIFF
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -294,7 +294,8 @@ class AiClient
     {
         return new TextExtractionBuilder(
             $registry ?? self::defaultRegistry(),
-            $document
+            $document,
+            self::$eventDispatcher
         );
     }
 
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -8,6 +8,7 @@ use Psr\EventDispatcher\EventDispatcherInterface;
 use Psr\SimpleCache\CacheInterface;
 use WordPress\AiClient\Builders\EmbeddingBuilder;
 use WordPress\AiClient\Builders\PromptBuilder;
+use WordPress\AiClient\Builders\TextExtractionBuilder;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
 use WordPress\AiClient\Providers\Contracts\ProviderAvailabilityInterface;
@@ -19,6 +20,7 @@ use WordPress\AiClient\Providers\ProviderRegistry;
 use WordPress\AiClient\Results\DTO\Embedding;
 use WordPress\AiClient\Results\DTO\EmbeddingResult;
 use WordPress\AiClient\Results\DTO\GenerativeAiResult;
+use WordPress\AiClient\Results\DTO\TextExtractionResult;
 
 /**
  * Main AI Client class providing both fluent and traditional APIs for AI operations.
@@ -84,6 +86,7 @@ use WordPress\AiClient\Results\DTO\GenerativeAiResult;
  * @phpstan-import-type Prompt from PromptBuilder
  * @phpstan-import-type EmbeddingInput from EmbeddingBuilder
  * @phpstan-import-type ProviderModelTuple from ModelResolver
+ * @phpstan-import-type DocumentInput from TextExtractionBuilder
  *
  * phpcs:ignore Generic.Files.LineLength.TooLong
  */
@@ -268,6 +271,30 @@ class AiClient
             $registry ?? self::defaultRegistry(),
             $input,
             self::$eventDispatcher
+        );
+    }
+
+    /**
+     * Creates a text extraction builder for fluent document text extraction (OCR / document parsing).
+     *
+     * Chain extractTextResult() to get the structured, per-page result, or extractText() to get
+     * the extracted content as a single markdown string.
+     *
+     * @since n.e.x.t
+     *
+     * @param DocumentInput|null $document Optional document to extract text from: a File instance,
+     *                                     or a string URL, data URI, base64 data, or local file path.
+     *                                     When the MIME type cannot be inferred from the input (e.g.
+     *                                     an extensionless URL), pass a File instance with an explicit
+     *                                     MIME type, or use the builder's withDocument($document, $mimeType).
+     * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
+     * @return TextExtractionBuilder The text extraction builder instance.
+     */
+    public static function document($document = null, ?ProviderRegistry $registry = null): TextExtractionBuilder
+    {
+        return new TextExtractionBuilder(
+            $registry ?? self::defaultRegistry(),
+            $document
         );
     }
 
@@ -493,6 +520,48 @@ class AiClient
     }
 
     /**
+     * Extracts text from a document using the traditional API approach.
+     *
+     * @since n.e.x.t
+     *
+     * @param DocumentInput $document The document to extract text from.
+     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
+     *                                                        or model configuration for auto-discovery,
+     *                                                        or null for defaults.
+     * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
+     * @return TextExtractionResult The structured extraction result.
+     */
+    public static function extractTextResult(
+        $document,
+        $modelOrConfig = null,
+        ?ProviderRegistry $registry = null
+    ): TextExtractionResult {
+        self::validateModelOrConfigParameter($modelOrConfig);
+        return self::applyModelOrConfig(self::document($document, $registry), $modelOrConfig)
+            ->extractTextResult();
+    }
+
+    /**
+     * Extracts text from a document and returns it as a single markdown string.
+     *
+     * @since n.e.x.t
+     *
+     * @param DocumentInput $document The document to extract text from.
+     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
+     *                                                        or model configuration for auto-discovery,
+     *                                                        or null for defaults.
+     * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
+     * @return string The extracted content, pages joined in order.
+     */
+    public static function extractText(
+        $document,
+        $modelOrConfig = null,
+        ?ProviderRegistry $registry = null
+    ): string {
+        return self::extractTextResult($document, $modelOrConfig, $registry)->toMarkdown();
+    }
+
+    /**
      * Creates a new message builder for fluent API usage.
      *
      * This method will be implemented once MessageBuilder is available.
@@ -604,7 +673,7 @@ class AiClient
      * Works with any builder that exposes the shared model resolution methods
      * (see {@see \WordPress\AiClient\Builders\Traits\ModelResolutionTrait}).
      *
-     * @template T of PromptBuilder
+     * @template T of PromptBuilder|TextExtractionBuilder
      *
      * @param T $builder The builder to configure.
      * @param ModelInterface|ModelConfig|null $modelOrConfig Specific model, model configuration,
--- a/src/Builders/TextExtractionBuilder.php
+++ b/src/Builders/TextExtractionBuilder.php
@@ -1,0 +1,169 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Builders;
+
+use WordPress\AiClient\Builders\Traits\ModelResolutionTrait;
+use WordPress\AiClient\Common\Exception\InvalidArgumentException;
+use WordPress\AiClient\Common\Exception\RuntimeException;
+use WordPress\AiClient\Files\DTO\File;
+use WordPress\AiClient\Providers\ModelResolver;
+use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
+use WordPress\AiClient\Providers\Models\DTO\ModelRequirements;
+use WordPress\AiClient\Providers\Models\TextExtraction\Contracts\TextExtractionModelInterface;
+use WordPress\AiClient\Providers\ProviderRegistry;
+use WordPress\AiClient\Results\DTO\TextExtractionResult;
+
+/**
+ * Fluent builder for extracting text from documents (OCR / document parsing).
+ *
+ * Text extraction transforms a document into structured, per-page content rather than generating
+ * a conversational response. Model selection and configuration are shared with
+ * {@see PromptBuilder} via the {@see ModelResolutionTrait}.
+ *
+ * @since n.e.x.t
+ *
+ * @phpstan-type DocumentInput string|File
+ */
+class TextExtractionBuilder
+{
+    use ModelResolutionTrait;
+
+    /**
+     * @var File|null The document to extract text from.
+     */
+    protected ?File $document = null;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param ProviderRegistry $registry The provider registry for finding suitable models.
+     * @param DocumentInput|null $document Optional initial document to extract text from.
+     */
+    public function __construct(ProviderRegistry $registry, $document = null)
+    {
+        $this->modelConfig = new ModelConfig();
+        $this->modelResolver = new ModelResolver($registry);
+
+        if ($document !== null) {
+            $this->withDocument($document);
+        }
+    }
+
+    /**
+     * Creates a deep clone of this builder.
+     *
+     * Clones the document and model configuration. Service objects are intentionally NOT cloned
+     * as they are shared dependencies.
+     *
+     * @since n.e.x.t
+     */
+    public function __clone()
+    {
+        if ($this->document !== null) {
+            $this->document = clone $this->document;
+        }
+
+        $this->modelConfig = clone $this->modelConfig;
+        $this->modelResolver = clone $this->modelResolver;
+    }
+
+    /**
+     * Sets the document to extract text from.
+     *
+     * @since n.e.x.t
+     *
+     * @param DocumentInput $document The document: a File instance, or a string URL,
+     *                                data URI, base64 data, or local file path.
+     * @param string|null $mimeType Optional MIME type of the document. Required when it
+     *                              cannot be inferred from the input (e.g. an extensionless
+     *                              URL such as `https://arxiv.org/pdf/1805.04770`). Ignored
+     *                              when a File instance is given.
+     * @return self
+     * @throws InvalidArgumentException If the document input is invalid.
+     */
+    public function withDocument($document, ?string $mimeType = null): self
+    {
+        if (is_string($document)) {
+            if (trim($document) === '') {
+                throw new InvalidArgumentException('Cannot create a document from an empty string.');
+            }
+            $document = new File($document, $mimeType);
+        }
+
+        if (!$document instanceof File) {
+            throw new InvalidArgumentException('Document must be a File instance or a string.');
+        }
+
+        $this->document = $document;
+
+        return $this;
+    }
+
+    /**
+     * Checks whether the current document and configuration are supported by an available model.
+     *
+     * @since n.e.x.t
+     *
+     * @return bool True if a suitable text extraction model is available.
+     */
+    public function isSupported(): bool
+    {
+        if ($this->document === null) {
+            return false;
+        }
+
+        $requirements = ModelRequirements::fromExtractionData($this->document, $this->modelConfig);
+
+        return $this->modelResolver->isSupported($requirements);
+    }
+
+    /**
+     * Extracts text from the configured document.
+     *
+     * @since n.e.x.t
+     *
+     * @return TextExtractionResult The structured extraction result.
+     * @throws InvalidArgumentException If no document is configured or model validation fails.
+     * @throws RuntimeException If the resolved model doesn't support text extraction.
+     */
+    public function extractTextResult(): TextExtractionResult
+    {
+        if ($this->document === null) {
+            throw new InvalidArgumentException(
+                'Cannot extract text without a document. Add one using withDocument().'
+            );
+        }
+
+        $requirements = ModelRequirements::fromExtractionData($this->document, $this->modelConfig);
+        $model = $this->modelResolver->resolve($requirements, $this->modelConfig);
+
+        if (!$model instanceof TextExtractionModelInterface) {
+            throw new RuntimeException(
+                sprintf(
+                    'Model "%s" does not support text extraction.',
+                    $model->metadata()->getId()
+                )
+            );
+        }
+
+        return $model->extractTextResult($this->document);
+    }
+
+    /**
+     * Extracts text from the configured document and returns it as a single markdown string.
+     *
+     * @since n.e.x.t
+     *
+     * @return string The extracted content, pages joined in order.
+     * @throws InvalidArgumentException If no document is configured or model validation fails.
+     * @throws RuntimeException If the resolved model doesn't support text extraction.
+     */
+    public function extractText(): string
+    {
+        return $this->extractTextResult()->toMarkdown();
+    }
+}
--- a/src/Builders/TextExtractionBuilder.php
+++ b/src/Builders/TextExtractionBuilder.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace WordPress\AiClient\Builders;
 
+use Psr\EventDispatcher\EventDispatcherInterface;
 use WordPress\AiClient\Builders\Traits\ModelResolutionTrait;
 use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
+use WordPress\AiClient\Events\AfterExtractTextEvent;
+use WordPress\AiClient\Events\BeforeExtractTextEvent;
 use WordPress\AiClient\Files\DTO\File;
 use WordPress\AiClient\Providers\ModelResolver;
 use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
 use WordPress\AiClient\Providers\Models\DTO\ModelRequirements;
+use WordPress\AiClient\Providers\Models\Enums\CapabilityEnum;
 use WordPress\AiClient\Providers\Models\TextExtraction\Contracts\TextExtractionModelInterface;
 use WordPress\AiClient\Providers\ProviderRegistry;
 use WordPress\AiClient\Results\DTO\TextExtractionResult;
@@ -36,17 +40,27 @@ class TextExtractionBuilder
     protected ?File $document = null;
 
     /**
+     * @var EventDispatcherInterface|null The event dispatcher for extraction lifecycle events.
+     */
+    private ?EventDispatcherInterface $eventDispatcher;
+
+    /**
      * Constructor.
      *
      * @since n.e.x.t
      *
      * @param ProviderRegistry $registry The provider registry for finding suitable models.
      * @param DocumentInput|null $document Optional initial document to extract text from.
+     * @param EventDispatcherInterface|null $eventDispatcher Optional event dispatcher for lifecycle events.
      */
-    public function __construct(ProviderRegistry $registry, $document = null)
-    {
+    public function __construct(
+        ProviderRegistry $registry,
+        $document = null,
+        ?EventDispatcherInterface $eventDispatcher = null
+    ) {
         $this->modelConfig = new ModelConfig();
         $this->modelResolver = new ModelResolver($registry);
+        $this->eventDispatcher = $eventDispatcher;
 
         if ($document !== null) {
             $this->withDocument($document);
@@ -56,8 +70,8 @@ class TextExtractionBuilder
     /**
      * Creates a deep clone of this builder.
      *
-     * Clones the document and model configuration. Service objects are intentionally NOT cloned
-     * as they are shared dependencies.
+     * Clones the document and model configuration. Service objects (resolver, event dispatcher)
+     * are intentionally NOT cloned as they are shared dependencies.
      *
      * @since n.e.x.t
      */
@@ -83,20 +97,23 @@ class TextExtractionBuilder
      *                              URL such as `https://arxiv.org/pdf/1805.04770`). Ignored
      *                              when a File instance is given.
      * @return self
-     * @throws InvalidArgumentException If the document input is invalid.
+     * @throws InvalidArgumentException If the document input is invalid, or if its MIME type is
+     *                                  not supported for text extraction.
      */
     public function withDocument($document, ?string $mimeType = null): self
     {
         if (is_string($document)) {
-            if (trim($document) === '') {
-                throw new InvalidArgumentException('Cannot create a document from an empty string.');
-            }
             $document = new File($document, $mimeType);
         }
 
         if (!$document instanceof File) {
             throw new InvalidArgumentException('Document must be a File instance or a string.');
         }
+
+        // Reject unsupported MIME types here, where the file is still the caller's own input, so
+        // that the error names the offending type instead of surfacing later as a resolution or
+        // provider failure.
+        ModelRequirements::extractionInputModality($document);
 
         $this->document = $document;
 
@@ -132,13 +149,16 @@ class TextExtractionBuilder
      */
     public function extractTextResult(): TextExtractionResult
     {
-        if ($this->document === null) {
+        $document = $this->document;
+
+        if ($document === null) {
             throw new InvalidArgumentException(
                 'Cannot extract text without a document. Add one using withDocument().'
             );
         }
 
-        $requirements = ModelRequirements::fromExtractionData($this->document, $this->modelConfig);
+        $capability = CapabilityEnum::textExtraction();
+        $requirements = ModelRequirements::fromExtractionData($document, $this->modelConfig);
         $model = $this->modelResolver->resolve($requirements, $this->modelConfig);
 
         if (!$model instanceof TextExtractionModelInterface) {
@@ -150,7 +170,13 @@ class TextExtractionBuilder
             );
         }
 
-        return $model->extractTextResult($this->document);
+        $this->dispatchEvent(new BeforeExtractTextEvent($document, $model, $capability));
+
+        $result = $model->extractTextResult($document);
+
+        $this->dispatchEvent(new AfterExtractTextEvent($document, $model, $capability, $result));
+
+        return $result;
     }
 
     /**
@@ -165,5 +191,20 @@ class TextExtractionBuilder
     public function extractText(): string
     {
         return $this->extractTextResult()->toMarkdown();
+    }
+
+    /**
+     * Dispatches an event if an event dispatcher is registered.
+     *
+     * @since n.e.x.t
+     *
+     * @param object $event The event to dispatch.
+     * @return void
+     */
+    private function dispatchEvent(object $event): void
+    {
+        if ($this->eventDispatcher !== null) {
+            $this->eventDispatcher->dispatch($event);
+        }
     }
 }
--- a/src/Events/AfterExtractTextEvent.php
+++ b/src/Events/AfterExtractTextEvent.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WordPress\AiClient\Events;
+
+use WordPress\AiClient\Files\DTO\File;
+use WordPress\AiClient\Providers\Models\Contracts\ModelInterface;
+use WordPress\AiClient\Providers\Models\Enums\CapabilityEnum;
+use WordPress\AiClient\Results\DTO\TextExtractionResult;
+
+/**
+ * Event dispatched after text has been extracted from a document.
+ *
+ * @since n.e.x.t
+ */
+class AfterExtractTextEvent
+{
+    /**
+     * @var File The document that was sent to the model.
+     */
+    private File $document;
+
+    /**
+     * @var ModelInterface The model that extracted text.
+     */
+    private ModelInterface $model;
+
+    /**
+     * @var CapabilityEnum The capability that was used for extraction.
+     */
+    private CapabilityEnum $capability;
+
+    /**
+     * @var TextExtractionResult The result from the model.
+     */
+    private TextExtractionResult $result;
+
+    /**
+     * Constructor.
+     *
+     * @since n.e.x.t
+     *
+     * @param File                 $document The document that was sent to the model.
+     * @param ModelInterface       $model The model that extracted text.
+     * @param CapabilityEnum       $capability The capability that was used for extraction.
+     * @param TextExtractionResult $result The result from the model.
+     */
+    public function __construct(
+        File $document,
+        ModelInterface $model,
+        CapabilityEnum $capability,
+        TextExtractionResult $result
+    ) {
+        $this->document = $document;
+        $this->model = $model;
+        $this->capability = $capability;
+        $this->result = $result;
+    }
+
+    /**
+     * Gets the document that was sent to the model.
+     *
+     * @since n.e.x.t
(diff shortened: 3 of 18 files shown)


dugyen:
Tested this PR locally (checked out `feature/text-extraction-poc`, `composer install`, ran the checks CI would run):

| Check | Result |
|---|---|
| `composer test` (full unit suite) | ✅ 1201 tests, 4352 assertions — all passing |
| Filtered to the new text-extraction tests | ✅ 29/29 passing |
| `composer phpcs` | ✅ clean |
| `composer phpstan` | ✅ no errors |
| Merge against `trunk` | ⚠️ conflicts in `src/AiClient.php` and `tests/unit/AiClientTest.php` |

101 tests suite-wide are flagged "risky" on PHP 8.5 due to a deprecated (no-op) `ReflectionProperty::setAccessible()` call — pre-existing pattern in the test suite, not introduced by this PR, and not a real failure.

Code looks solid overall — the main actionable item is rebasing onto current `trunk` to resolve the `AiClient.php` conflict before merge.

dugyen:
Follow-up: a closer read of the diff for possible improvements, beyond the test run.

**Likely bugs**

1. `PageDimensions::fromArray()` (`src/Results/DTO/PageDimensions.php`) and `ExtractedPage::fromArray()` (`src/Results/DTO/ExtractedPage.php`) pass array values straight into `int $width`/`int $height`/`int $pageNumber` constructor params without casting. With `declare(strict_types=1)`, a payload where these arrive as floats (e.g. `"pageNumber": 1.0` from a JSON-decoding provider) throws an uncaught `TypeError` instead of the library's normal `InvalidArgumentException`. The sibling new DTO `BoundingBox::fromArray()` already guards against this with explicit `(float)` casts on its numeric fields — worth adding the analogous `(int)` casts here for consistency.

2. `ModelRequirements::fromExtractionData()` (`src/Providers/Models/DTO/ModelRequirements.php`) hand-rolls the file→modality mapping (`$document->isImage() ? image() : document()`) instead of reusing the existing `inputModalityForPart()` helper a few lines below, which already correctly distinguishes image/audio/video/document/text. As written, an audio or video file passed to `withDocument()` is silently mapped to `document` modality rather than rejected or matched to an audio/video-capable model — likely to produce a confusing provider-side failure instead of a clear client-side error.

**Design/consistency gaps**

3. `TextExtractionBuilder` has no `EventDispatcherInterface` support at all, unlike `PromptBuilder`/`EmbeddingBuilder`, which both dispatch Before/After lifecycle events. `AiClient::document()` correspondingly doesn't forward `self::$eventDispatcher`. Any listener hooked into the shared event bus (logging/metrics/etc.) will be blind to text-extraction calls — worth confirming this is an intentional scope cut for the PoC rather than an oversight.

4. `TextExtractionBuilder::withDocument()` duplicates the string-or-`File` coercion already in `PromptBuilder::withFile()`/`MessageBuilder::withFile()`, and the copy has already drifted — it adds an empty-string guard the other two don't have.

**Minor**

5. `TextExtractionResult::toText()` is a one-line alias of `toMarkdown()` with no behavioral difference (unlike `GenerativeAiResult::toText()`, which does real filtering) — an extra public method/doc/test to maintain for no gain.

None of these block the PR — the happy-path wiring is clean and well-tested — but #1 and #2 look like real correctness bugs worth fixing before merge.

dugyen:
Suggested fixes for the points above:

**1. Missing `(int)` casts in `fromArray()`**
```php
// PageDimensions::fromArray()
return new self(
    (int) $array[self::KEY_WIDTH],
    (int) $array[self::KEY_HEIGHT],
    isset($array[self::KEY_DPI]) ? (int) $array[self::KEY_DPI] : null
);

// ExtractedPage::fromArray()
return new self(
    (int) $array[self::KEY_PAGE_NUMBER],
    $array[self::KEY_MARKDOWN],
    $images,
    isset($array[self::KEY_DIMENSIONS]) ? PageDimensions::fromArray($array[self::KEY_DIMENSIONS]) : null
);
```
Matches the `(float)` casts already used in `BoundingBox::fromArray()`, and keeps bad input surfacing as `InvalidArgumentException` (via the `< 1` checks in the constructors) rather than an uncaught `TypeError`.

**2. Reuse `inputModalityForPart()` in `fromExtractionData()`**
Rather than the inline ternary, either call the existing helper against a synthetic `MessagePart`, or extract a shared `modalityForFile(File $file): ?ModalityEnum` that both `inputModalityForPart()` and `fromExtractionData()` call. Then explicitly reject audio/video (and anything `inputModalityForPart()`-equivalent returns `null` for) with an `InvalidArgumentException` in `withDocument()`/`fromExtractionData()` rather than silently falling through to `document` modality — text extraction's docblock already states it only supports image/document/text, so an unsupported file type should fail fast with a clear message, not surface as a downstream "no model found" or provider error.

**3. Event dispatcher gap**
If this is meant to reach parity with `PromptBuilder`/`EmbeddingBuilder` rather than stay PoC-scoped, add an `EventDispatcherInterface` param to `TextExtractionBuilder`'s constructor (mirroring `EmbeddingBuilder`), wire `AiClient::document()` to pass `self::$eventDispatcher`, and dispatch `Before/AfterExtractTextEvent` around the `$model->extractTextResult($this->document)` call in `extractTextResult()`. If it's an intentional PoC-only cut, a one-line note in the PR description would save the next reader from wondering.

**4. `withDocument()` duplication**
Consider hoisting the string-or-`File` coercion (`is_string($document) ? new File(...) : $document`) into a small static helper on `File` itself (e.g. `File::fromInput($input, ?string $mimeType): File`) that all three builders (`PromptBuilder::withFile()`, `MessageBuilder::withFile()`, `TextExtractionBuilder::withDocument()`) call — and decide once whether the empty-string guard belongs there for everyone, rather than only in the newest copy.

**5. `toText()` alias**
Either drop `toText()` and have callers use `toMarkdown()` directly, or keep it only if a near-term follow-up (e.g. plain-text-without-markdown stripping) is planned — otherwise it's a maintenance no-op copied from `GenerativeAiResult`'s naming without its actual behavior.

saarnilauri:
Thanks @dugyen for checking out the branch, running the checks, and then reading the diff. That second pass caught a real bug.

All five points are now addressed in 162b40c, two of them differently than suggested. The branch is also rebased onto current `trunk`, which resolves the `AiClient.php` and `AiClientTest.php` conflicts you hit.

1. **Missing `(int)` casts in `fromArray()`** fixed in `PageDimensions` and `ExtractedPage`. Worth recording that the sharper form of this bug is that `AbstractDataTransferObject::isArrayShape()` catches only `InvalidArgumentException`, so a `TypeError` escaping `fromArray()` breaks that method's boolean contract. Note also that `strict_types` permits int-to-float widening, so the missing `(int)` casts mattered more than the `(float)` casts already present in `BoundingBox`.
2. **`fromExtractionData()` bypassing `inputModalityForPart()`** real bug, confirmed: `new File('https://example.com/podcast.mp3')` produced `inputModalities: ["document"]`. Fixed by extracting a shared `modalityForFile()` that both `inputModalityForPart()` and extraction use, rather than routing a `File` through a synthetic `MessagePart`. Unsupported types are now rejected in `withDocument()` with a message naming the MIME type, so `isSupported()` keeps its boolean contract and never sees an invalid file.
3. **Event dispatcher gap** not an intentional cut; an oversight against this PR's own spec in #268. `BeforeExtractTextEvent` / `AfterExtractTextEvent` added and wired through `AiClient::document()`. They carry a `File` rather than `list<MessagePart>`, so they parallel the embedding events without literally mirroring them.
4. **`withDocument()` coercion duplication** resolved by removing the drifted empty-string guard rather than adding `File::fromInput()`. `new File('')` already throws `InvalidArgumentException`, so the guard only reworded an existing error while creating the divergence. All three call sites now behave identically and `File` remains the single validator of its own input. The `File::fromInput()` refactor touches `PromptBuilder` and `MessageBuilder` behavior this PR does not otherwise change, so it is better as a follow-up.
5. **`toText()` alias** removed. Beyond being a no-op, `GenerativeAiResult::toText()` filters, so a same-named method with different semantics is worse than no method.

Two further changes prompted by that review:

- **`getPageCount()`** was `count($this->pages)`, which cannot represent what #268 promised ("pages processed", billing-relevant). It is now an optional constructor argument defaulting to the number of returned pages, so connectors can pass the provider's own count (Mistral's `usage_info.pages_processed`). For page-priced providers, where `TokenUsage` is zeros, this is the only billing signal in the result.
- **`ExtractedPage`** now validates that its images are `ExtractedImage` instances, matching how `TextExtractionResult` validates its pages. These DTOs are populated by third-party connectors, which is where runtime validation earns its keep.

Two related points that came out of your review are documented in the PR description rather than here, since they are design decisions rather than fixes:

- **Scope** a table of what this PR defers from #268's step 1 (`ExtractedBlock`, the `KEY_EXTRACTION_*` config keys and their fluent methods, the abstract base class, async operations) and why. Reading the proposal next to the diff, the omissions look like oversights; they are not.
- **Model resolution vs. #274** — #274 landed after this branch was opened and made an explicit model required for embeddings. This PR keeps the resolver for extraction on the grounds that #274's rationale is vector comparability, which extraction has no equivalent of. Happy to align if you or the maintainers prefer uniformity.

Verification on 162b40c: 1250 tests / 4471 assertions passing (44 extraction-specific, up from 29), `composer phpcs` and `composer phpstan` clean locally, and CI green across PHP 7.4 / 8.0 / 8.4 / 8.5 including `codecov/patch`.

Worth noting on the PHP 8.5 risky-test flag you spotted: still the pre-existing `ReflectionProperty::setAccessible()` pattern, and the new tests follow the same convention as their neighbours rather than diverging. Probably worth its own issue.

<sub>These responses were drafted with the assistance of Claude Code (Anthropic) and reviewed by me; the fixes were applied and verified locally before pushing.</sub>

PR #274, opened 2026-08-13
Force a model to be passed when generating embeddings instead of relying on the model resolver

### LINKED ISSUE #242: Implement embedding generation contracts and client APIs

dkotter:
## What?

Instead of relying on the `ModelResolver` to determine which model should be used when running the `EmbeddingBuilder` (which is what the `PromptBuilder` does), require a specific model to be passed in and if not, return an error.

## Why?

Embedding vectors created by one model aren't compatible with vectors created by another model. For this reason it's important that a specific model is always provided instead of relying on a `ModelResolver` to determine this, which has a high likelihood of choosing a different model at various points.

The original version of embeddings (added in #244) closely followed what we already do in the `PromptBuilder`, but we realized that we should deviate slightly to force a model be passed. This PR makes that update and now an error will be returned if someone tries to use the `EmbeddingBuilder` without passing in a specific model.

## How?

- Modify our various `generate` methods to require a model be passed in
- Extract some functionality out of the `ModelResolutionTrait` into a new `ModelConfigurationTrait` and use that in our `EmbeddingBuilder`
- Ensure the `EmbeddingBuilder` sets the model properly and validates that the model is available and will work for the request
- Add a new `getUnmetRequirements` method that will tell us all of the requirements that aren't met, allowing us to provide more detailed error messages

## Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Opus 5
Used for: Evaluating the existing code, iterating on a plan to make the above changes and then executing that plan. All code was reviewed and tested by me

## Testing Instructions

Hard to test this PR on it's own as it requires an AI Provider that supports embeddings. Currently we have upstream PRs that add this support in but those haven't been released yet. Easiest way to test is the following:

1. Remove the old packages. These have been renamed (these are fixed in #275)

```bash
composer remove --dev --no-update \
  wordpress/anthropic-ai-provider \
  wordpress/google-ai-provider \
  wordpress/openai-ai-provider
```

2. Configure the OpenAI fork

```bash
composer config repositories.ai-provider-for-openai vcs \
  https://github.com/chubes4/ai-provider-for-openai
```

3. Install the providers from the right spot (specific branch or release)

```bash
composer require --dev \
  "wordpress/ai-provider-for-anthropic:^1.0" \
  "wordpress/ai-provider-for-google:dev-feature/embeddings" \
  "wordpress/ai-provider-for-openai:dev-feature/issue-32-openai-embeddings"
```

4. Add a `.env` file and on one line, add `GOOGLE_API_KEY=YOUR KEY HERE` and on the next line add `OPENAI_API_KEY=YOUR KEY HERE`
5. Run the integration test suite: `composer test:integration` (Note there are a couple errors with function calling tests but those are existing issues, not related to this PR)

After the above, you can also test directly using our `cli.php` file:

```bash
OPENAI_API_KEY=123456 php cli.php 'Your text here' --providerId=openai --modelId=text-embedding-3-small --outputFormat=embedding-json
```

You should see output on the command line that shows the embedding result.

## Changelog Entry

> Changed - Require a specific model be passed when generating embeddings instead of relying on a model resolver. Note this is a breaking change for anyone that has started to use embedding functionality and will require updates be made.

### DIFF
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -250,6 +250,10 @@ class AiClient
      * embedding generation method such as generateEmbedding() or generateEmbeddings() to produce
      * one vector per input.
      *
+     * A model must be specified on the returned builder via usingModel() or usingProviderModel().
+     * Embedding vectors are only comparable to other vectors from the same model, so no model is
+     * selected automatically.
+     *
      * @since 1.4.0
      *
      * @param EmbeddingInput|list<EmbeddingInput>|null $input Optional initial input(s) to embed.
@@ -415,67 +419,80 @@ class AiClient
     /**
      * Generates embeddings using the traditional API approach.
      *
+     * A model is required. Embedding vectors are only comparable to other vectors from the same
+     * model, so no model is selected automatically.
+     *
      * @since 1.4.0
      *
      * @param EmbeddingInput|list<EmbeddingInput> $input The input(s) to embed.
-     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
-     *                                                        or model configuration for auto-discovery,
-     *                                                        or null for defaults.
+     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
+     *                                                          instance or as a
+     *                                                          [provider ID, model ID] tuple.
+     * @param ModelConfig|null $modelConfig Optional model configuration to apply.
      * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
      * @return EmbeddingResult The embedding result.
      *
-     * @throws \InvalidArgumentException If the input format is invalid.
-     * @throws \RuntimeException If no suitable model is found.
+     * @throws \InvalidArgumentException If the input or model format is invalid, or if the model
+     *                                   cannot fulfill the request.
      */
     public static function generateEmbeddingResult(
         $input,
-        $modelOrConfig = null,
+        $model,
+        ?ModelConfig $modelConfig = null,
         ?ProviderRegistry $registry = null
     ): EmbeddingResult {
-        self::validateModelOrConfigParameter($modelOrConfig);
-        return self::applyModelOrConfig(self::input($input, $registry), $modelOrConfig)
+        return self::getConfiguredEmbeddingBuilder($input, $model, $modelConfig, $registry)
             ->generateEmbeddingResult();
     }
 
     /**
      * Generates an embedding using the traditional API approach.
      *
+     * A model is required. Embedding vectors are only comparable to other vectors from the same
+     * model, so no model is selected automatically.
+     *
      * @since 1.4.0
      *
      * @param EmbeddingInput $input The input to embed.
-     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
-     *                                                        or model configuration for auto-discovery,
-     *                                                        or null for defaults.
+     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
+     *                                                          instance or as a
+     *                                                          [provider ID, model ID] tuple.
+     * @param ModelConfig|null $modelConfig Optional model configuration to apply.
      * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
      * @return Embedding The generated embedding vector.
      */
     public static function generateEmbedding(
         $input,
-        $modelOrConfig = null,
+        $model,
+        ?ModelConfig $modelConfig = null,
         ?ProviderRegistry $registry = null
     ): Embedding {
-        return self::generateEmbeddingResult($input, $modelOrConfig, $registry)->getEmbedding();
+        return self::generateEmbeddingResult($input, $model, $modelConfig, $registry)->getEmbedding();
     }
 
     /**
      * Generates embeddings for a list of inputs using the traditional API approach.
      *
+     * A model is required. Embedding vectors are only comparable to other vectors from the same
+     * model, so no model is selected automatically.
+     *
      * @since 1.4.0
      *
      * @param list<EmbeddingInput> $inputs The inputs to embed.
-     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
-     *                                                        or model configuration for auto-discovery,
-     *                                                        or null for defaults.
+     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
+     *                                                          instance or as a
+     *                                                          [provider ID, model ID] tuple.
+     * @param ModelConfig|null $modelConfig Optional model configuration to apply.
      * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
      * @return list<Embedding> The generated embedding vectors.
      */
     public static function generateEmbeddings(
         array $inputs,
-        $modelOrConfig = null,
+        $model,
+        ?ModelConfig $modelConfig = null,
         ?ProviderRegistry $registry = null
     ): array {
-        self::validateModelOrConfigParameter($modelOrConfig);
-        return self::applyModelOrConfig(self::input($inputs, $registry), $modelOrConfig)
+        return self::getConfiguredEmbeddingBuilder($inputs, $model, $modelConfig, $registry)
             ->generateEmbeddings();
     }
 
@@ -540,12 +557,58 @@ class AiClient
     }
 
     /**
+     * Configures an EmbeddingBuilder with the required model and optional configuration.
+     *
+     * @param EmbeddingInput|list<EmbeddingInput> $input The input(s) to embed.
+     * @param mixed $model The model to use, either as a ModelInterface instance or as a
+     *                     [provider ID, model ID] tuple.
+     * @param ModelConfig|null $modelConfig Optional model configuration to apply.
+     * @param ProviderRegistry|null $registry Optional custom registry to use.
+     * @return EmbeddingBuilder Configured embedding builder.
+     * @throws InvalidArgumentException If the model parameter is not of a supported type.
+     */
+    private static function getConfiguredEmbeddingBuilder(
+        $input,
+        $model,
+        ?ModelConfig $modelConfig = null,
+        ?ProviderRegistry $registry = null
+    ): EmbeddingBuilder {
+        $builder = self::input($input, $registry);
+
+        // Apply the configuration first, so that it takes precedence over the model's own config.
+        if ($modelConfig !== null) {
+            $builder->usingModelConfig($modelConfig);
+        }
+
+        if ($model instanceof ModelInterface) {
+            return $builder->usingModel($model);
+        }
+
+        if (
+            is_array($model)
+            && array_is_list($model)
+            && count($model) === 2
+            && is_string($model[0])
+            && is_string($model[1])
+        ) {
+            return $builder->usingProviderModel($model[0], $model[1]);
+        }
+
+        throw new InvalidArgumentException(
+            'Model must be a ModelInterface instance or a [provider ID, model ID] tuple. ' .
+            'Embeddings are only comparable to other embeddings from the same model, so a model ' .
+            'is required. ' .
+            sprintf('Received: %s', is_object($model) ? get_class($model) : gettype($model))
+        );
+    }
+
+    /**
      * Applies a model or model configuration to a builder.
      *
      * Works with any builder that exposes the shared model resolution methods
      * (see {@see \WordPress\AiClient\Builders\Traits\ModelResolutionTrait}).
      *
-     * @template T of PromptBuilder|EmbeddingBuilder
+     * @template T of PromptBuilder
      *
      * @param T $builder The builder to configure.
      * @param ModelInterface|ModelConfig|null $modelOrConfig Specific model, model configuration,
--- a/src/AiClient.php
+++ b/src/AiClient.php
@@ -12,6 +12,7 @@ use WordPress\AiClient\Common\Exception\InvalidArgumentException;
 use WordPress\AiClient\Common\Exception\RuntimeException;
 use WordPress\AiClient\Providers\Contracts\ProviderAvailabilityInterface;
 use WordPress\AiClient\Providers\Contracts\ProviderInterface;
+use WordPress\AiClient\Providers\ModelResolver;
 use WordPress\AiClient\Providers\Models\Contracts\ModelInterface;
 use WordPress\AiClient\Providers\Models\DTO\ModelConfig;
 use WordPress\AiClient\Providers\ProviderRegistry;
@@ -82,6 +83,7 @@ use WordPress\AiClient\Results\DTO\GenerativeAiResult;
  *
  * @phpstan-import-type Prompt from PromptBuilder
  * @phpstan-import-type EmbeddingInput from EmbeddingBuilder
+ * @phpstan-import-type ProviderModelTuple from ModelResolver
  *
  * phpcs:ignore Generic.Files.LineLength.TooLong
  */
@@ -425,9 +427,7 @@ class AiClient
      * @since 1.4.0
      *
      * @param EmbeddingInput|list<EmbeddingInput> $input The input(s) to embed.
-     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
-     *                                                          instance or as a
-     *                                                          [provider ID, model ID] tuple.
+     * @param ModelInterface|ProviderModelTuple $model The model to use.
      * @param ModelConfig|null $modelConfig Optional model configuration to apply.
      * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
      * @return EmbeddingResult The embedding result.
@@ -454,9 +454,7 @@ class AiClient
      * @since 1.4.0
      *
      * @param EmbeddingInput $input The input to embed.
-     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
-     *                                                          instance or as a
-     *                                                          [provider ID, model ID] tuple.
+     * @param ModelInterface|ProviderModelTuple $model The model to use.
      * @param ModelConfig|null $modelConfig Optional model configuration to apply.
      * @param ProviderRegistry|null $registry Optional custom registry. If null, uses default.
      * @return Embedding The generated embedding vector.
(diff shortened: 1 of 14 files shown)


src/AiClient.php
```diff
@@ -415,67 +419,80 @@ public static function generateVideoResult(
...
-        $modelOrConfig = null,
+        $model,
+        ?ModelConfig $modelConfig = null,
         ?ProviderRegistry $registry = null
     ): EmbeddingResult {
-        self::validateModelOrConfigParameter($modelOrConfig);
-        return self::applyModelOrConfig(self::input($input, $registry), $modelOrConfig)
+        return self::getConfiguredEmbeddingBuilder($input, $model, $modelConfig, $registry)
             ->generateEmbeddingResult();
     }
```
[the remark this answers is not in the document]
dkotter: Yes, this is intentionally a breaking change

ColinM-sys:
## Tested against a local Ollama (no API keys needed)

Since Fueled/ai-provider-for-ollama#69 shipped in 1.2.0, Ollama is a zero-cost way to exercise this PR end-to-end, so I ran it that way instead of the Google/OpenAI path in the testing instructions.

**Setup**
- This branch @ `604a714` (8 commits on top of `trunk` `a31b0ec`), composer deps from lock.
- `fueled/ai-provider-for-ollama` 1.2.0 (`dev-develop 81bee69`) installed as a Composer path dependency; registered via `AiClient::defaultRegistry()->registerProvider(OllamaProvider::class)`.
- Ollama 0.32.0 on the host; `nomic-embed-text:latest` (768-dim, `/api/show` reports `capabilities: ["embedding"]`) as the embedding model, `llama3.1:8b` (`completion, tools`) as the text-only negative case. `OLLAMA_HOST` + a dummy `OLLAMA_API_KEY` (the registry's `{PROVIDER}_API_KEY` env resolution).

**Static / unit**
- `phpunit --testsuite unit`: **OK (1190 tests, 4321 assertions)** on PHP **7.4.33, 8.1.34, 8.4.24, 8.5.9**.
- `phpstan analyse`: no errors.

**Live integration — 36/36 pass on PHP 8.5 and 7.4**
I wrote a script that mirrors `tests/integration/Google/EmbeddingGenerationIntegrationTest.php` case-for-case against Ollama, then adds cases for the new builder behavior:

- All 12 Google-test equivalents pass: single / list / `withInput()`, `usingModel(OllamaProvider::model(...))` binds transport + auth, `usingDimensions(256)` is honored (the provider passes `dimensions` to `/api/embed`), batch 2→2, traditional API with `[provider, model]` tuple, result metadata (`provider=ollama`, model id, `promptTokens=11`), no-model rejected with the new message, text model rejected with `Model "llama3.1:8b" from provider "ollama" does not support embedding generation.`, `File` input rejected locally (`Unsupported options: inputModalities ([image]).`), unknown model rejected.
- `isSupported()`: `true` for the embedding model; `false` (no throw) for a text model, an unknown model id, an unconfigured provider (`openai` with no key) and an unregistered provider id; throws only when no model is set — matches the docblock.
- Generating with an unconfigured provider throws the `Provider "openai" is not registered, or is not configured with valid credentials` message.
- Config precedence: a model instance's own config (`dimensions=128`) is used when the builder sets none; builder `usingDimensions(256)` wins over instance config; `isSupported()` leaves a caller-provided instance's config untouched; replacing the model (`usingModel()` then `usingProviderModel()`) drops the old model's config — full 768 dims, no stale 128 leak.
- `__clone`: changing dims on the clone doesn't affect the original; `usingRequestOptions()` is bound onto the model at generation time.
- Traditional API: `generateEmbeddings(list, [provider, model], ModelConfig{dims:128})` → 2×128; `generateEmbeddingResult(input, ModelInterface)` works; the old `generateEmbedding($input, ModelConfig)` and `($input, null)` signatures are both rejected with the tuple/instance message.
- Batch guard 3→3; same input twice → identical vectors; cosine sanity (related pair 0.653 > unrelated 0.392).

**Trunk vs. this branch, same vendor + same Ollama:** with no model given, `AiClient::input($text)->generateEmbeddingResult()` on `trunk` silently auto-selects `ollama / nomic-embed-text:latest`; on this branch it throws `An embedding model must be specified…`. That's the behavior change, working as intended.

**`cli.php`:** `--outputFormat=embedding-result-json` and `embedding-json --dimensions=64` work; missing `--modelId` → the new CLI error; a text model → the builder's "does not support embedding generation" error; the `--modelPreference` ignored warning fires; the text-generation path is unaffected.

<details>
<summary>Full harness output (PHP 8.5.9)</summary>

```
php-ai-client 1.4.0 | PHP 8.5.9 | OLLAMA_HOST=http://host.docker.internal:11434
ollama configured: true

== Mirror of Google EmbeddingGenerationIntegrationTest ==
PASS  testSingleEmbeddingGeneration  — 768 dims
PASS  testSingleEmbeddingGenerationInputs (list input)  — 768 dims
PASS  testSingleEmbeddingGenerationWithInput (withInput)  — 768 dims
PASS  testEmbeddingGenerationWithModelInstance (usingModel binds deps)  — 768 dims
PASS  testEmbeddingGenerationWithDimensions (usingDimensions 256)  — 256 dims honored by Ollama /api/embed
PASS  testBatchEmbeddingGeneration (2 inputs, one vector each)  — 2 x 256
PASS  testTraditionalApiWithProviderModelTuple  — 768 dims
PASS  testEmbeddingResultMetadataAndTokenUsage  — provider=ollama model=nomic-embed-text:latest promptTokens=11
PASS  testGenerationWithoutModelIsRejected  — threw InvalidArgumentException: "An embedding model must be specified. ..."
PASS  testTextGenerationModelIsRejected (llama3.1:8b)  — threw InvalidArgumentException: "Model "llama3.1:8b" from provider "ollama" does not support embedding generation."
PASS  testFileInputIsRejected (local, no request)  — threw InvalidArgumentException: "Model "nomic-embed-text:latest" from provider "ollama" cannot fulfill this embedding request. Unsupported options: inputModalities ([image])."
PASS  testUnknownModelIsRejected  — threw InvalidArgumentException: "No model with ID definitely-not-a-real-model was found in the provider"

== Additional cases for #274's builder changes ==
PASS  isSupported(): true for embedding model
PASS  isSupported(): false for text model (no throw)
PASS  isSupported(): false for unknown model id (no throw)
PASS  isSupported(): false for unconfigured provider (openai, no key)
PASS  isSupported(): false for unregistered provider id
PASS  isSupported(): throws when no model set  — threw InvalidArgumentException: "An embedding model must be specified. ..."
PASS  generate with unconfigured provider throws clear message  — threw InvalidArgumentException: "Provider "openai" is not registered, or is not configured with valid credentials. ..."
PASS  usingProviderModel rejects empty provider / model ids  — both rejected
PASS  no inputs configured is rejected  — threw InvalidArgumentException: "Cannot generate embeddings from empty input. Add content using withInput()."
PASS  BREAKING: old signature generateEmbedding($input, ModelConfig) now rejected  — threw InvalidArgumentException: "Model must be a ModelInterface instance or a [provider ID, model ID] tuple. ..."
PASS  BREAKING: generateEmbedding($input, null) rejected  — threw InvalidArgumentException: "Model must be a ModelInterface instance or a [provider ID, model ID] tuple. ..."
PASS  traditional API: generateEmbeddings(list, tuple, ModelConfig dims=128)  — 2 x 128
PASS  traditional API: generateEmbeddingResult(input, ModelInterface instance)  — nomic-embed-text:latest, 768 dims
PASS  config precedence: model instance config dims=128 used when builder sets none  — 128
PASS  config precedence: builder usingDimensions(256) beats model instance config dims=128  — 256
PASS  isSupported() leaves a caller-provided model instance untouched  — instance dims still 128 after isSupported()
PASS  replacing the model replaces its config (usingModel then usingProviderModel)  — full 768 dims — stale 128 config did not leak
PASS  clone isolation: cloned builder dims do not affect original  — original 128, clone 256
PASS  usingRequestOptions is bound onto the API model at generation time  — connectTimeout=5.0 visible on model after generate
PASS  INFO: request-option timeout actually enforced by transport (connect timeout 1ms)  — no error — connect completed within 1ms; not a #274 concern
PASS  batch count guard: 3 inputs => 3 vectors  — 3
PASS  determinism: same input twice => identical vector  — identical
PASS  semantic sanity: related text scores higher than unrelated (cosine)  — related 0.653 > unrelated 0.392
PASS  INFO: bare model id without tag ("nomic-embed-text" vs ":latest")  — REJECTED — No model with ID nomic-embed-text was found in the provider

RESULT: 36 passed, 0 failed
```
</details>

**Notes (none blocking)**

1. **Downstream impact on the AI plugin.** The `EmbeddingBuilder` public surface goes from `usingModel, usingModelPreference, usingProvider` on `trunk` to `usingModel, usingProviderModel` here. The AI plugin's `generate_embeddings()` wrapper in `includes/helpers.php` calls both removed methods (`$args['provider']` → `usingProvider()`, `$args['model_preference']` → `usingModelPreference()`), so bringing this back into the plugin (the WordPress/ai#946 follow-up tracked in WordPress/ai#962) means changing that wrapper to take an explicit provider + model (e.g. `$args['provider']` + `$args['model']` → `usingProviderModel()`), and the semantic-search code in WordPress/ai#891 / #943 that passes `model_preference => [ $model ]` adjusts the same way. Expected given the changelog's breaking-change note — just flagging the exact call sites. Happy to pick that up as part of the #962 foundation work.
2. **Ollama-specific, not this PR:** the Ollama provider's model directory keys models by the full `name:tag`, so `usingProviderModel('ollama', 'nomic-embed-text')` is rejected while `'nomic-embed-text:latest'` works. Worth a line in that provider's docs; I can open that over on Fueled/ai-provider-for-ollama.
3. **Testing-instructions suggestion:** with Fueled#69 released, Ollama is a no-key way to run this — `composer config repositories.ollama vcs https://github.com/Fueled/ai-provider-for-ollama`, `composer require --dev fueled/ai-provider-for-ollama`, then `OLLAMA_API_KEY=ollama php cli.php '…' --providerId=ollama --modelId=nomic-embed-text:latest --outputFormat=embedding-json`. If it'd be useful I can turn the run above into a `tests/integration/Ollama/EmbeddingGenerationIntegrationTest.php` that skips unless `OLLAMA_HOST` is set, as a follow-up PR.

Net: behaves exactly as described, no regressions found across PHP 7.4–8.5, and the only action item is the downstream wrapper change in the AI plugin.

src/AiClient.php
```diff
@@ -415,67 +419,80 @@ public static function generateVideoResult(
...
+     * A model is required. Embedding vectors are only comparable to other vectors from the same
+     * model, so no model is selected automatically.
+     *
      * @since 1.4.0
      *
      * @param EmbeddingInput|list<EmbeddingInput> $input The input(s) to embed.
-     * @param ModelInterface|ModelConfig|null $modelOrConfig Optional specific model to use,
-     *                                                        or model configuration for auto-discovery,
-     *                                                        or null for defaults.
+     * @param ModelInterface|array{0: string, 1: string} $model The model to use, either as an
```
JasonTheAdams: Instead of doing `array{0: string, 1: string}` in various places, can we make a type for this and use/import it? That should help clarify what this is without having to explain it in every comment.

src/Builders/EmbeddingBuilder.php
```diff
@@ -141,17 +248,45 @@ public function usingDimensions(int $dimensions): self
...
+            throw new InvalidArgumentException(self::NO_MODEL_MESSAGE);
+        }
+
+        try {
+            $model = $this->locateModel();
+        } catch (InvalidArgumentException $e) {
+            // The model is unusable: its provider is not registered or configured, or the provider
+            // has no model with the given ID. Either way it cannot fulfill the request.
+            return false;
+        }
```
JasonTheAdams, line 288: `isSupported()`'s docblock promises that "any reason the specified model cannot fulfill the request is reported as false," but this catch only covers InvalidArgumentException. On the `usingProviderModel()` path, `locateModel()` → `getProviderModel()` → `getModelMetadata()` can trigger the provider's list-models HTTP request (when it isn't already cached by the availability check), and a transport failure there surfaces as a ResponseException — which extends RuntimeException, not InvalidArgumentException, so it escapes `isSupported()` despite the documented contract.

Since ResponseException and both custom exceptions implement AiClientExceptionInterface, catching that interface here would make the behavior match the docblock:

```php
try {
    $model = $this->locateModel();
} catch (AiClientExceptionInterface $e) {
    // The model is unusable: its provider is not registered or configured, the provider has no
    // model with the given ID, or the provider could not be reached. Either way it cannot
    // fulfill the request.
    return false;
}
```

JasonTheAdams (changes requested):
Thank you, @dkotter! I left a couple of suggestions.

src/AiClient.php
dkotter: Makes sense, added in https://github.com/WordPress/php-ai-client/pull/274/changes/0b886ec50d23b59a2aab495badc9580ee16a3cf4

src/Builders/EmbeddingBuilder.php
dkotter: Yeah, I missed that exception type. Addressed in https://github.com/WordPress/php-ai-client/pull/274/changes/ec5fdbc747bb380066fa95dfe55d53a2a62684ef

JasonTheAdams: I noticed in your commit that you replaced `InvalidArgumentException` with the new exception. Did you mean to replace it or have it as both?

src/AiClient.php
JasonTheAdams: Thanks! Mind adding a comment to the definition line that explains the shape?

JasonTheAdams (approved):
Left to small questions, @dkotter, but I believe this is good to go once those are addressed. Please make sure to run the integration tests to confirm it's working.

Thanks for doing this!

src/Builders/EmbeddingBuilder.php
dkotter: @JasonTheAdams Not sure I completely follow? I replaced `InvalidArgumentException` with `AiClientExceptionInterface` which I think matches your code example above. Are you suggesting to keep `InvalidArgumentException` and just add `AiClientExceptionInterface` as well?

src/AiClient.php
dkotter: Added code comment in https://github.com/WordPress/php-ai-client/pull/274/commits/275fb4735b730534e23ca867880e404d78d41623

src/Builders/EmbeddingBuilder.php
JasonTheAdams: Sorry, I didn't mean for my example to be an exact drop-in, but I can see how that was confusing. The `isSupported()` method is meant to always return false. We can either just catch all exceptions through `Exception` or list out each exceptions, which includes `AiClientExceptionInterface` and `InvalidArgumentException` (drilling into nested methods). It may be worth simply doing the former, but I'm not sure if we'd want to prevent unexpected exceptions from getting through. I don't think it's a big deal, so I'm good either way.

ColinM-sys:
Re-validated on `275fb47` (all three review-response commits included):

- Unit suite: **OK, 1191 tests / 4322 assertions** (PHP 8.5); phpstan clean.
- The 36-case live Ollama harness from my report above: **36/36, no regressions**. (The official integration suite needs Google/OpenAI keys, so a local Ollama with `nomic-embed-text` remains my zero-key stand-in for it.)

On the open `isSupported()` catch question — some live data that may help settle it. What each failure mode actually throws on this branch (generate path, real Ollama provider):

| Case | Thrown | `AiClientExceptionInterface`? | `isSupported()` |
| --- | --- | :-: | :-: |
| Unregistered provider | `Common\Exception\InvalidArgumentException` | yes | `false` |
| Registered, unconfigured (no key) | `Common\Exception\InvalidArgumentException` | yes | `false` |
| Unknown model ID | `Common\Exception\InvalidArgumentException` | yes | `false` |
| Text-generation model | `Common\Exception\InvalidArgumentException` | yes | `false` |
| Unreachable host (cold start) | `Http\Exception\NetworkException` | yes | `false` |

The client's own `InvalidArgumentException` implements `AiClientExceptionInterface` (all of the client's exceptions do — the interface's docblock says that's its purpose), so the current single `catch (AiClientExceptionInterface $e)` already covers both exceptions named in this thread; nothing the client itself throws can escape. Broadening to `\Exception` would only additionally swallow non-client exceptions (e.g. a third-party library exception a provider implementation failed to wrap), which are arguably programming errors better surfaced than reported as "not supported."

One nuance that might merit a sentence in the docblock: because provider availability and model metadata are cached for the remainder of the request (as documented), a provider that was reachable earlier in the same request and then becomes unreachable still reports `isSupported() === true` — no HTTP happens on the later check. A cold start against an unreachable host correctly returns `false`. Consistent with the documented caching; just noting it so `isSupported()` isn't read as a liveness probe.

src/Builders/EmbeddingBuilder.php
dkotter: Parsing through our code, it appears that in almost all situations, catching `AiClientExceptionInterface` is what we'd want here. The one area where this isn't accurate is if a request fails with a status code that isn't in the 3xx-5xx range, that throws a native `\RuntimeException` and won't currently be caught.

So a couple options here as I see it:

1. Widen the catch here to be just `Exception`, which catches that one edge case from our code and potential bugs from third party code. But that does mean we hide actual errors that may be better being surfaced
2. Keep it as `AiClientExceptionInterface` and look to update `ResponseUtil.php` to return our custom `RuntimeException` instead of the native one, which then ensures all of our code is properly caught
3. Keep it as `AiClientExceptionInterface` and not worry about edge cases?

Any preference on these @JasonTheAdams?

JasonTheAdams: Ah, I was thinking about this wrong. Thanks for the clarification (and you as well, @ColinM-sys). Let's go with 3. The `RuntimeException` in `ResponseUtil` is very much a "what on earth?" final catch and should never be reached. If it is we want folks to figure out what happened.

Issue #277, opened 2026-08-19
MessagePart cannot represent source attribution, so every provider drops citations


juanlentino:
`MessagePart` (`src/Messages/DTO/MessagePart.php`) can carry exactly seven things:
`channel`, `type`, `thoughtSignature`, `text`, `file`, `functionCall`,
`functionResponse`. None of them can hold a citation.

All three first-party providers return source attribution, in three different
shapes (per vendor documentation):

| Provider | Field |
|---|---|
| Anthropic | `citations` on text blocks — `cited_text`, `document_index`, `document_title`, typed location |
| OpenAI | `annotations` — `type: "url_citation"`, `url`, `title`, `start_index`, `end_index`, plus `sources` |
| Gemini | `citationMetadata.citationSources` and `groundingMetadata`; `annotations`/`url_citation` on the Search-grounding surface |

And all three provider packages currently drop it. Verified by reading the source:
`AnthropicTextGenerationModel::parseResponseContentMessagePart()`'s `case 'text'`
keeps only `$partData['text']`; `GoogleTextGenerationModel` and
`OpenAiTextGenerationModel` contain no citation/grounding/annotation handling at all.

This is a gap rather than a data-loss bug — no provider currently *requests*
citations, so nothing is being lost in production today. But the parse paths are
not ready, and no provider can be made ready while the DTO has nowhere to put the
result.

**Suggested shape:** `thoughtSignature` is the precedent — an opaque,
provider-supplied, round-trippable field already accepted on `MessagePart`. A
citation carrier could follow it, normalising the three vendor shapes into one
representation (source identifier/URI and optional title).

Related: #268 (provider-agnostic text extraction) — same theme of getting
structured meaning out of documents.

Happy to open a PR once maintainers settle the shape; the API-surface decision is
yours.

the-hercules:
@juanlentino I agree that we need a normalized way to handle citations across all providers. 

To address the API-surface blocker you mentioned, I am currently implementing a normalized `Citation` object. Here is the approach I am taking:

1. **New `Citation` DTO:** I've created a `Citation` class to serve as a uniform representation across all vendors. It contains two properties:
   - `uri` (string): The source identifier (e.g., URL or document index).
   - `title` (string|null): An optional title for the source.

[...]

PR #283, opened 2026-08-30
Add Citation DTO core contract to MessagePart

the-hercules:
Implements #277 (Sets up core DTO foundation).
This PR establishes the core DTO foundation to support provider-agnostic source citations, addressing the API-
  surface blocker discussed in #277.

 Following the feedback and consensus in the issue, this introduces a normalized `Citation` object and attaches it
  to `MessagePart`. The DTO is shaped to represent the common citation formats from Google, OpenAI, and Anthropic
  while keeping the sources checkable:

  - **Typed Source Identity**: Replaces a generic URI with explicit, nullable `url` (for remote sources) and
  `documentIndex` (for request-provided documents) to avoid string sniffing.
  - **Text Offset Ranges**: Includes `startIndex` and `endIndex` to anchor the citation locally to the owning
  `MessagePart`'s text string.
    - **Quoted Text**: Includes `quotedText` to cleanly carry Anthropic's block-based `cited_text`.
    - **Validation**: Enforces that offset ranges are strictly non-negative and properly ordered (`startIndex <=
  endIndex`), throwing an `InvalidArgumentException` otherwise.

Note: This PR handles the core contract/DTOs within the agnostic client. Follow-up PRs in the respective provider plugin repositories (ai-provider-for-anthropic, etc.) will be required to update their parsing logic to utilize this new format.
  
  **AI Disclosure**
  Model - `Gemini 3.1 Pro`
  Usage - `Implementation of functionality and test after discussing in Plan Mode`
  Human Loop - `Reviewed code against the issue's requirements`

### DIFF
--- a/src/Messages/DTO/Citation.php
+++ b/src/Messages/DTO/Citation.php
@@ -87,6 +87,12 @@ class Citation extends AbstractDataTransferObject
         ?int $endIndex = null,
         ?string $quotedText = null
     ) {
+        if ($documentIndex !== null && $documentIndex < 0) {
+            throw new InvalidArgumentException(
+                sprintf('Citation documentIndex must be non-negative, got %d.', $documentIndex)
+            );
+        }
+
         if ($startIndex !== null && $startIndex < 0) {
             throw new InvalidArgumentException(
                 sprintf('Citation startIndex must be non-negative, got %d.', $startIndex)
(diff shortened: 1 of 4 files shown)


zahidgh (commented):
One validation gap: `documentIndex` represents an index into the request’s documents array, but the constructor and JSON schema currently allow negative values. A `Citation(null, -1, ...)` can therefore serialize successfully and leave providers to handle an invalid reference. Please reject `documentIndex < 0`, add `minimum: 0` to its schema, and cover the guard with a unit test.

the-hercules:
@zahidgh Implemented your suggestion, take a look when you get time. Thanks!

PR #282, opened 2026-08-26
[#281] Add generic annotations to function declarations

superdav42:
## Summary

Revised following stakeholder feedback to add **only generic function metadata**, while tool-search behavior is explored in provider plugins.

```php
$function = new FunctionDeclaration(
    'get_weather',
    'Gets the weather',
    null,
    ['deferredLoading' => true]
);

$metadata = $function->getMetadata();
```

The SDK stores annotations but does not interpret their names or values. The motivating provider experiment can combine function metadata with existing `ModelConfig::setCustomOptions(['deferredLoading' => true])`; no feature-specific core option is introduced.

## Scope

**One production file:** `src/Tools/DTO/FunctionDeclaration.php`.

- Optional fourth `array $metadata = []` constructor argument.
- `getMetadata()` accessor and optional `metadata` array/JSON representation.
- Empty metadata is omitted, preserving the serialized shape of existing declarations; missing metadata restores to an empty array.
- Existing constructor calls and PHP 7.4 compatibility are retained.

The previous proposal's `ProviderData` class, message type, serialization changes, and tests have been removed. There are **no changes** to messages, `PromptBuilder`, `ModelConfig`, model requirements, or enums. There is no `usingToolSearch()` method or dedicated deferred-loading property.

## Provider experiment and limitations

Companion implementation: [WordPress/ai-provider-for-openai #50](https://github.com/WordPress/ai-provider-for-openai/pull/50).

The provider owns annotation semantics, automatic activation above 10 functions, model eligibility, and request-level overrides. It translates recognized annotations into native wire fields rather than blindly forwarding metadata. Automatic/request-level behavior also works with released SDK 1.3.1; only per-function annotations need this extension.

The experiment uses **OpenAI-managed state via the existing `previous_response_id` custom option**, with only new input/function results on subsequent turns. Stateless replay of search output is explicitly outside this proposal. This avoids adding any new message representation before the feature has demonstrated value.

For #281. This is an extension point, not a claim to complete provider adoption or establish tool-search utility.

## Reference and review map

- `src/Tools/DTO/FunctionDeclaration.php`: follows its existing constructor/accessor/schema/array transformation pattern.
- `tests/unit/Tools/DTO/FunctionDeclarationTest.php`: legacy serialization shape, arbitrary nested annotations, boolean false/null preservation, array/JSON round trips, model-config round trips, and clone/getter independence for array metadata.
[...]

Issue #281, opened 2026-08-23
Add provider-agnostic deferred tool loading and tool discovery support


superdav42:
## Context

Function declarations are currently sent to the model in full on every request. This works well for a small set of functions, but becomes expensive when an application exposes dozens or hundreds of tools, such as a large WordPress Abilities API catalog.

OpenAI and Anthropic now support deferred tool loading. Applications still declare the available tools, but selected definitions are withheld from the model context until the model searches for and loads them.

This issue proposes a provider-agnostic representation for:

- marking function declarations as deferred;
- requesting tool discovery;
- representing tool discovery calls and results when they need to be carried across turns;
- declaring whether a model or provider supports native deferred loading.

Provider implementations would translate this common configuration into their native API formats.

### Lower token usage and cost

Tool parameter schemas count as input tokens. Sending a large tool catalog on every request can consume a material portion of the context window before the conversation begins.

Deferred loading lets the model see only the tools needed for the current task.

### Better prompt caching

OpenAI and Anthropic append discovered tools later in the context instead of changing the cached prompt prefix. This preserves cache reuse as tools are discovered.

### Larger tool catalogs

Applications could register a broad catalog of WordPress abilities without manually choosing a small subset before every prompt.

### A portable API

Both OpenAI and Anthropic use the `defer_loading` concept, although their discovery protocols differ. Representing the intent in PHP AI Client would prevent applications from depending on provider-specific request options.

Providers without native support would work as they always have by including all tools in the system prompt without the benefits of deferred loading and support. Alternatively we could implement the behavior client side with a dedicated search_tools tool but this would probably be best in a follow up issue.

### OpenAI

OpenAI supports tool search in the Responses API on GPT-5.4 and later models.

To use it, a request:

1. adds `{"type": "tool_search"}` to `tools`;
2. marks deferred functions or MCP servers with `defer_loading: true`;
3. optionally groups functions into namespaces.

OpenAI supports two execution modes:

- hosted search, where OpenAI selects and loads tools within the same response;
- client-executed search, where the model returns a `tool_search_call` and the client supplies a `tool_search_output`.

Hosted responses can include `tool_search_call` and `tool_search_output` items before the eventual function call. These items should not be silently lost if they are needed for conversation replay.

OpenAI doc says the hosted search would be preferred if all the available tools are known and defined so we should focus on the hosted search.

Documentation:

- [OpenAI tool search](https://developers.openai.com/api/docs/guides/tools-tool-search)
[...]

PR #289, opened 2026-09-10
[#281] Add ProviderData for provider-native message parts

superdav42:
## Why this DTO is needed

Providers return native conversation items that are neither text, files, nor client-executed function calls/responses. Dropping these items loses information needed for replay or attribution; representing them as ordinary function calls gives them the wrong meaning. `ProviderData` gives these items a serializable, provider-tagged place in message history without adding a core type for every provider-specific format.

This extracts the message-data portion previously proposed in #282 into an independent PR. #282 remains focused on function annotations.

## Issues this helps resolve

- For #281: preserve tool-discovery calls/results in their original position in conversation history so providers can replay them on subsequent turns.
- For https://github.com/WordPress/ai-provider-for-google/issues/32: give Google's `toolCall` / `toolResponse` parts a representation that retains their payloads, including signatures and search attribution data, instead of throwing or discarding them.

This is the shared DTO foundation, not a complete fix for either issue. Provider adapters still need to parse and replay the new part type; the Google request configuration fix also remains provider-owned.

## Changes

- `src/Messages/DTO/ProviderData.php`: provider ID and opaque data getters, array transformation and JSON schema, following the existing `AbstractDataTransferObject` pattern.
- `src/Messages/DTO/MessagePart.php` and `src/Messages/Enums/MessagePartTypeEnum.php`: a `PROVIDER_DATA` part with construction, access, serialization and cloning support. Existing message shapes remain unchanged.
- Corresponding tests under `tests/unit/Messages/`: array/JSON round trips, ordered mixed-message history, signatures/channels, missing fields, schema and clone behavior.

```php
$part = new MessagePart(new ProviderData('google', $partData));
```

Adapters must check `getProviderId()` before interpreting or replaying `getData()`. The DTO does not enforce that check or validate the native payload. No provider implementation, builder API, function annotation or tool-search policy changes are included. PHP 7.4 compatibility is retained.

## Verification

- `composer test:unit`: **1,213 tests / 4,426 assertions passed**.
- `composer lint`: **PHPCS and PHPStan passed**.
- `git diff --check`: passed.
- No paid integration calls were run.

---
[aidevops.sh](https://aidevops.sh) v3.32.317 plugin for [OpenCode](https://opencode.ai) v1.18.30 with gpt-6-astra spent 1d 1h and 2,585,669 tokens on this with the user in an interactive session.

### DIFF
--- a/src/Messages/DTO/MessagePart.php
+++ b/src/Messages/DTO/MessagePart.php
@@ -24,6 +24,7 @@ use WordPress\AiClient\Tools\DTO\FunctionResponse;
  * @phpstan-import-type FileArrayShape from File
  * @phpstan-import-type FunctionCallArrayShape from FunctionCall
  * @phpstan-import-type FunctionResponseArrayShape from FunctionResponse
+ * @phpstan-import-type ProviderDataArrayShape from ProviderData
  *
  * @phpstan-type MessagePartArrayShape array{
  *     channel: string,
@@ -32,7 +33,8 @@ use WordPress\AiClient\Tools\DTO\FunctionResponse;
  *     text?: string,
  *     file?: FileArrayShape,
  *     functionCall?: FunctionCallArrayShape,
- *     functionResponse?: FunctionResponseArrayShape
+ *     functionResponse?: FunctionResponseArrayShape,
+ *     providerData?: ProviderDataArrayShape
  * }
  *
  * @extends AbstractDataTransferObject<MessagePartArrayShape>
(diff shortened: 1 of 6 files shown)


JasonTheAdams (commented):
@superdav42 Why would provider data be considered part of the content of the `MessagePart`? This is more like arbitrary data, annotations like the other PR, that describe something about the `MessagePart` to the provider, right? Is there a scenario where there would be a `MessagePart` that contains only provider data?

**Edit:**
Reading this over more, it seems like the answer is yes. Can you please provide reference to some provider docs that shows this behavior? I want to make sure this is the right place to put this sort of thing.

I'm also confused why we're passing the provider ID to the `ProviderData`. Would a given set of messages potentially contain data for multiple providers?

